PR #14, opened 2014-12-11
Update configuration.php

ayoubabid:
The previous configuration show only the first problem. Now you get the hole summary about hosts and services situation.


lippser:
Hi,

We think it's best to show the service grid filtered by problems by default because in most cases you have so many hosts and services that displaying the unfiltered grid makes no sense. Although the grid supports filters, it still lacks a filter editor. Once we add a filter editor, we review having the grid filtered by problems by default. For now I'll move the service grid link beneath the problems section and close this pull request. Thanks for your effort nonetheless.

Cheers,
Eric

PR #17, opened 2015-01-12
setup-welcome: fix cli doc

resmo:
* `createDirectory` does not exist, this should be `direcotory`.  
* For changing the group option `--group` must be used. 
* Default config directory is `/etc/icingaweb2` not `/etc/icingaweb`.


lippser:
Thanks for the patch ;) I'll apply it in a second.

Cheers,
Eric

PR #24, opened 2015-04-10
Removed quotation marks in password example

thelfensdrfer:
Otherwise the quotation marks are included in the password.


lippser:
Hi,

Thanks for the path. I applied it to master branch.

Cheers,
Eric

PR #27, opened 2015-05-05
do not require bind during setup (#9224)

chulkilee:
Reference: https://dev.icinga.org/issues/9224

### DIFF
--- a/application/forms/Config/Resource/LdapResourceForm.php
+++ b/application/forms/Config/Resource/LdapResourceForm.php
@@ -113,7 +113,7 @@ class LdapResourceForm extends Form
             'text',
             'bind_dn',
             array(
-                'required'      => true,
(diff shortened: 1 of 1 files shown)


lippser:
Hi,

Thanks for creating the issue and submitting your pull request. The issue has been resolved in the lastest master.

Cheers,
Eric

chulkilee:
Thanks for the fix.

Here are some suggestions.

- It would be nice to have the actual commit for this PR. In this case, 250f863faa462cbd9e9f71b5cf42c0da36c8477e
- Do not use GitHub PR if this repo is just for mirroring. No PR is actually merged so far. If the rule is to use dev.icinga.org for collaboration, keep it focused. If you need more engagement via GitHub, do it in GitHub way.

PR #32, opened 2015-06-17
Changing password fields in backend to hide passwords when typing

realitygaps:
When adding and editing users, the password displays as it is typed. It probably makes sense to use a password field for that so that other people watching dont see the typed password on the screen.


lippser:
Hi,

Thanks for the patch. I applied it to the master branch.

Best,
Eric

PR #34, opened 2015-06-25
Update services.phtml

Muhkoenig:
Replace \<br> with \n in plugin output of services list view.


lippser:
Hi,

I close this one because we're already using the pluginOutput helper in the services view script for replacing br tags. Further your regex would not match <br />.

Best,
Eric

PR #35, opened 2015-06-26
Update icingacli script to be portable

xert:
Current shebang `#!/usr/bin/php` is not correct for example on FreeBSD.

### DIFF
--- a/bin/icingacli
+++ b/bin/icingacli
@@ -1,4 +1,4 @@
-#!/usr/bin/php
+#!/usr/bin/env php
 <?php
 /* Icinga Web 2 | (c) 2013-2015 Icinga Development Team | GPLv2+ */
 


larsengels:
Yes, please use "#!/usr/bin/env php" and "#!/usr/bin/env python"

Thomas-Gelf:
Hi xert,

we had a related (and reverted) change a couple of months ago: https://dev.icinga.org/issues/7789. I still do not really like env. However I could live with what I proposed at the time being: correct php shebang line in ALL built packages and env only for those (like me) who run Icinga Web 2 directly from a git clone.

Could you please open an issue for discussion on https://dev.icinga.org/projects/icingaweb2/issues, referring to the issue mentioned above and this pull request?

Cheers,
Thomas

xert:
Opened https://dev.icinga.org/issues/9615

larsengels:
What is wrong with env? It's what all Shell / Python / Perl / $FOO scripts should use because it is portable.

lippser:
Hi xert,

I applied your patch to the master branch. Thanks for that one.

Best,
Eric

PR #39, opened 2015-07-22
DbUserBackend - inspect reports 0 users when one does exist

rkubik:
I notice that when I have a single user, ie. an administrator, I cannot create a Configuration/Authentication User Backend because the form "Edit User Backend" fails with error message: "Cannot... 0 active users". Looking in the file "DbUserBackend.php", the comparision below seems odd to me:
    
    $users = $this->select()->where('is_active', true)->count();
    if ($users > 1) {
        $insp->write(sprintf('%s active users', $users));
    } else {
        return $insp->error('0 active users', $users);
    }

If I have 2+ users, report "2 active users", otherwise I have "0 active users". Should that be:

    if ($users > 0) {


lippser:
Hi,

Thanks for the fix. I applied your change to master branch.

Best,
Eric

PR #40, opened 2015-07-24
icingcli does not exist on deb based systems

formorer:
Topic says it all, icingacli doesn't exist as a package on deb based systems.


lippser:
Hi,

I applied your patch. Thanks for the update.

Best,
Eric

PR #44, opened 2015-09-22
Fixes for ldap group auth

analogrithems:
The current LdapUserGroupBackend was incomplete and suffered from a little over zealous copy pasta.  It had over written certain functions that where unnecessary such as the constructor and a table validator.  This patch aims to clean those up.  Additionally it also makes this group auth work with posixGroup that use the username as the member identifier and not just inetGroups that use the full dn


analogrithems:
handled in a new pr

PR #45, opened 2015-09-30
This is a minor fix in follow up to the recent fixes for the ldap backend.  

analogrithems:
This will fully work with a normal posixGroup objectClass backend.


mcm:
+1. This fixed an issue we were having because we use groupsOfNames instead of regular groups.

lippser:
Hi guys,

Could please test our latest master w/o your patches applied?

Best,
Eric

PR #46, opened 2015-10-01
Feature/datetime format configuration option and default change

jeremydonahue:
Adds datetime_format as a configurable option in Configuration > Application and uses it to format the timestamp in service and host history views.


lippser:
Hi,

Thanks for the PR but I'll close it. We changed the datetime format to a more reasonable default for every view. If you still like to see this feature, please have a look at \Icinga\Date\DateFormatter which is used for date and time rendering. Format changes should happen here because this affects all views - not only history. Some time ago I implemented date and time rendering according to the user's locale but we decided against it. This may be an option too.

Cheers,
Eric

PR #50, opened 2015-10-08
Add CMDB Hook

lum3n:
We would like to add a CMDB Hook to display further information in host-/service-pages.

### DIFF
--- a/library/Icinga/Web/Hook/CmdbHook.php
+++ b/library/Icinga/Web/Hook/CmdbHook.php
@@ -1,0 +1,22 @@
+<?php
+namespace Icinga\Web\Hook;
+
+use Icinga\Exception\ProgrammingError;
+use Icinga\Module\Monitoring\Object\MonitoredObject;
+
+abstract class CmdbHook extends WebBaseHook
+{
+    final public function __construct()
+    {
+        $this->init();
+    }
+
+    protected function init()
+    {
+    }
+
+    public function getHtml(MonitoredObject $object)
+    {
+        throw new ProgrammingError('This function needs to be implemented in Hook.');
+    }
+}
--- a/modules/monitoring/application/views/scripts/partials/object/detail-content.phtml
+++ b/modules/monitoring/application/views/scripts/partials/object/detail-content.phtml
@@ -47,6 +47,6 @@
     </tbody>
     </table>
     <?php endif ?>
-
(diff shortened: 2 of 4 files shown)


xam-stephan:
:+1:

Thomas-Gelf:
Hi lum3n,

while your code absolutely makes sense I would prefer to not teach people how to "enrich" the monitoring detail view with custom HTML. Excessive use would lead to many data sources being queried at once. Imagine a massive outage, network or storage problems. You do not want components like a CMDB to slow down or break your monitoring frontend in situations where you need it at most.

I wrote some modules talking to CMDBs and similar, usually based on the Host/ServiceActions Hook. My "whatevercmdb"-modules provide small preview pages showing external content, the monitoring detail view has action links pointing to that page.

We could also easily adjust the monitoring module to support more than one single grapher hook implementation at once. That way your module could provide a fake grapher, being allowed to feed custom HTML. I'm aware of the fact that implementing a "CMDB" hook would feel better. Still, I do not want to tell people that such thing exists.

Cheers,
Thomas

NB: Could you please open an issue on https://dev.icinga.org referencing this pull requests for farther discussion?

lum3n:
Hi Thomas,

thank you for your reply.

This makes sense to me but is still sad to hear. Especially because we are using it successfully for weeks now. We were happy to see the Icinga Web 2s modularity which finaly enabled us to combine multiple tools (CMDB, ticket tool and more monitoring/alerting systems) to a "single point of contact" with the Web2 as its base. We already got great feedback from other teams who don't need to use 2-3 different tools at once any more when working on incidents.

To the point of stability: As an example, we replicated the necessary CMDB-parts to a local database whereby we are even able to avoid network problems. So in my opinion it is up to you and especially your responsibility what you do in this hook. Of course you have to take care of what you do but it also gives you the ability to display the exactly, more company-specific, information which is a huge time and effort safer.

Alternatively it would be great, if you could alter the monitoring module to support multiple grapher as you suggested, so that we can keep on using our functions.

I am going to open the issue as well.

Thank you and regards,
Ken

lum3n:
I opened the following issue: https://dev.icinga.org/issues/10463

Thomas-Gelf:
We agreed on a compromise, I'll close this pull request in favour of https://dev.icinga.org/issues/10463

PR #51, opened 2015-10-09
Escape < and > to avoid false tags

bebehei:


lippser:
Hi,

Thanks for your PR. Your changes made it to the master branch.

Cheers,
Eric

PR #56, opened 2015-12-18
Fix the wrong mask for temp. directory in FileCache.php

kobmaki:
Detail see https://dev.icinga.org/issues/10892.

ref #10892


Thomas-Gelf:
Committed, thank you!

PR #59, opened 2016-02-09
Add SSL support to MySQL database resources

leeclemens:
fixes 11115

### DIFF
--- a/application/forms/Config/Resource/DbResourceForm.php
+++ b/application/forms/Config/Resource/DbResourceForm.php
@@ -138,6 +138,53 @@ class DbResourceForm extends Form
                 'description'       => $this->translate('The password to use for authentication')
             )
         );
+        if ($offerMysql) {
+            $this->addElement(
+                'text',
+                'ssl_key',
+                array(
+                    'required'          => false,
+                    'label'             => $this->translate('SSL Key'),
+                    'description'       => $this->translate('The SSL client key file path')
+                )
+            );
+            $this->addElement(
+                'text',
+                'ssl_cert',
+                array(
+                    'required'          => false,
+                    'label'             => $this->translate('SSL Certificate'),
+                    'description'       => $this->translate('The SSL certificate file path')
+                )
+            );
+            $this->addElement(
+                'text',
+                'ssl_ca',
+                array(
+                    'required'          => false,
+                    'label'             => $this->translate('SSL CA'),
+                    'description'       => $this->translate('The SSL Certificate Authority certificate file path')
+                )
+            );
+            $this->addElement(
+                'text',
+                'ssl_capath',
+                array(
+                    'required'          => false,
+                    'label'             => $this->translate('SSL CA Path'),
+                    'description'       => $this->translate('The SSL trusted SSL CA certificates in PEM format directory path')
+                )
+            );
+            $this->addElement(
+                'text',
+                'ssl_cipher',
+                array(
+                    'required'          => false,
+                    'label'             => $this->translate('SSL Cipher'),
+                    'description'       => $this->translate('The SSL list of permissible ciphers')
+                )
+            );
+        }
         $this->addElement(
             'text',
             'charset',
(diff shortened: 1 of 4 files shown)


leeclemens:
Still fighting with SELinux to complete testing this, but wanted to get code review started

nilmerg:
Hi Lee,

thanks for your efforts to contribute to this project! But before we will merge your changes, you will need to adjust a few things.

Those empty INI-properties in [resources.ini.erb](https://github.com/leeclemens/icingaweb2/commit/c703dd14c5c50c92bb6dec7c3140d5351b67e462#diff-9e46db422a4b97f360d4d47cc1421d4d) are redundant. Please do not add such unless they have got a meaningful value.

The way you are adding [form fields](https://github.com/leeclemens/icingaweb2/commit/c703dd14c5c50c92bb6dec7c3140d5351b67e462#diff-1872579057b0b6daa4ab08e7694809a6) is incomplete. The first argument to addElement() is the type name of the element. (e.g. text) Additionally, the property key of each element that is called the same as its element causes an html element property with this name to be created. I guess you wanted to set a default value for it, for which you should use the 'value' key and a value of your choice. (null in this case causes the field being empty by default, which is the default anyway)

The "if" you have added in [DbConnection.php](https://github.com/leeclemens/icingaweb2/commit/c703dd14c5c50c92bb6dec7c3140d5351b67e462#diff-e8919bff1da1c565889fe24432c6b514) is also redundant as there is a switch case underneath it with a case for mysql.

Changing [vendor library files](https://github.com/leeclemens/icingaweb2/commit/c703dd14c5c50c92bb6dec7c3140d5351b67e462#diff-016089da19d2a29c34d97ca605b15d61) is not an option for us. Those libraries exist only for backwards compatibility or simplified distribution. They are not supposed to be adjusted to match our requirements. If there is functionality you miss, try to add it in a way that works with previous, current and, preferable, future versions of the library.

Best,
Johannes

leeclemens:
Hi Johannes,

Thank you for taking the time to help me out.

I hadn't considered updating the Zend files wouldn't be an option. Unfortunately that is what actually enables using SSL for the connection (the call to `mysqli_ssl_set`).

I just found, almost 7 years ago a bug was created and in 2010 a patch was provided.
Still marked Open. Only comment is the patch.
http://framework.zend.com/issues/browse/ZF-6140

Thanks,
Lee

leeclemens:
Updated without modifying Vendor libraries...had to put similar code in 3 places to set ['driver_options'] and use Zend PDO package specifically (wasn't working when I tried with IUS 5.6 packages)

leeclemens:
Just saw the new release, any chance this feature can be at least added to the roadmap for a future release?

leeclemens:
Just wanted to note that MySQL SSL support was merged in to icinga2: https://github.com/Icinga/icinga2/pull/77
https://dev.icinga.org/issues/9725

gunnarbeutner:
Ah, yeah, I guess this is done then. :) Thanks!

leeclemens:
@gunnarbeutner Forgive my confusion, MySQL SSL support was added to icinga2 (IDO connection),  this PR is to also add MySQL SSL support to icingaweb2's connections.

gunnarbeutner:
Oh, sorry. Re-opening. :)

nilmerg:
Hi Lee,

thanks for the update and sorry for the late response. I had a look at it last week. It's now in a feature-branch and targeted for v2.4.0.

There were some changes necessary though:
* This will introduce *six* new form elements so they shouldn't be shown until the user decides to tinker with them, so I've added an additional setting that enables those options
* You've forgot to update the summary page for the database wizard page of the framework wizard. I've added this for you
* Since this will currently only work with php 5.4+, I've added a check that causes the new form elements not to be shown for earlier versions
* The *weirdness* you've mentioned in our tracker is a bug, which is now fixed

Once v2.4.0 is out, you'll need to re-configure your resources because the new option disables all ssl options by default.

Best,
Johannes

leeclemens:
Hi Johannes,

Thank you for taking the time to clean this up! Greatly appreciative :)

Thanks,
Lee

PR #65, opened 2016-06-29
doc: document 'encryption' directive for LDAP

chris-reeves:
Documentation for the 'encryption' directive for the LDAP resource type was missing (I had to go digging through the source). This PR adds a single line of documentation describing that directive.


lippser:
Hi,

Thanks for your patch. I applied it to the master branch.

Best regards,
Eric

PR #67, opened 2016-09-15
Allow table header in pluginoutput

friesoft:
Add another html entity which is allowed (`<th>`). This allows us to render tables with a nice looking heading.

Fixes #12125


lippser:
Hi,

Thanks for the patch. It's upstream now.

Best regards,
Eric

PR #68, opened 2016-09-15
Allow authentication based on arbitary http header providing a username

friesoft:
Added field to External Backend to specify key for lookup in _SERVER array

Fixes #12137


friesoft:
Not needed anymore as the REMOTE_USER variable can be set quite easily (see ticket #12137)

PR #69, opened 2016-09-15
Fix links in about doc

zabacad:
Without this change, the links in the about doc are broken.


lippser:
Hi,

Thanks for the patch. I applied it to the master branch.

Best regards,
Eric

PR #71, opened 2016-10-20
Add FakeRequest file to fix tests

GenevievePlantin:
Icinga/Web/Url.php needs the class FakeRequest for test purpose
https://github.com/Icinga/icingaweb2/blob/master/library/Icinga/Web/Url.php#L142

### DIFF
--- a/library/Icinga/Cli/FakeRequest.php
+++ b/library/Icinga/Cli/FakeRequest.php
@@ -1,0 +1,8 @@
+<?php
+namespace Icinga\Cli;
+
+use Icinga\Web\Request;
+
+class FakeRequest extends Request
+{
+}


pboutin:
👍 we need this since `getRequest()` is broken in Cli mode without it.

jprivard:
Same here :+1:

ardeois:
@lippser or @nilmerg, maybe you can help merge this?

nilmerg:
I'll pass this on to the original author, since I doubt that the proposed solution is sufficient. (Why declaring a "FakeRequest" in the first place when it's not necessary?)

Just out of curiosity, what's the reason that four of you are confronted with this issue now? This bug is already for more than two years upstream. :D

ardeois:
@nilmerg haha you got us here, we're part of the same team actually.
Well we did this PR mostly because we run unit tests for the icingaweb2-director-module, and they fail if this file doesn't exist.

nilmerg:
Well, that explains a lot. ;) I've been informed that your tests will now run through successfully even without this fix.

I'll close this request therefore and we'll come up with a proper solution by ourselves. (https://dev.icinga.org/issues/7051)

PR #72, opened 2016-10-31
remove duplicate dot

hebbet:


nilmerg:
Hi,

there is rather a dot missing, I guess. It should instead read "Scheduling checks..." as at the time the notification is shown the issued operation may still be in progress.

Additionally, there are other forms also affected by this bug which you may want to adjust as well. :)

lippserd:
@nilmerg is right. It should be three instead of just two dots.

PR #73, opened 2016-12-22
Hook.php: Problem with module permissions

s1cc:
Noticed while using the graphite module. User had permissions 'module/graphite' but couldnt see graphs, only users with permission '*' could see graphs.
This function returns the module name as '\Graphite' but it should return 'graphite'. This is not a beautiful fix but works for me.
Maybe instead if this consider setting MODULE_PREFIX_LENGTH to 15 instead of 14? Then it would include the leading "\\". Dont know if this would break something else though.

### DIFF
--- a/library/Icinga/Application/Hook.php
+++ b/library/Icinga/Application/Hook.php
@@ -165,12 +165,12 @@ class Hook
         return lcfirst(
             substr(
                 $class,
-                ClassLoader::MODULE_PREFIX_LENGTH,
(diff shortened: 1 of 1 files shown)


lippserd:
Fixed w/ #2688. Thanks for your efforts though.

blessendor:
I still have a problem with graphite permissions in the icingaweb2 (2.6.1-1.jessie)

Current version of Hook.php:
https://pastebin.com/raw/iegNDnQv

And user permissions screen:
http://take.ms/j7NAy

PR #74, opened 2017-01-04
Add travis tests

lazyfrosch:


lazyfrosch:
Tests should be running fine now.

There are issues with 5.3:

    Parse error: syntax error, unexpected '[' in /home/travis/build/Icinga/icingaweb2/test/php/regression/Bug11831Test.php on line 33

And 7.1:

    Failed asserting that exception of type "ArgumentCountError" matches expected exception "\Icinga\Exception\NotWritableError". Message was: "Too few arguments to function 

Please review @lippser @nilmerg , @Thomas-Gelf

lazyfrosch:
> I pushed fixes for the failing tests. My eyes still hurt ;-) Travis is all green right now.

That what *automated* tests are for :wink:

> Thanks for the good work! There are traces of Icinga Director, are any tests referring to that module?

Oh, forgot to remove that after testing.

> And could we drop the composer part, as phpunit and phpcs are available as packages?

There is no good way to properly install them, and the latest version in a test environment like travis.

Same reason why I did it like this in director, standard procedure.

lazyfrosch:
@lippserd Tests are green.

I had to also fix locale setup for non UTF-8 systems. The Travis environment is iso-8859.

As we discussed, the config only builds tags, `master` and PR.

Will add phpcs, which will change the travis.yml a bit, so hold on

lazyfrosch:
I think we should exclude `phpcs` for now.

@lippserd pls review and merge

lazyfrosch:
Had a short discussion with @Thomas-Gelf 

It's actually weird that `setlocale` won't accept locales on Travis. The system seems ISO-8859-1 by default. When I generate `de_DE.UTF-8` it won't get used, my patch adds that Icingaweb2 tries to load an ISO-8859-1 locale.

Tom suggested that won't work like this.

lazyfrosch:
@nilmerg it's pretty slow and **lots** of errors and warnings, we should review that separately....

lazyfrosch:
@lippserd Rebased and updated commit message

lazyfrosch:
PHPUnit 6 seems to make us problems:

    PHP Fatal error:  Class 'PHPUnit_Framework_TestCase' not found in /home/travis/build/Icinga/icingaweb2/library/Icinga/Test/BaseTestCase.php on line 35

nilmerg:
@lippserd @Thomas-Gelf  Any chance to get you to review this? Some of the changes may be of use in another project ;)

Thomas-Gelf:
Hi @nilmerg!

First thoughts:
* can we get rid of Mockery?
* adding non-utf8 encodings is a no-go for me. Platform issues should be fixed on the platform, not by introducing dirty (and eventually dangerous) hacks in our code
* composer is not needed to get phpcs/phpunit running on Travis, but shipping composer.lock & similar suggests that Icinga Web 2 might be a composerized applications - which isnt' true at all
* please see `director` or `businessprocess` for examples on how to get phpunit 6.x "support"

Cheers,
Thomas

lazyfrosch:
Replaced composer, still not finished...

lazyfrosch:
By the way, the phpcs summary looks like:

```
php phpcs.phar -p --standard=PSR2 --extensions=php --encoding=utf-8 application/ library/Icinga modules/ test/
EEEEEE................W.....................................  60 / 735 (8%)
........E................................................... 120 / 735 (16%)
........................E.....................W.....WW.W..W. 180 / 735 (24%)
E.............W...WWW..WWWWWWWW..E...........W.W............ 240 / 735 (33%)
.....E..E....EE............................................. 300 / 735 (41%)
...........W................................................ 360 / 735 (49%)
...........................W................................ 420 / 735 (57%)
.........E..............W.....EEEEEEEEEEEE.............W.... 480 / 735 (65%)
......................................W..................... 540 / 735 (73%)
...........E..........WW.W.W.......W..W.WW....W....WW.W..WW. 600 / 735 (82%)
.W...WWWWW..EE.EW................................W.......... 660 / 735 (90%)
.....................WEEWWE.WE.WE..........EE.......WE.E..E. 720 / 735 (98%)
........E.....W
```

Thats why its disabled for now...

lazyfrosch:
So, I pinned down phpunit to 5.7/4.8, so we can use Mockery 0.9.9

Mockery 1.0.0-alpha would allow us to go higher in PHPUnit versioning, but I'm not sure about support for older PHP versions then.

Tests are green now. phpcs to be reviewed later...

@lippserd please review and merge!

PR #2692, opened 2017-01-18
Remove default settings from config.ini

### LINKED ISSUE #594: [dev.icinga.com #6080] Remove default settings from config.ini
This issue has been migrated from Redmine: https://dev.icinga.com/issues/6080

**Created by elippmann on 2014-04-29 10:13:32 +00:00**

Assignee: _jmeyer_
Status: _Feedback_ 
Target Version: _Backlog_
Last Update: _2017-01-13 11:06:04 +00:00 (in Redmine)_

---

As a user I do not want to see default settings in the config.ini because I want to keep my configuration readable, simple and I only want to know what I've actually changed. The web interface however must know of every configuration directive available, but that is not part of this issue. If I ever change a configuration value back to its default via the web interface, the configuration is removed from the config.ini.

**Changesets**

_2017-01-13 10:59:15 +00:00_ by _nhilverling_ 1623722e4d44113ee766e262da8b057e74403e6e

    Form: Implement attribute defaultValues
    
    refs #6080

_2017-01-13 10:59:27 +00:00_ by _nhilverling_ 07372d6e9ec09464f01926dd664427bfb2cef726

    Form: Implement form element attribute ignoreDefault
    
    refs #6080

_2017-01-13 10:59:27 +00:00_ by _nhilverling_ cba4bad29ada9f6ad7f4279923a4ee4ea15d298c

    GeneralConfigForm: Use element attribute ignoreDefault in createElements
    
    refs #6080

N-o-X:
fixes #594


application/forms/Config/UserBackend/LdapBackendForm.php
```diff
@@ -146,6 +146,7 @@ public function createElements(array $formData)
             'filter',
             array(
                 'preserveDefault'   => true,
+                'ignoreDefault'     => true,
```
nilmerg: I'm afraid but this is not a static default what's inserted here, but a suggestion that's otherwise thrown away and never used if 'ignoreDefault' is set to true.

lippserd:
Please add tests before merging.

nilmerg:
I'm not gonna merge this right now, I have just approved the requested changes without knowing that will approve the entire merge request :D

library/Icinga/Application/Config.php
```diff
@@ -350,11 +350,18 @@ public function saveIni($filePath = null, $fileMode = 0660)
             throw new LogicException('You need to pass $filePath or set a path using Config::setConfigFile()');
         }
 
-        if (! file_exists($filePath)) {
-            File::create($filePath, $fileMode);
-        }
+        $writer = $this->getIniWriter($filePath, $fileMode);
+        if ($writer->render() !== '') {
```
nilmerg: Since IniWriter::write() does also call IniWriter::render() we should probably cache what's been rendered to avoid doing it twice.

library/Icinga/Application/Config.php
```diff
@@ -350,11 +350,18 @@ public function saveIni($filePath = null, $fileMode = 0660)
...
+            if (! file_exists($filePath)) {
+                File::create($filePath, $fileMode);
+            }
 
-        $this->getIniWriter($filePath, $fileMode)->write();
+            $writer->write();
+        } else {
+            if (! file_exists($filePath)) {
```
nilmerg: I guess this should be the other way round? ;-)

lippserd:
What is the purpose of the ignoreDefault attribute?

nilmerg:
@lippserd You may want to take a look at the referenced issue #594

lippserd:
@nilmerg The issue neither explains the purpose of the ignoreDefault attribute nor does it state its requirement :) Do I have to add this attribute to trigger the remove default functionality? If so, why?

nilmerg:
@lippserd It's necessary to add this attribute because it's non-standard functionality. And since we don't have a global registry which default values are utilized in what parts of the code, we can neither make this functionality the default just for configuration forms.

lippserd:
Introducing another config attribute is not an option for me. We should postpone this feature.

PR #2725, opened 2017-02-08
TimezoneDetect: support also negative timezone offsets

### LINKED ISSUE #2716: icingaweb2 doesn't detect browser timezone
I have my timezone in icingaweb2 set to browser so it'll use my browser timezone settings, but it always defaults to UTC.

I checked on several sites to verify the browser timezone is correct and it shows America/Chicago which is correct.

Al2Klimov:
fixes #2716


lippserd:
Hi Alex,

I'd like to see some tests here.

Best regards,
Eric

Al2Klimov:
@lippserd done

Issue #2609, opened 2016-11-08
Provide control in the filter editor to update dashlets from their related views


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/13071

**Created by elippmann on 2016-11-08 14:42:59 +00:00**

At the moment we have two ways to update dashlets. It is possible to update a dashlet from the dashboard settings page. There, we don't have the possibility to use the filter editor to update the dashlet's URL and no preview. The other way is to delete the old dashlet and create a new one.
When navigating from a dashlet to its view, I suggest to provide a control to directly update the dashlet, if the view contains a filter editor.

PR #2734, opened 2017-02-09
Provide control in the filter editor to update dashlets from their re…

Al2Klimov:
…lated views

refs #2609


nilmerg:
The dashboards are currently re-worked. Closing here.

PR #2735, opened 2017-02-09
Correctly show whether a check source is reachable

### LINKED ISSUE #2524: [dev.icinga.com #12431] If CheckSource != Object, the icon is wrong when either is in a different state
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12431

**Created by TheFlyingCorpse on 2016-08-14 18:42:05 +00:00**

Assignee: _elippmann_
Status: _Feedback_ 
Target Version: _(none)_
Last Update: _2017-01-10 08:11:59 +00:00 (in Redmine)_

---

If CheckSource != Object, the icon is wrong when Object is in a different state than CheckSource (and the other way).

It doesnt account for it in the code here:

    icingaweb2/modules/monitoring/application/views/scripts/show/components/checksource.phtml

**Changesets**

_2016-12-19 16:18:14 +00:00_ by _(unknown)_ 7450dd5811a885e4a1e9d1b8d984d7fabecd7eef

    Correctly show whether a check source is reachable
    
    refs #12431

Al2Klimov:
fixes #2524

### DIFF
--- a/modules/monitoring/application/views/scripts/show/components/checksource.phtml
+++ b/modules/monitoring/application/views/scripts/show/components/checksource.phtml
@@ -5,9 +5,9 @@
     </th>
     <td>
         <?= $this->escape($object->check_source) ?>
-    <?php if ($object->is_reachable !== null): ?>
-        <span class="check-source-meta"><?= (bool) $object->is_reachable ? $this->translate('is reachable') : $this->translate('is not reachable') ?></span>
-        <?php if ((bool) $object->is_reachable) {
+    <?php if (isset($checkSource->is_reachable)): ?>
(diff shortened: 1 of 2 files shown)


dnsmichi:
Btw - check_source and is_reachable are not necessarily bound to each other.

`is_reachable` is determined by host/service dependencies in Icinga 2 and updated accordingly, if i.e. a parent fails. If there are no dependencies defined, it will always be "reachable". 

`check_source` is just the node name where the plugin was executed. 

I'd suggest to move Reachability into a separate line, away from the Check Source. This removes possible questions.

lippserd:
I close this in favor of #2953

lippserd:
I close this in favor of #2953

PR #2739, opened 2017-02-09
Icinga.Timer: use setTimeout() instead of setInterval()

### LINKED ISSUE #2407: [dev.icinga.com #11693] Icinga.Timer can lock up the browser
This issue has been migrated from Redmine: https://dev.icinga.com/issues/11693

**Created by haasn on 2016-04-27 15:35:20 +00:00**

Assignee: _elippmann_
Status: _Feedback_ 
Target Version: _Backlog_
Last Update: _2016-10-04 09:50:44 +00:00 (in Redmine)_

---

Icinga.Timer is implemented using setInterval, which is dangerous: If the self.tick() takes longer than the tick interval (250ms) to execute, it will keep stacking the workload higher and higher until the browser becomes completely unresponsive.

I suggest implementing Icinga.Timer like this instead:

initialize: function () {
var self = this;
var f = function() { self.tick(); setTimeout(f, self.interval); };
f();
},

**Changesets**

_2016-06-01 12:03:48 +00:00_ by _aklimov_ 313cd8ca9a581884f6023862ad07e0b1a5db78db

    Icinga.Timer: use setTimeout() instead of setInterval()
    
    refs #11693

Al2Klimov:
fixes #2407

### DIFF
--- a/public/js/icinga/timer.js
+++ b/public/js/icinga/timer.js
@@ -41,6 +41,8 @@
 
 
         this.lastRuntime = [];
+
+        this.isRunning = false;
     };
 
     Icinga.Timer.prototype = {
@@ -49,8 +51,16 @@
          * The initialization function starts our ticker
          */
         initialize: function () {
+            this.isRunning = true;
+
             var _this = this;
-            this.ticker = setInterval(function () { _this.tick(); }, this.interval);
+            var f = function () {
+                if (_this.isRunning) {
+                    _this.tick();
+                    setTimeout(f, _this.interval);
+                }
+            };
+            f();
         },
 
         /**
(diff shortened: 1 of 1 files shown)


public/js/icinga/timer.js
```diff
@@ -49,8 +51,16 @@
          * The initialization function starts our ticker
          */
         initialize: function () {
-            var _this = this;
-            this.ticker = setInterval(function () { _this.tick(); }, this.interval);
+            this.isRunning = true;
+
+            var that = this;
```
lippserd: Please use `var _this = this;`

public/js/icinga/timer.js
```diff
@@ -112,6 +122,7 @@
          * Our destroy function will clean up everything. Unused right now.
          */
         destroy: function () {
+            this.isRunning = false;
 
             if (this.ticker !== null) {
                 clearInterval(this.ticker);
```
lippserd: The code parts starting from line 127 are free to remove, right?

public/js/icinga/timer.js
Al2Klimov: done

public/js/icinga/timer.js
Al2Klimov: yes and done

PR #2752, opened 2017-02-14
ConfigForm: transformEmptyValuesToNull does not handle values correctly

### LINKED ISSUE #2751: ConfigForm: transformEmptyValuesToNull does not handle values correctly
transformEmptyValuesToNull should only set empty strings and empty arrays to null

N-o-X:
fixes #2751

### DIFF
--- a/application/forms/ConfigForm.php
+++ b/application/forms/ConfigForm.php
@@ -58,7 +58,7 @@ class ConfigForm extends Form
     {
         $sections = array();
         foreach ($this->getValues() as $sectionAndPropertyName => $value) {
-            if ($value === '' || (is_array($value) && empty($value))) {
+            if ($value === '' || $value === false || $value === []) {
                 $value = null; // Causes the config writer to unset it
             }
 
@@ -137,6 +137,8 @@ class ConfigForm extends Form
      */
     public static function transformEmptyValuesToNull(array $values)
     {
-        return array_map(function ($v) { return ($v === '' || (is_array($v) && empty($v))) ? null : $v; }, $values);
+        return array_map(function ($v) {
+            return ($v === '' || $v === false || $v === []) ? null : $v;
+        }, $values);
     }
 }
--- a/application/forms/ConfigForm.php
+++ b/application/forms/ConfigForm.php
@@ -58,7 +58,7 @@ class ConfigForm extends Form
     {
         $sections = array();
         foreach ($this->getValues() as $sectionAndPropertyName => $value) {
-            if (empty($value)) {
+            if ($value === '' || (is_array($value) && empty($value))) {
                 $value = null; // Causes the config writer to unset it
             }
 
@@ -137,8 +137,6 @@ class ConfigForm extends Form
      */
     public static function transformEmptyValuesToNull(array $values)
     {
-        return array_map(function ($v) {
-            return empty($v) ? null : $v;
-        }, $values);
+        return array_map(function ($v) { return ($v === '' || (is_array($v) && empty($v))) ? null : $v; }, $values);
     }
 }
--- a/application/forms/ConfigForm.php
+++ b/application/forms/ConfigForm.php
@@ -57,11 +57,7 @@ class ConfigForm extends Form
     public function onSuccess()
     {
         $sections = array();
-        foreach ($this->getValues() as $sectionAndPropertyName => $value) {
-            if ($value === '' || $value === false || $value === []) {
-                $value = null; // Causes the config writer to unset it
-            }
-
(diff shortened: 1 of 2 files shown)


application/forms/ConfigForm.php
```diff
@@ -58,7 +58,7 @@ public function onSuccess()
     {
         $sections = array();
         foreach ($this->getValues() as $sectionAndPropertyName => $value) {
-            if (empty($value)) {
+            if ($value === '' || (is_array($value) && empty($value))) {
```
nilmerg: Slightly different solution: `if ($value === '' || $value === false || $value === array()) {`
`false` gets converted to `''`

application/forms/ConfigForm.php
```diff
@@ -137,8 +137,6 @@ protected function writeConfig(Config $config)
      */
     public static function transformEmptyValuesToNull(array $values)
     {
-        return array_map(function ($v) {
-            return empty($v) ? null : $v;
-        }, $values);
+        return array_map(function ($v) { return ($v === '' || (is_array($v) && empty($v))) ? null : $v; }, $values);
```
nilmerg: Same as above, please check for `false` as well. And fix the codestyle, the previous style was already valid ;)

test/php/library/Icinga/Web/ConfigFormTest.php
```diff
@@ -0,0 +1,52 @@
...
+/* Icinga Web 2 | (c) 2017 Icinga Development Team | GPLv2+ */
+
+namespace Tests\Icinga\Web;
+
+use Icinga\Forms\ConfigForm;
+use Icinga\Test\BaseTestCase;
+
+class ConfigFormTest extends BaseTestCase
+{
+    public function testWhetherTransformEmptyValuesToNullHandlesValuesCorrectly()
```
nilmerg: Do you think it's possible to split these tests into a few more test cases? :P

test/php/library/Icinga/Web/ConfigFormTest.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        $values = array(
```
nilmerg: Mostly complete, but please add a case for `false` as well.

nilmerg:
We had a short discussion with @mxhash as there was a confusion whether the new minimum version (5.6) for exchange does also apply to Web 2. My last information was that we keep it at 5.3 for Web 2, but @N-o-X used Traits and other new features for a small Web 2 library component with the approval by @lippserd. I guess we need a clarification here. Any volunteers?

Thomas-Gelf:
@nilmerg: Icinga Web 2 itself is still 5.3 unless announced otherwise. Raising the minimum version requires quite some time for pre-announcement, packaging etc. Current plans are to require 5.6 the first time for the `icingadb` module, so we can use something new as an incentive for people to upgrade their systems.

In the meantime, Director 1.4 will start with a "first shot", requiring 5.4 - just to get a first feeling of how that would work out. With Director 1.3.2 being released at the same time with less features but 5.3 support to make everyone happy. We could use that announcement to suggest people to better already move to 5.6 (or 7.x).

While 5.4 only forces RHEL6-Users to switch to SCL, 5.6 isn't even the default on RHEL7. There is at least 7.0 in SCL, but our packaging probably needs to be prepared to work with those packages. This could involve switching to FPM per default - so expect quite some preparation work. Nothing complicated, just: it needs to be done.

Features targeting a new major Icinga Web 2 version might use modern PHP features, but they currently cannot find their way to the master. @lippserd: please correct me if you had other plans. In case we start working with non-5.3 feature-branches we should make sure our test-platform makes sure that no breaking code makes it to master.

lippserd:
@nilmerg I know we already had some discussions on this one but I'm not sure whether the related issue is still relevant. Please share your thoughts on this one.

nilmerg:
@lippserd This is still an issue if I'm not mistaken, as this is the cause why Icinga Web 2 will not save disabled checkboxes properly. (e.g. In the application settings -> disable stacktraces)

lippserd:
@nilmerg Why is `false` an empty value for us? If I set a config value to `false`, I expect it to be `false` not empty which gives me the default value.

Imo we should only test for the empty string and the empty array.

nilmerg:
@lippserd Can't remember the exact rationale behind this, but short tests showed that a checkbox element that is not checked and has the `uncheckedValue` option set to `false` has the empty string when calling Form::getValues(). Hence I may chose to intentionally add this case since our IniWriter will most likely convert `false` to an empty string as well. ('' == false)

Saving a checkbox element without this option set to false causes to save "0" as value in the ini file with this change. Which is the expected result.

lippserd:
@nilmerg Ok. Thanks for the feedback.

PR #2778, opened 2017-02-24
DbUserBackend: don't fail at validation if there aren't any active users

### LINKED ISSUE #2598: [dev.icinga.com #12977] Adding an empty user backend fails
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12977

**Created by twidhalm on 2016-10-25 08:43:06 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-10-25 08:43:06 +00:00 (in Redmine)_

---

Hi,

We have Icinga Web 2 with LDAP as authentication backend. Some time later we want local Icinga Web 2 users, too and add a MySQL database as a user backend. Validation of the empty backend fails and adding has to be forced

Steps to reproduce:

-   Set up Icinga Web 2 with LDAP as authentication backend
-   Create a MySQL Database and import `/usr/share/doc/icingaweb2/schema/mysql.schema.sql`
-   Add this database as a resource to Icinga Web 2
-   Add this resource as a user backend

Validation of the new backend fails. The detail area shows that the connection is successful but the check fails with the message that 0 active users exist.

When I already have a working backend and add a new one where I want to add my local users, having 0 active is what I'd expect. :-)

While this works when I enforce saving the new backend it would be nice if Icinga Web 2 recognized that there is already a working backend and the new one is a local one. So having 0 active users is not an error.

Al2Klimov:
fixes #2598

### DIFF
--- a/library/Icinga/Authentication/User/DbUserBackend.php
+++ b/library/Icinga/Authentication/User/DbUserBackend.php
@@ -285,12 +285,7 @@ class DbUserBackend extends DbRepository implements UserBackendInterface, Inspec
         $insp = new Inspection('Db User Backend');
         $insp->write($this->ds->inspect());
         try {
-            $users = $this->select()->where('is_active', true)->count();
-            if ($users > 0) {
-                $insp->write(sprintf('%s active users', $users));
-            } else {
-                return $insp->error('0 active users', $users);
-            }
+            $insp->write(sprintf('%s active users', $this->select()->where('is_active', true)->count()));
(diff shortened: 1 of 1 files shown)


Al2Klimov:
## The complex way

Check if there are any other backends configured and count them.

### Caveats

Is this a new backend or am I just editing...

## What I did

Just don't fail on 0 users.

### Why

Because if I'm able to create a backend, I'm already authn-ed.

PR #2798, opened 2017-03-17
Disable auto-refresh when page is not visible

### LINKED ISSUE #2761: Disable auto-refresh of inactive tabs
People here tend to have 1-10 Icinga tabs open, mostly in the background, so the tabs continue to poll the service in the background spamming some DB queries and causing unnecessary load.

Add Javascript based on `visibilityState` to suspend activity when the tab is not active. Grafana added this recently and that helped a lot.

There is no need to update the page when it is not visible.

towolf:
Icinga offers the user to enable or disable 'auto-refresh' with a
static preference.

But 'auto-refresh' is the default and our users often have dozens of
Icinga tabs open in their browser, which they are not looking at.

The background tabs lead to significant load on the database from
repetitive queries to keep the invisible UI fresh.

This PR adds a visibility listener, which disables auto-refresh when the
page is not visible.

Fixes #2761

### DIFF
--- a/public/js/icinga/events.js
+++ b/public/js/icinga/events.js
@@ -105,6 +105,8 @@
             // Note: It is important that this is the first handler for this event!
             $(document).on('rendered', { self: this }, this.applyHandlers);
 
+            $(document).on('visibilitychange', { self: this }, this.onVisibilityChange);
+
             $.each(this.icinga.behaviors, function (name, behavior) {
                 behavior.bind($(document));
             });
@@ -170,6 +172,18 @@
             var icinga = event.data.self.icinga;
             icinga.logger.info('Unloading Icinga');
             icinga.destroy();
+        },
+
+        onVisibilityChange: function (event) {
+            var icinga = event.data.self.icinga;
+
+            if (document.visibilityState === undefined || document.visibilityState === 'visible') {
+                icinga.loader.autorefreshSuspended = false;
+                icinga.logger.debug('Page visible, enabling auto-refresh');
+            } else {
+                icinga.loader.autorefreshSuspended = true;
+                icinga.logger.debug('Page invisible, disabling auto-refresh');
+            }
         },
 
         /**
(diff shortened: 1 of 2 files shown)


hloeffler:
LGTM.
Please merge.

Thomas-Gelf:
Related discussion in #2761

public/js/icinga/events.js
```diff
@@ -172,6 +174,18 @@
             icinga.destroy();
         },
 
+        onVisibilityChange: function (event) {
+            var icinga = event.data.self.icinga;
+
+            if !!(document.visibilityState === undefined || document.visibilityState === 'visible') {
```
lippserd: No need to add a double negation here. Syntax is wrong anyway.

public/js/icinga/events.js
```diff
@@ -172,6 +174,18 @@
...
+                icinga.logger.info('Page visible, enabling auto-refresh');
+                icinga.loader.enableAutorefresh()
```
lippserd: The call to `enableAutorefresh()` enables auto refresh even if it was disabled before using the console. We could introduce a new property `autorefreshSuspended` which is updated internally by the `onVisibilityChange` handler. The loader could then decide whether auto refresh is enabled by taking the properties `autorefreshEnabled` and `autorefreshSuspended` into account. I'm open to other suggestions though.

towolf: For me the use case "some guy fiddled with the console" too exotic to handle it.

If you want you can take it from here ...

towolf:
Sorry I don't have time to pursue this. Too much actual work.

Please continue. It should not be too tricky ...

lippserd:
Hi Feu,

Could you please have look at this PR and the requested changes?

Cheers,
Eric

theFeu:
Hi @towolf,
thanks for the input on the change!

Since we changed your code fundamentally we couldn't leave you as the author for it, sorry for that.
Instead we thanked you in the commit message, I hope you don't mind :)

Everything should work like you intended it.

KR,
Feu

towolf:
Thanks for getting this up to spec!  Please merge.

Issue #2824, opened 2017-04-24
Allow for hiding downtimestart and downtimeend events


leeclemens:
We have a lot of servers which do not change state very often, but have recurring scheduled downtimes. This adds a lot of clutter to the History tab, making it very difficult to pinpoint the most recent state change.

While a more robust solution may be to add the full filter functionality to the History tab, a quick link to hide/show may prove useful for others as well. Also, the filter would require configuration upon viewing each host/service, so this works more as a saved search/quick search.

lippserd:
Hi,

I updated the PR some time ago but did not comment this issue. It would be better to implement  a control which allows to filter for all event types. I'll close this issue.

Cheers,
Eric

PR #2825, opened 2017-04-24
Add quick link to filter DowntimeStart and DowntimeEnd historical events

leeclemens:
Fixes: #2824

### DIFF
--- a/modules/monitoring/application/views/scripts/partials/object/host-header.phtml
+++ b/modules/monitoring/application/views/scripts/partials/object/host-header.phtml
@@ -38,3 +38,39 @@ use Icinga\Module\Monitoring\Object\Host;
         </td>
     </tr>
 </table>
+<div class="quick-actions">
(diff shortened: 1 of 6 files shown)


lippserd:
Hi,

Thanks for the PR and sorry that it took me so long to give you feedback. I like the idea of having quick filters for the event types. But I think we should provide a control which allows to filter for all event types.
There is no need to introduce new queries. The eventhistoryquery supports the `type` column to filter for the following values:
* hard_state
* soft_state
* dt_start
* dt_end
* dt_comment
* dt_comment_deleted
* comment
* comment_deleted
* ack
* ack_deleted
* flapping
* flapping_deleted
* notify

The control should be a set of toggleable elements for:
* States
* Downtimes
* Comments
* Notifications
* Acknowledgements
* Flapping

If you still want to help here, I'll guide you.

Cheers,
Eric

PR #2845, opened 2017-05-19
GetTextTranslationHelper: Make sure we ignore any dotfile or vendor dir

lazyfrosch:
Some of our modules ship vendor data, or tend to add scripts adding icingaweb2 and other modules with a vendor/ directory.

This patch avoids the translation tool to scan these directories for translation calls.

### DIFF
--- a/modules/translation/library/Translation/Util/GettextTranslationHelper.php
+++ b/modules/translation/library/Translation/Util/GettextTranslationHelper.php
@@ -473,7 +473,7 @@ class GettextTranslationHelper
             $filepath = $directory . DIRECTORY_SEPARATOR . $filename;
             if (preg_match('@^[^\.].+\.(' . implode('|', $this->sourceExtensions) . ')$@', $filename)) {
                 $file->fwrite($filepath . PHP_EOL);
-            } elseif (is_dir($filepath)) {
+            } elseif (! is_link($filepath) && is_dir($filepath)) {
                 $subdirs[] = $filepath;
             }
         }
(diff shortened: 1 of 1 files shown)


lazyfrosch:
Closed it by accident...

lazyfrosch:
I'd really like to see this in 2.5.0 as it helps updating translations, when vendor data (like icingaweb2) is present within the source directory.

modules/translation/library/Translation/Util/GettextTranslationHelper.php
```diff
@@ -467,10 +467,13 @@ private function getSourceFileNames($directory, File $file)
 
         $subdirs = array();
         while (($filename = readdir($directoryHandle)) !== false) {
+            if (substr($filename, 0, 1) === '.' || $filename === 'vendor') {
```
lippserd: Use `$filename[0]` instead of `substr`.

modules/translation/library/Translation/Util/GettextTranslationHelper.php
```diff
@@ -467,10 +467,13 @@ private function getSourceFileNames($directory, File $file)
...
+                continue;
+            }
             $filepath = $directory . DIRECTORY_SEPARATOR . $filename;
             if (preg_match('@^[^\.].+\.(' . implode('|', $this->sourceExtensions) . ')$@', $filename)) {
                 $file->fwrite($filepath . PHP_EOL);
-            } elseif (is_dir($filepath) && !preg_match('@^(\.|\.\.)$@', $filename)) {
+            } elseif (!is_link($filepath) && is_dir($filepath)) {
```
lippserd: Please add a space after the `!`.

PR #2866, opened 2017-06-20
Wizard: show LDAP connection settings discovery failure reasons

### LINKED ISSUE #2865: Wizard: LDAP connection settings discovery failure reasons are silented

1. Wizard
2. Authentication Type: LDAP
3. LDAP Discovery: lolcat.icinga.com
4. press next

Al2Klimov:
fixes #2865

### DIFF
--- a/library/Icinga/Protocol/Ldap/LdapConnection.php
+++ b/library/Icinga/Protocol/Ldap/LdapConnection.php
@@ -154,6 +154,13 @@ class LdapConnection implements Selectable, Inspectable
     protected $discoverySuccess;
 
     /**
+     * The cause of the discovery's failure
+     *
+     * @var Exception|null
+     */
+    private $discoveryError;
+
+    /**
      * Whether the current connection is encrypted
      *
      * @var bool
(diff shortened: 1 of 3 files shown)


library/Icinga/Protocol/Ldap/LdapConnection.php
```diff
@@ -276,6 +285,16 @@ public function discoverySuccessful()
     }
 
     /**
+     * Get {@link discoveryError}
```
lippserd: Don't link private/protected properties in the public interface doc please.

modules/setup/application/forms/LdapDiscoveryPage.php
```diff
@@ -74,6 +74,8 @@ public function isValid($data)
                 $this->discovery = Discovery::discoverDomain($data['domain']);
                 if ($this->discovery->isSuccess()) {
                     return true;
+                } else {
+                    $this->getElement('domain')->addError($this->discovery->getError()->getMessage());
```
lippserd: The error message should not be bound to the domain field. It should be a form error instead.

lippserd:
LGTM. I hand this over to @nilmerg

modules/setup/application/forms/LdapDiscoveryPage.php
```diff
@@ -74,6 +74,8 @@ public function isValid($data)
                 $this->discovery = Discovery::discoverDomain($data['domain']);
                 if ($this->discovery->isSuccess()) {
                     return true;
+                } else {
+                    $this->addError($this->discovery->getError()->getMessage());
```
nilmerg: Please use Form::error instead since it's ours and generally used in the Wizard. Form::addError is from Zend and may differ in style.

PR #2881, opened 2017-07-04
Feature/allow modules to throw custom http exceptions 2880

### LINKED ISSUE #2880: Allow modules to throw custom HttpExceptions
Our ErrorController currently has all HTTP status codes hardcoded. This prevents any module from responding to non-API requests with other HttpExceptions than defined in Icinga\Exception\Http.

My proposal is to allow IcingaException to overwrite an exception's code, make HttpException non-abstract and add basic support for header definitions.

Throwing a custom HTTP exceptions then looks like the following:

```php
<?php

$error = new HttpException('%s temporarily unavailable', $applicationTitle);
$error->setHeaders(array(array('X-Status-Reason', 'Ongoing maintenance update')));
$error->setCode(503);
throw $error;
```

nilmerg:
resolves #2880

### DIFF
--- a/application/controllers/ErrorController.php
+++ b/application/controllers/ErrorController.php
@@ -6,9 +6,7 @@ namespace Icinga\Controllers;
 use Zend_Controller_Plugin_ErrorHandler;
 use Icinga\Application\Icinga;
 use Icinga\Application\Logger;
-use Icinga\Exception\Http\HttpBadRequestException;
-use Icinga\Exception\Http\HttpMethodNotAllowedException;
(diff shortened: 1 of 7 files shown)


nilmerg:
Hey Eric,

I didn't want to over-engineer this and I have to admit.. throwing a LogicException was just an excuse for this.

While I can't and won't understand what's bad about developers being able to overwrite status codes, your solution is really a good idea. But for the sake of a clean interface, I'll presume to keep method setHeader() in addition to setHeaders().

Best,
Johannes

lippserd:
lgtm

PR #2913, opened 2017-08-07
Setup Wizard: Fix error code on Debian; Docs: Revamp and split

dnsmichi:
- Error code 1698 is not handled
- Additional docs for required database/user for Debian
- Docs split into basic setup (packages), moved everything else into upgrading, advanced, troubleshooting

@lippserd The next Gin is on you :-P


dnsmichi:
Updated the MySQL bits and removed the URL checker, as discussed offline @lippserd :)

PR #2919, opened 2017-08-10
Config: Fix cached loading of module config

lazyfrosch:
Since the module copied the array to a local variable, the loaded
Config object never got saved to `static::$modules`

I noticed that during Unit testing, I wasn't able to play with module config in memory. Always had to save to disk...

### DIFF
--- a/library/Icinga/Application/Config.php
+++ b/library/Icinga/Application/Config.php
@@ -416,13 +416,12 @@ class Config implements Countable, Iterator, Selectable
             self::$modules[$modulename] = array();
         }
 
-        $moduleConfigs = self::$modules[$modulename];
-        if (! isset($moduleConfigs[$configname]) || $fromDisk) {
-            $moduleConfigs[$configname] = static::fromIni(
+        if (! isset(self::$modules[$modulename][$configname]) || $fromDisk) {
+            self::$modules[$modulename][$configname] = static::fromIni(
                 static::resolvePath('modules/' . $modulename . '/' . $configname . '.ini')
             );
         }
-        return $moduleConfigs[$configname];
+        return self::$modules[$modulename][$configname];
     }
 
     /**


library/Icinga/Application/Config.php
```diff
@@ -412,17 +412,16 @@ public static function app($configname = 'config', $fromDisk = false)
      */
     public static function module($modulename, $configname = 'config', $fromDisk = false)
     {
-        if (! isset(self::$modules[$modulename])) {
-            self::$modules[$modulename] = array();
+        if (! isset(static::$modules[$modulename])) {
+            static::$modules[$modulename] = array();
         }
 
-        $moduleConfigs = self::$modules[$modulename];
```
Thomas-Gelf: This line contains the original bug, an array is copied as such and not as a reference

library/Icinga/Application/Config.php
```diff
@@ -412,17 +412,16 @@ public static function app($configname = 'config', $fromDisk = false)
      */
     public static function module($modulename, $configname = 'config', $fromDisk = false)
     {
-        if (! isset(self::$modules[$modulename])) {
-            self::$modules[$modulename] = array();
+        if (! isset(static::$modules[$modulename])) {
+            static::$modules[$modulename] = array();
```
Thomas-Gelf: @lazyfrosch: when working with static properties, I'd suggest to use self::$property instead of static::$property. Motivation: there are late static method calls, but no late static properties. So, with B and C inheriting from A, B::$property equals C::$property, with $property defined in A - no matter whether you are going to use self or static. So, while replacing self with static has no effect in this commit, I'd opt for keeping self - as it doesn't suggest people not so deep into PHP a reality that doesn't exist. Namely, late static properties. At least not unless you break their reference in the child class.

lazyfrosch: @lippserd always told me to prefer `static`

Thomas-Gelf: I absolutely second this when it goes to method calls. Disagree for properties, not because it is wrong, but because it gives a developer extending your class the wrong feeling about having the possibility to work with his own $property, while instead sharing it with all other classes inheriting from the same base - unless they explicitly override the property by defining their own one. Instead when reading 'self' in the base class, he will hopefully think twice about what would be going on. Not saying that there aren't cases where static might absolutely be justified. Just, they break when the first one writes a class not re-declaring the very same property. So, my concerns here are about developers hurting themselves. @lippserd?

lazyfrosch: I'm only interested in fixing the caching issue.

Not going to nit pick, feel free to change my commit.

lazyfrosch:
Updated the commit to focus on the actual problem.

We should include this in the next release

library/Icinga/Application/Config.php
lippserd: Good point Tom :) I just thought use static so you don't have to wrap around your head whether to use self or static. But I like the rule of thumb to use self for props and consts and static for method calls and factory methods.

PR #2923, opened 2017-08-17
Form/Element/Number: The field should be valid when empty

lazyfrosch:
This failed me when trying to save the ResourceForm without having set a port number.

    is not a valid number

I guess because `$value !== ''` we should also test for null...


lazyfrosch:
The problem musst be introduced by a change since last release

lazyfrosch:
Btw. this also fails when resetting a value to its default (it is then null)

lazyfrosch:
@nilmerg can you have a look?

nilmerg:
LGTM :)

Issue #2792, opened 2017-03-14
isMemberAttributeAmbiguous() might struggle with strange DNs


xenuser:
I am running Icingaweb2 2.4.1 on SLES 12 with two AD backends.

One of the AD backends stores/returns the users the following way (when Icinga2 runs ldapsearches, e.g.):

```

CN=Hoebel\, Valentin (012-Somestring),OU=Users,OU=de,DC=some,DC=addirectory,DC=net


```

When viewing users and Groups from this AD within Icingaweb2, Icingaweb2 fails to resolve the group memberships of the relevant users. 

Now there is this function isMemberAttributeAmbiguous() within LdapUserGroupBackend.php which might struggle with this CN. When I temporarily modify this function and always return 0, the group memberships are resolved within Icingaweb2.

Can you guys please tell me if this is the expected behaviour (e.g. throw out objects/attributes with nasty content) or if there might be some mistake.

Either way, the current situation gave me no clue what went wrong. There were no errors in the logs, it just seems that Icingaweb2 stops the ldapqueries which normally are run when everything works fine (e.g. continue with resolving sAMAccountName attributes after receiving group memberships).

this might relate to #2146

tillsc:
I'm having the same problem here

lippserd:
I think #2924 will help here.

lippserd:
Hi guys,

Do you still have issues w/ the latest Web 2 version?

Best regards,
Eric

xenuser:
Hi Eric,

thanks for asking. Yes, the problem still exists.

lippserd:
Hi,

Thanks for the feedback.

@Al2Klimov Please have a look at this one.

Cheers,
Eric

Al2Klimov:
Hello @xenuser and thank you for reporting!

Please could you share these config files:
* `/etc/icingaweb2/authentication.ini`
* `/etc/icingaweb2/groups.ini`
* `/etc/icingaweb2/resources.ini`

Best,
AK

xenuser:
Hello @Al2Klimov ,

thanks for getting in touch.

Unfortunately, I can observe the bug on a customer system and I am not allowed to share their configs.

I will have a thought about how I can provide the requested data (e.g. replacing sensitive strings, such as customer names, with random strings etc.).

xenuser:
Just an idea. Might it help to provide some user and group DNs which will get obfuscated by me?

[...]

PR #2924, opened 2017-08-18
Problems in LDAP membership resolution

lazyfrosch:
This PR addresses issues in resolving DN and memberships under Active Directory, and other LDAP that use DN for membership information.

## Existing behavior

Icingaweb2 tries to detect if the membership attribute of groups is a DN.

The decision was based on comparing a found sample value to the configured base DN.

## Example

Let's say we have a LDAP base: `dc=germany,dc=example,dc=com`

We try to detect memberships, and by lucky chance retrieve a value like this:

    cn=techaccount123,ou=groups,dc=example,dc=com

This is of course a entity of another domain, but referenced here.

## The problem

Icingaweb2 compares that value, by expecting that `dc=germany,dc=example,dc=com` would be part of its domain. Which of course is not.

The assertion fails, and icingaweb2 does not resolv memberDN to username. The user can't be assigned to any group.

## Solution

This PR does:

* Adding a isDN check function
* Using that function to detect the membership type (deprecating the old logic)

Refs #2792

### DIFF
--- a/library/Icinga/Authentication/UserGroup/LdapUserGroupBackend.php
+++ b/library/Icinga/Authentication/UserGroup/LdapUserGroupBackend.php
@@ -10,6 +10,7 @@ use Icinga\Data\ConfigObject;
 use Icinga\Exception\ConfigurationError;
 use Icinga\Exception\ProgrammingError;
 use Icinga\Protocol\Ldap\LdapException;
+use Icinga\Protocol\Ldap\LdapUtils;
 use Icinga\Repository\LdapRepository;
 use Icinga\Repository\RepositoryQuery;
 use Icinga\User;
@@ -438,6 +439,11 @@ class LdapUserGroupBackend extends LdapRepository implements UserGroupBackendInt
     /**
      * Return whether the attribute name where to find a group's member holds ambiguous values
      *
+     * This tries to detect if the member attribute of groups contain:
+     *
+     *  full DN -> distinguished name of another object
+     *  other   -> ambiguous field referencing the member by userNameAttribute
+     *
      * @return  bool
      *
      * @throws  ProgrammingError    In case either $this->groupClass or $this->groupMemberAttribute
@@ -463,7 +469,8 @@ class LdapUserGroupBackend extends LdapRepository implements UserGroupBackendInt
                 ->setUnfoldAttribute($this->groupMemberAttribute)
                 ->setBase($this->groupBaseDn)
                 ->fetchOne();
-            $this->ambiguousMemberAttribute = !$this->isRelatedDn($sampleValue);
+
+            $this->ambiguousMemberAttribute = ! LdapUtils::isDn($sampleValue);
         }
 
         return $this->ambiguousMemberAttribute;
--- a/library/Icinga/Protocol/Ldap/LdapUtils.php
+++ b/library/Icinga/Protocol/Ldap/LdapUtils.php
@@ -19,8 +19,9 @@ class LdapUtils
      * UTF-8 chars like German umlauts would otherwise be escaped and shown
      * as backslash-prefixed hexcode-sequenzes.
      *
-     * @param  string  DN
-     * @param  boolean Returns 'type=value' when true and 'value' when false
+     * @param  string  $dn        DN
+     * @param  boolean $with_type Returns 'type=value' when true and 'value' when false
+     *
      * @return string
      */
     public static function explodeDN($dn, $with_type = true)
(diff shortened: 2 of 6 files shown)


lazyfrosch:
@lippserd @nilmerg Rather complex issue, let me know if anything is "ambiguous" :wink:

nilmerg:
Well, to me a "simple" name is ambiguous compared to a "distinguished" name :)

Anyway, I don't think a regular expression match is necessary here. Isn't this bug only affecting ActiveDirectory instances? What about hardcoding a check if it's a "msldap" type of backend?

lazyfrosch:
@nilmerg 
* I updated the crazy regexp
* This is not only AD, it affects every LDAP using DN member attributes

lazyfrosch:
Ok lets keep the ambiguous naming, I just updated the docs, because it was't clear for me what ambiguous means here.

library/Icinga/Authentication/UserGroup/LdapUserGroupBackend.php
```diff
@@ -463,7 +469,8 @@ protected function isMemberAttributeAmbiguous()
                 ->setUnfoldAttribute($this->groupMemberAttribute)
                 ->setBase($this->groupBaseDn)
                 ->fetchOne();
-            $this->ambiguousMemberAttribute = !$this->isRelatedDn($sampleValue);
+
+            $this->ambiguousMemberAttribute = !LdapUtils::isDn($sampleValue);
```
lippserd: Please add a space after the exclamation mark.

library/Icinga/Protocol/Ldap/LdapUtils.php
```diff
@@ -62,11 +64,42 @@ public static function implodeDN($parts)
...
+     *
+     * Note: This is a rather complex regex, open to better solutions!
+     *
+     * See test for examples.
+     *
+     * @param  mixed $value
+     *
+     * @return bool
+     */
+    public static function isDN($value)
```
lippserd: Please rename the function to `isDn`. I know that the other functions have a capital N but this is just lame. We call them with the lowercase n anyway. So it would make sense to rename the other functions as well. No bc here, PHP does not care about case sensitiveness when calling methods.
Please remove the note and reference to tests from the doc of the function.
Instead of implementing a custom logic for validating DNs which is prone to errors (no offence, just look at this [beast](https://github.com/zendframework/zend-ldap/blob/master/src/Dn.php#L588)) I would only check whether `ldap_dn2ufn` returns not false.

library/Icinga/Repository/LdapRepository.php
```diff
@@ -74,6 +74,9 @@ protected function getNormedAttribute($name)
      *
      * Will use the current connection's root DN if $baseDn is not given.
      *
+     * @deprecated This was only used by LdapUserGroupBackend::isMemberAttributeAmbiguous
```
lippserd: Deprecate the function and state that it will be removed with version 2.6.0.

test/php/library/Icinga/Protocol/Ldap/LdapUtilsTest.php
```diff
@@ -0,0 +1,41 @@
...
+        'testuser',
+        'heinzimüller',
+        'test.user@example.com',
+        'test,user@example.com',
+    );
+
+    public function testIsDnForValidValues()
+    {
+        foreach(static::$validDn as $dn) {
+            $this->assertTrue(LdapUtils::isDN($dn), 'DN should be tested as valid value: ' . $dn);
```
lippserd: Call `isDn()` here please and add a space after the foreach.

test/php/library/Icinga/Protocol/Ldap/LdapUtilsTest.php
```diff
@@ -0,0 +1,41 @@
...
+        }
+    }
+
+    public function testIsDnForInvalidValues()
+    {
+        foreach(static::$invalidDn as $dn) {
+            $this->assertFalse(LdapUtils::isDN($dn), 'DN should be tested as invalid value: ' . $dn);
```
lippserd: Call `isDn()` here please and add a space after the foreach.

lazyfrosch:
Updated to suggestions

lazyfrosch:
Seems like travis changed module behavior in precise, ldap is now compiled in...

Loading the module only in Precise for 5.3 then

lazyfrosch:
Should be fine now @lippserd

PR #2933, opened 2017-08-31
Update Csv.php

### LINKED ISSUE #2808: CSV export should create CSV files with ending .csv
I noticed that CSV exports via Icingaweb2 create files such as "services", but they don't have an ending (like ".csv").

This is ok for me, but some customers/users use Windows and as you know, Windows won't open Excel or another tool if you double click on a file without file ending.

I therefore would like to suggest that CSV exports come as ".csv".
In an ideal world, one would be able to configure if .csv is wanted or not.

Mikesch-mp:
Export CSV with file extension set to '.csv'. This would fix #2808


dnsmichi:
PR contains a commit from #2935. The three remaining commits should be squashed too :)

Mikesch-mp:
Will resend it in a minute

Issue #2941, opened 2017-09-05
Icinga Web 2: limit parameter is ignored in json output


flourish86:
When requesting items (services/hosts) as json, the `limit` parameter won't be applied. The response contains all matching items.

## Expected Behavior
If `limit` is given as a URL parameter in json output (`format=json`), the number of items should be limited accordingly.

## Current Behavior
Currently `limit` is completely ignored when output format is requested as `json` (`format=json`).

## Possible Solution
Apply `limit` parameter when set while `format` is set to `json`, ignore when not set and respond all items available.

## Steps to Reproduce (for bugs)
`limit`parameter works for regular output, e.g.
`http://localhost:8080/icingaweb2/monitoring/list/services?service_problem=1&dir=desc&limit=2&sort=service_severity`

it won't work, if `format` parameter is set to `json`, e.g.
http://localhost:8080/icingaweb2/monitoring/list/services?service_problem=1&dir=desc&limit=2&sort=service_severity&format=json

## Context
Trying to retrieve only a limited number of items by setting `limit`URL parameter didn't work as expected, when `format` is set to `json`, see _Expected Behavior_

## Your Environment
Icinga Web 2 vagrant box, commit 75b6d500

lippserd:
Hi Flo,

The `sort` and `page` parameter are ignored as well. I already created a new issue #2967. Closing this one.

Cheers,
Eric

PR #2944, opened 2017-09-06
Add limit parameter

nbuchwitz:
Added limit parameter to handleFormatRequest matching all 3 formats (sql, json, csv)

refs #2941


lippserd:
Hi Nicolai,

Thanks for the PR. I created a new issue #2967 for this topic because the parameters `page` and `sort` are ignored as well. I close your PR but help for the new issue would be much appreciated.

Best,
Eric

PR #2947, opened 2017-09-07
Add note about SELinux to CentOS install doc

rhopfer:


dnsmichi:
That should be `icingaweb2-selinux` right?

rhopfer:
Yes, sorry. I fixed it.

dnsmichi:
Thanks 👍

PR #2961, opened 2017-09-15
Option for default schedule downtime for all services

tfylling:
Adds support for an optional option 'hostdowntime_all_services' in the settings section of the monitoring options. If set to true, this will make the "All Services" checkbox in the schedule host downtime form default to checked.


lippserd:
Hi,

Thanks for the PR.

Cheers,
Eric

PR #2962, opened 2017-09-17
Enable PHP CodeSniffer in travis testing

lazyfrosch:
* [x] Enable phpcs test on every build
* [x] Update phpcs, except for PHP 5.3
* [x] Speed up travis builds by only requiring sudo on precise
* [x] Code cleanup and rule update
* [x] Update icingacli test commands

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: php
 dist: trusty
-sudo: required
+sudo: false
 
 php:
   - '5.4'
@@ -13,7 +13,10 @@ matrix:
   include:
   - php: '5.3'
     dist: precise
+    sudo: required
     env:
+    - PHPCS_VERSION=2.9.1
+    - LOCALE_GEN=1
     - ENABLE_LDAP=1
 
 services:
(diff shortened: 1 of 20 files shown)


lazyfrosch:
Hm the rule ref didn't work correctly...

lazyfrosch:
It's green<sup>TM</sup>

@lippserd please review, and merge?

lippserd:
Could you do me a big favour and move the top-level xml files to the test directory? I just don't see why they should pollute the root directory. Is phpcs happy if we add `/** @noinspection PhpMissingBreakStatementInspection */` above the offending case statements instead of adding a comment inside the cases?

lazyfrosch:
Updated the comments.

Seems like phpcs doesn't like the config the be in another location, it's always using relative paths to the dirname of the config...

lippserd:
Hi Markus,

Our icingacli test commands should consider those phpcs config files too. Could you also take care of that please?

Cheers,
Eric

lazyfrosch:
@lippserd updated the commits accordingly

We should fix https://github.com/Icinga/icinga-packaging/issues/50 for the 2.5.0 package release.

PR #2973, opened 2017-09-27
Update 02-Installation.md

Gr4cchus:
The getting started docs on the main website https://www.icinga.com/docs/icinga2/latest/doc/02-getting-started/ and here https://github.com/Icinga/icingaweb2/blob/master/doc/02-Installation.md do not include the mention of `icinga2-selinux` except outside of the quick setup scope in the section titled selinux(mentions just 1 of the 2 packages) which is required otherwise problems like this https://github.com/Icinga/icinga2/issues/5096 will develop; an example of trying to run "Check now" on a service via the web interface:
```
sudo grep icinga /var/log/audit/audit.log | audit2allow


#============= httpd_t ==============
allow httpd_t var_run_t:fifo_file { getattr open };
```

### DIFF
--- a/doc/02-Installation.md
+++ b/doc/02-Installation.md
@@ -139,7 +139,7 @@ apt-get install icingaweb2
 ```
 yum install icingaweb2 icingacli
 ```
-If you have [SELinux](90-SELinux.md) enabled, the package `icingaweb2-selinux` is also required.
+If you have [SELinux](90-SELinux.md) enabled, the package `icingaweb2-selinux` and `icinga2-selinux` is also required.
 For RHEL/CentOS please read the [package repositories notes](02-Installation.md#package-repositories-rhel-notes).
 
 **SLES and openSUSE**:


dnsmichi:
I tend to disagree here; SELinux for icinga2 is documented separately. The `icinga2-selinux` package is not required for example if you're running Icinga Web 2 on a separate host.

lippserd:
Hi,

Thanks for the PR. `icinga2-selinux` is part of the Icinga 2 installation and as Michi said, Web 2 and Icinga 2 can be installed on different hosts. So, I don't think that we should add this as requirement to the Web 2 docs.

But @dnsmichi maybe it is a good idea to add a line about selinux to https://github.com/Icinga/icinga2/blob/master/doc/02-getting-started.md#installing-icinga-2-
What do you think?

Cheers,
Eric

dnsmichi:
Oh, I thought it was documented there. Sorry my mistake. I'll take care, thanks.

Cheers,
Michael

dnsmichi:
Added and backported to support/2.7, now live on https://www.icinga.com/docs/icinga2/latest/doc/02-getting-started/#selinux

lippserd:
That was fast. Nice 👍

PR #2981, opened 2017-09-29
Fix not working string[string] type hint in code docs

### LINKED ISSUE #2980: Fix not working SomeType[string] in code docs

Al2Klimov:
fixes #2980


Al2Klimov:
1. it doesn't work
2. if it worked, it wouldn't give us much

PR #2983, opened 2017-09-29
Export CSV files with .csv extension

### LINKED ISSUE #2808: CSV export should create CSV files with ending .csv

Mikesch-mp:
Fixes #2808

### DIFF
--- a/modules/monitoring/library/Monitoring/Controller.php
+++ b/modules/monitoring/library/Monitoring/Controller.php
@@ -45,7 +45,13 @@ class Controller extends IcingaWebController
         }
         if ($this->_getParam('format') === 'csv'
             || $this->_request->getHeader('Accept') === 'text/csv') {
-            Csv::fromQuery($query)->dump();
+            $response = $this->getResponse();
+            $response
+                ->setHeader('Content-Type', 'text/csv')
+                ->setHeader('Cache-Control', 'no-store')
+                ->setHeader('Content-Disposition', 'attachment; filename=' . $this->getRequest()->getActionName() . '.csv')
+                ->appendBody((string) Csv::fromQuery($query))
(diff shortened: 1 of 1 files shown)


lippserd:
Hi Carsten,

Thanks for the update. I wanted to fix the last bits and squash your commits but unfortunately I'm not allowed to push to your branch. Could you please change the code block to the following:

```php
        if ($this->_getParam('format') === 'csv'
            || $this->_request->getHeader('Accept') === 'text/csv') {
            $response = $this->getResponse();
            $response
                ->setHeader('Content-Type', 'text/csv')
                ->setHeader('Cache-Control', 'no-proxy')
                ->setHeader('Content-Disposition', 'attachment; filename=' . $this->getRequest()->getActionName() . '.csv')
                ->appendBody((string) Csv::fromQuery($query))
                ->sendResponse();
            exit;
        }
```

I changed the indent and the call to Csv::dump().

If you don't want to mess with git rebase just do the following:

* Checkout your branch
* Run `git reset --hard d11775a6b8f6483b18f4e9875cdc258f55a8da9e`
* Change the code
* Run `git commit -a` and use the following commit message
> Export CSV files with .csv extension
>
> refs #2808
* Run `git push -f`

Cheers,
Eric

Mikesch-mp:
HI Eric,

did it and many thanks for the git tipps :)

Grüße,
Carsten

PR #2999, opened 2017-10-06
LdapBackendForm: move discoveries from createElements() to isValidPartial()

### LINKED ISSUE #2997: Wizard / authentication backend: "discover" doesn't overwrite base DN

Al2Klimov:
fixes #2997

### DIFF
--- a/application/forms/Config/UserBackend/LdapBackendForm.php
+++ b/application/forms/Config/UserBackend/LdapBackendForm.php
@@ -116,6 +116,18 @@ class LdapBackendForm extends Form
             );
         }
 
+        if ($isAd) {
+            // ActiveDirectory defaults
+            $userClass = 'user';
+            $filter = '!(objectClass=computer)';
+            $userNameAttribute = 'sAMAccountName';
+        } else {
+            // OpenLDAP defaults
+            $userClass = 'inetOrgPerson';
+            $filter = null;
+            $userNameAttribute = 'uid';
+        }
+
         $this->addElement(
             'text',
             'user_class',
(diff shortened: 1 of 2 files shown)


Al2Klimov:
@nilmerg I think that's what you requested (offline) – and it fixes #2997 OOTB.

application/forms/Config/UserBackend/LdapBackendForm.php
```diff
@@ -265,38 +242,92 @@ public function createElements(array $formData)
...
+                $filter = '!(objectClass=computer)';
+                $userNameAttribute = 'sAMAccountName';
+            } else {
+                // OpenLDAP defaults
+                $userClass = 'inetOrgPerson';
+                $filter = null;
+                $userNameAttribute = 'uid';
+            }
+
+            $formData['user_class'] = $userClass;
```
nilmerg: @Al2Klimov Please change $formData['filter'] and $formData['base_dn'] only if the user did not enter anything yet. The filter is currently only a hardcoded suggestion for AD and therefore only applicable if the user did not choose one manually yet. The base_dn manually entered by the user is probably more correct especially due to the fact that there is actually something entered.

application/forms/Config/UserBackend/LdapBackendForm.php
```diff
@@ -108,35 +114,6 @@ public function createElements(array $formData)
...
-                } catch (Exception $e) {
-                    $this->warning(sprintf(
-                        $this->translate('Failed to discover the chosen LDAP connection: %s'),
-                        $e->getMessage()
-                    ));
-                }
-            }
-        }
-
-        if ($isAd || $hasAdOid) {
```
nilmerg: @Al2Klimov Removing this block breaks re-population of default configurations. Especially in case of AD backends. (Try opening a AD backend configuration in Icinga Web 2)

nilmerg:
LGTM, but please check if the wizard is still behaving as desired.

PR #3002, opened 2017-10-09
Keep HTTP connections to the Icinga 2 API alive

### LINKED ISSUE #3001: Keep HTTP connections to the Icinga 2 API alive
This would save us a lot of expensive TLS handshakes.

Al2Klimov:
fixes #3001

### DIFF
--- a/modules/monitoring/library/Monitoring/Web/Rest/RestRequest.php
+++ b/modules/monitoring/library/Monitoring/Web/Rest/RestRequest.php
@@ -14,6 +14,13 @@ use Icinga\Module\Monitoring\Exception\CurlException;
 class RestRequest
 {
     /**
+     * cURL handle wrapper (for persistent connections)
+     *
+     * @var Curl
+     */
+    protected static $curl;
+
+    /**
      * Request URI
      *
      * @var string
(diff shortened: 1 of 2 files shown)


modules/monitoring/library/Monitoring/Web/Rest/RestRequest.php
```diff
@@ -209,6 +216,7 @@ public function send()
             "Host: {$url['host']}",
             "Content-Type: {$this->contentType}",
             'Accept: application/json',
+            'Connection: keep-alive',
```
dnsmichi: Imho this shouldn't be hardcoded in this generic class, but left to the caller (e.g. the command transport).

Al2Klimov: @dnsmichi Some other headers *are* already hardcoded – so what.

dnsmichi: It is a behavioural change. I wouldn't go with that argument of other default headers also being hardcoded but rather make it optional.

lippserd:
We should wait for the release of `ipl-http` and use this lib instead of the `RestRequest`.

PR #3009, opened 2017-10-10
icingaweb-module-monitoring: depend on php-xml

### LINKED ISSUE #3008: Missing dependency php-xml on Ubuntu 16.04

Al2Klimov:
fixes #3008

### DIFF
--- a/packages/debian/control
+++ b/packages/debian/control
@@ -28,7 +28,7 @@ Description: Icingaweb documentation module
 
 Package: icingaweb-module-monitoring
 Architecture: any
-Depends: icingaweb-common
+Depends: icingaweb-common, php-xml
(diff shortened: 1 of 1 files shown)


Al2Klimov:
@lazyfrosch Unfortunately I could not perform a before-after test for 297e8ae as I failed to build packages from source.

lazyfrosch:
The package source in this directory is very very outdated.

What do we need XML for?

Correct place to update would be https://github.com/Icinga/icinga-packaging/tree/deb/snapshot/icingaweb2

Al2Klimov:
Thank you for the repo.

No, we don't need XML, but DOM which is part of the XML package.

Issue #2509, opened 2016-08-03
[dev.icinga.com #12295] group_name_attribute should be "sAMAccountName" by default


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12295

**Created by mwaldmueller on 2016-08-03 06:21:48 +00:00**

Assignee: _jmeyer_
Status: _Feedback_ 
Target Version: _Backlog_
Last Update: _2017-01-11 10:07:37 +00:00 (in Redmine)_

---

group\_name\_attribute should be "sAMAccountName" by default not "gid"

**Changesets**

_2016-12-20 10:26:22 +00:00_ by _nhilverling_ 23f7544a4c84b5836c3233134f4ea3abf788faf6

    Use placeholder for elements instead of ignore & required
    
    refs #12295

_2016-12-20 10:33:46 +00:00_ by _nhilverling_ 69eaaec512cecca2782e6a2bf4788eeea2590d14

    Use default values while validating user backend
    
    refs #12295

_2017-01-09 15:56:18 +00:00_ by _nhilverling_ 1129a29ef7afbb36bb0660914f5c1842c5da0799

    Use default values while validating user backend
    
    refs #12295

_2017-01-09 16:03:53 +00:00_ by _nhilverling_ 277f4b0ddf7d0f495e463bcff87eb6cbd65419a5

    Form: Return values as null if element has "ignore" and "preserveDefault"
    
    refs #12295

_2017-01-09 16:06:26 +00:00_ by _nhilverling_ 594f2485e3c814494ed4393cf750ea414c409eac

    Form: Unset defaults if element has attribute "disabled" in preserveDefaults
    
    refs #12295

nilmerg:
Will be fixed once #594 is resolved.

PR #3010, opened 2017-10-10
Bugfix/cant create resources and command transports 2998

### LINKED ISSUE #2998: Can't create resources and command transports which require a port
It's not possible to create a new Command Transport for the monitoring module using the Icinga 2 API via GUI:

bildschirmfoto 2017-10-06 um 15 39 16

bildschirmfoto 2017-10-06 um 11 27 02

## Environment
* Icinga Web 2 version and modules (System - About): 2.4.2
* Version used (`icinga2 --version`): r2.7.1-1
* Operating System and version: CentOS Linux release 7.4.1708 (Core)

nilmerg:
Note that this change heavily relies on the assumption that populate() was necessary to get isSubmitted() to work. In case the call to Form::populate() is required due to additional reasons this change will most likely break them. However, I couldn't find any other reason and tests (not exhaustive <- only Web 2 and core modules) showed no errors.

The most important change is that at the time isSubmitted() is called the form may not be already created anymore. Any form relying on this case (as shown in d118b5c) may not work properly anymore. This is due to an already wrong implementation, though.

fixes #2998
refs #2509

### DIFF
--- a/library/Icinga/Web/Form.php
+++ b/library/Icinga/Web/Form.php
@@ -19,18 +19,6 @@ use Icinga\Web\Form\Element\CsrfCounterMeasure;
 /**
  * Base class for forms providing CSRF protection, confirmation logic and auto submission
  *
- * @method $this setDefaults(array $defaults) {
- *     Use `Form::populate()' for setting default values for elements instead because `Form::setDefaults()' does not
- *     create the form via `Form::create()'.
- *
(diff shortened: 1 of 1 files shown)


nilmerg:
@lippserd Pushed alternative fix.

lippserd:
Thanks!

PR #3020, opened 2017-10-13
[WIP] Abstract storage via buckets containing items

Al2Klimov:


Al2Klimov:
As discussed w/ @lippserd (offline): too abstract and too complicated

PR #3025, opened 2017-10-17
Implement toggleable containers

### LINKED ISSUE #3024: Toggleable flyover containers
## Abstract

Every toggleable flyover container starts with a container element with the CSS class `flyover-container` which has a unique ID per column. It contains at least one clickable element with the CSS class `flyover-toggle` and exactly one container element with the CSS class `flyover-content`. The `flyover-content` element is empty and not being displayed. Once any of the `flyover-toggle` elements is being clicked, the CSS class `flyover-collapsed` is being added to or removed from the `flyover-content` element.

If it's removed, the content is being displayed and the URL in its `data-flyover-href` is loaded. If it's added, the content is being undisplayed and the loaded content is destroyed.

```html
<div class="flyover-container" id="graphsrange-customdatetimepicker">
  <button class="flyover-toggle"></button>
  <div class="flyover-content flyover-collapsed" data-flyover-href="/icingaweb2/graphite/graphsrange-customdatetimepicker"></div>
</div>
```

```css
.flyover-content .flyover-collapsed {
  display: none;
}
```

## Auto-Refresh

If a column refreshes, all its open flyovers are backed up and restored once the render event is fired.

Al2Klimov:
fixes #3024

### DIFF
--- a/library/Icinga/Web/JavaScript.php
+++ b/library/Icinga/Web/JavaScript.php
@@ -26,6 +26,7 @@ class JavaScript
         'js/icinga/behavior/autofocus.js',
         'js/icinga/behavior/tooltip.js',
         'js/icinga/behavior/sparkline.js',
+        'js/icinga/behavior/subcontainer.js',
         'js/icinga/behavior/tristate.js',
         'js/icinga/behavior/dropdown.js',
         'js/icinga/behavior/navigation.js',
--- a/public/js/icinga/behavior/subcontainer.js
+++ b/public/js/icinga/behavior/subcontainer.js
@@ -11,34 +11,18 @@
 
     var columnContainerIdPattern = /^col/;
 
-    function SubContainer(icinga) {
-        Icinga.EventListener.call(this, icinga);
-
-        this.on('rendered', this.onRendered, this);
+    function onRenderedCollapsible(event) {
+        backupSubContainer($(event.target));
     }
 
-    SubContainer.prototype = Object.create(Icinga.EventListener.prototype);
-
-    SubContainer.prototype.onRendered = function(e) {
-        if ($(e.target).hasClass("collapsible")) {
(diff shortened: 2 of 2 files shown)


Al2Klimov:
@lippserd I'm afraid I can't replace .collapsible w/ .container (as discussed offline) as the rendered event already gets fired for .container and we have to difference between .container and .collapsible.

Al2Klimov:
In addition I'm afraid that the last commit works (you were right), but it doesn't solve the memory leak. (Very bad for people like me who almost never turn off their WS.)

Al2Klimov:
In addition I'm afraid that auto-refreshing of the toggled container requires . collapsible s to have (also?) .container, but – again – the rendered event...

Also we already discovered that every such container seems to has to have data-icinga-refresh PRE-set. https://github.com/Icinga/icingaweb2-module-graphite/pull/49/commits/e750c302dd8c1dc18fe2a36c2d5bfaf3eaf2723f#diff-68db69d79c61179557c6557107b02d3cL51 (Goodbye setAutorefreshInterval()...)

lippserd:
Hi Alex,

We're getting closer and closer. Could you please use `closest()` instead of `parents(...).first()` and `find('foo:first')` instead of `find('foo').first()`?

Any chance to not require one to add the CSS class `collapsed` to `subcontainer-content`?

Is there a branch in graphite module repository where I can play with an example? I'd like to see a real use case.

Cheers,
Eric

Al2Klimov:
"Is there a branch (...)" – I'll make one. (I didn't make any because we didn't have a "stable" CSS interface, but now we almost have one...)

Al2Klimov:
@lippserd Here you are: https://github.com/Icinga/icingaweb2-module-graphite/tree/feature/use-togglable-subcontainers-44

nilmerg:
Don't know what this is about, but #3024 has been closed by another PR and the flyout behavior got removed with the last update.

PR #3026, opened 2017-10-17
Fix php fatal error on monitoring/health/info when asking for json

jrgensen:
When calling /icingaweb2/monitoring/health/info?format=json you get a php error because we ask for query on query object.

<b>Fatal error</b>:  Uncaught Error: Call to undefined method Icinga\Module\Monitoring\Backend\Ido\Query\ProgramstatusQuery::getQuery() in /usr/share/icingaweb2/modules/monitoring/library/Monitoring/Controller.php:49

This PR fixes this


lippserd:
Hi,

Thanks for the PR 👍 

Cheers,
Eric

PR #3027, opened 2017-10-17
Respect limits and sort orders for format requests

### LINKED ISSUE #2967: Limit, page and sort URL parameters ignored if the format parameter is used
When requesting services, hosts and groups in a format other than HTML, the parameters `limit`, `page` and `sort` won't be respected.

## Expected Behavior

If any of the above parameters is given in the URL and the `format` parameter is set to `json`, `csv` or `sql`, the items in the output must be affected accordingly. If the `limit` parameter is not set, we must retain the old behaviour and return all items nonetheless.

## Current Behavior

All of the above parameters are ignored. All versions of Web 2 are affected.

## Possible Solution

Call `ListController::filterQuery()` after the pagination, limit and sort control have been set up but only apply the default limit if the format is HTML.

`limit`parameter works for regular output, e.g.
`/icingaweb2/monitoring/list/services?service_problem=1&dir=desc&limit=2&sort=service_severity`

It won't work if `format` parameter is set to `json` for example, e.g.
`/icingaweb2/monitoring/list/services?service_problem=1&dir=desc&limit=2&sort=service_severity&format=json`

nilmerg:
If anyone wants a more sophisticated fix I'll happily refactor our controller mess :grin: 

fixes #2967

### DIFF
--- a/modules/monitoring/library/Monitoring/Controller.php
+++ b/modules/monitoring/library/Monitoring/Controller.php
@@ -32,33 +32,44 @@ class Controller extends IcingaWebController
 
     protected function handleFormatRequest($query)
     {
-        if ($this->_getParam('format') === 'sql') {
-            echo '<pre>'
-                . htmlspecialchars(wordwrap($query->dump()))
-                . '</pre>';
-            exit;
+        $desiredContentType = $this->getRequest()->getHeader('Accept');
+        if ($desiredContentType === 'application/json') {
+            $desiredFormat = 'json';
+        } elseif ($desiredContentType === 'text/csv') {
+            $desiredFormat = 'csv';
+        } else {
+            $desiredFormat = strtolower($this->params->get('format', 'html'));
         }
-        if ($this->_getParam('format') === 'json'
-            || $this->_request->getHeader('Accept') === 'application/json') {
-            $response = $this->getResponse();
-            $response
-                ->setHeader('Content-Type', 'application/json')
-                ->setHeader('Cache-Control', 'no-store')
-                ->setHeader('Content-Disposition', 'inline; filename=' . $this->getRequest()->getActionName() . '.json')
-                ->appendBody(Json::encode($query->getQuery()->fetchAll()))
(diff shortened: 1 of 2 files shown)


modules/monitoring/library/Monitoring/Controller.php
```diff
@@ -31,22 +31,32 @@ protected function moduleInit()
 
     protected function handleFormatRequest($query)
     {
-        if ($this->_getParam('format') === 'sql') {
-            echo '<pre>'
-                . htmlspecialchars(wordwrap($query->dump()))
-                . '</pre>';
-            exit;
+        if ($this->getRequest()->getHeader('Accept') === 'application/json') {
```
nilmerg: Don't call getHeader multiple times

modules/monitoring/library/Monitoring/Controller.php
```diff
@@ -31,22 +31,32 @@ protected function moduleInit()
...
+            $desiredFormat = 'json';
+        } elseif ($this->getRequest()->getHeader('Accept') === 'text/csv') {
+            $desiredFormat = 'csv';
+        } else {
+            $desiredFormat = strtolower($this->_getParam('format'));
```
nilmerg: Use $this->params instead

modules/monitoring/library/Monitoring/Controller.php
```diff
@@ -31,22 +31,32 @@ protected function moduleInit()
...
         }
-        if ($this->_getParam('format') === 'json'
-            || $this->_request->getHeader('Accept') === 'application/json') {
-            header('Content-type: application/json');
-            echo json_encode($query->getQuery()->fetchAll());
-            exit;
+
+        if ($desiredFormat && $desiredFormat !== 'html' && ! $this->getRequest()->has('limit')) {
```
nilmerg: Use $this->params instead and use 'html' as default when fetching the format parameter above

nilmerg:
@lippserd Refactored as discussed.

PR #3031, opened 2017-10-18
Vagrantfile: One box for all providers plus local customization

lippserd:
This PR introduces the following changes to our Vagrantfile:
* All providers use the box `boxcutter/centos7` which supports all our used providers and is not tied to a minor path like the bento boxes
* Devs may customize the Vagratfile locally via .Vagrantfile.local


lippserd:
@nilmerg Could you please verify whether the new box works for vmware?

PR #3038, opened 2017-10-22
Doc: Fix module name in translation example

nbuchwitz:
Changed module name from ``module`` to ``yourmodule`` according to the example description.


nilmerg:
Hey @nbuchwitz,

thanks for the correction, but shouldn't you have exchanged `development` instead? Because `module` is part of the command :)

nbuchwitz:
Oops, I'corrected it and will squash both commits, when I'm at the office.

lippserd:
Hi Nicolai,

Thanks for the PR.

Cheers,
Eric

PR #3041, opened 2017-10-23
Implement auto-detaching DOM elements

### LINKED ISSUE #3039: Auto-detaching DOM elements
... which are attached back on autorefresh.

Al2Klimov:
fixes #3039

### DIFF
--- a/library/Icinga/Web/JavaScript.php
+++ b/library/Icinga/Web/JavaScript.php
@@ -24,6 +24,7 @@ class JavaScript
         'js/icinga/timezone.js',
         'js/icinga/behavior/application-state.js',
         'js/icinga/behavior/autofocus.js',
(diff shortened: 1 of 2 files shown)


lippserd:
Hi Alex,

I'd like to see the following changes:

* Don't use `Object.create()`
* Rely on IDs. Don't use `getCSSPath()`.
* Use `find(selector:first)` instead of `find(selector).first()`

Why are you calling `$container.parent()`?

Cheers,
Eric

Al2Klimov:
@lippserd "Why are you calling $container.parent()?" – because I was using `getCSSPath()`.

Al2Klimov:
@lippserd: You asked for a test scenario – here you are: https://github.com/Icinga/icingaweb2-module-graphite/issues/72

PR #3047, opened 2017-10-23
Resource config forms: centralize name element

### LINKED ISSUE #3046: Resource config forms: centralize name element
## Current Behavior

Each resource config form defines the name element by itself.

## Expected Behavior

A parent form of a resource config form defines that element.

## Reasons

* every resource has a name by definition
* that name is the same "kind of" name for all resource types
* developers of future resource types (e.g. #2790, #3035) have less to type
* developers of current resource types have less to change (in case there's something to change)

Al2Klimov:
fixes #3046

### DIFF
--- a/application/forms/Config/Resource/DbResourceForm.php
+++ b/application/forms/Config/Resource/DbResourceForm.php
@@ -74,15 +74,6 @@ class DbResourceForm extends Form
         }
 
         $this->addElement(
-            'text',
-            'name',
-            array(
-                'required'      => true,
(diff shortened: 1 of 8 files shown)


Al2Klimov:
As discussed w/ @nilmerg (offline): We may use traits (PHP 5.6) in 2.5 and 2.5 is the next release.

In this case a trait is "the only clean way" as we have more than one "parent form" (see issue).

Al2Klimov:
see issue

Issue #562, opened 2014-04-06
[dev.icinga.com #5949] cli --format output broken


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/5949

**Created by mfriedrich on 2014-04-06 16:42:45 +00:00**

Assignee: _(none)_
Status: _Closed_  (closed on _2014-04-29 15:27:05 +00:00_)
Target Version: _(none)_
Last Update: _2014-04-29 15:27:05 +00:00 (in Redmine)_

---

    [root@icinga2a icinga2]# icingacli monitoring list --unhandled --format '$host$: $service$'
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$
    $host$: $service$

dnsmichi:
Unfortunately this was never documented, so it bit me again during today's training. I'll prepare a PR which updates the CLI command help output, also for `--unhandled` -> `--problems`.

PR #3052, opened 2017-10-24
Fix help for icingacli monitoring list (--unhandled is --problems, etc.)

dnsmichi:
The help text still showed the wrong paramter, while it was not documented
as possible parameter either. In addition to that #562 unveils that a
custom --format parameter requires selected columns to do so.

This patch adds that into the documentation which is rendered as CLI command
help text then.

refs #562


dnsmichi:
@lippserd please consider merging this for 2.5, this has been sitting there too long. Seen in today's Icinga 2 training.

Issue #2954, opened 2017-09-08
Use PHP's "new" password hashing API and "stronger" algorithm


lippserd:
As discussed in #2436 and #2738 we have to replace MD5-Crypt w/ something "stronger". We should do this once we raise our PHP dependency to 5.6. This allows us to use the new functions related to password hashing which have been introduced w/ PHP 5.5. The default there is bcrypt which I would stick to then.

Al2Klimov:
FYI:
"Um (...) Angriffe zu verhindern, sollten Webseiten-Betreiber die Passwörter ihrer Nutzer mit Verfahren wie bcrypt (...) speichern. Diese sind darauf ausgelegt, dass ein Angriff möglichst viel Ressourcen benötigt (...)." – c't 20/17 S. 43 Abs. 2

Translation from German via Google Translate (verified by me):
"To prevent (...) attacks, web site operators should store the passwords of their users using methods like bcrypt (...). These are designed so that an attack requires as much resources as possible (...)."

PR #3061, opened 2017-11-06
Modify authentication function to support alternative algorithms

ss23:
The existing usage of crypt() was borderline incorrect. This simplified function will allow hashes of other types (e.g. bcrypt) and thus mitigate #2954 (use password_hash) until this can be implemented.

The getSalt protected method was also removed as this is no longer required, though this can be added again in future.

I will make an additional pull request to migrate to bcrypt with crypt.

### DIFF
--- a/library/Icinga/Authentication/User/DbUserBackend.php
+++ b/library/Icinga/Authentication/User/DbUserBackend.php
@@ -225,9 +225,7 @@ class DbUserBackend extends DbRepository implements UserBackendInterface, Inspec
     {
         try {
             $passwordHash = $this->getPasswordHash($user->getUsername());
-            $passwordSalt = $this->getSalt($passwordHash);
-            $hashToCompare = $this->hashPassword($password, $passwordSalt);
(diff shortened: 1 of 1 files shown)


library/Icinga/Authentication/User/DbUserBackend.php
```diff
@@ -225,9 +225,7 @@ public function authenticate(User $user, $password)
     {
         try {
             $passwordHash = $this->getPasswordHash($user->getUsername());
-            $passwordSalt = $this->getSalt($passwordHash);
-            $hashToCompare = $this->hashPassword($password, $passwordSalt);
-            return $hashToCompare === $passwordHash;
+            return (crypt($password, $passwordHash) === $passwordHash);
```
lippserd: Please remove the brackets around the statement.

ss23:
@lippserd Changes have been pushed :smile:

lippserd:
That's a commit pace I can work with :D

PR #3063, opened 2017-11-06
Improve Password hashing and validation for DbUserBackend

### LINKED ISSUE #2954: Use PHP's "new" password hashing API and "stronger" algorithm

ss23:
Authentication will continue to work as before and simply migrate to a new method once users change their password.

This is another small step before #2954 is implemented, but provides the same level of security (the parameters for this change are the same that would be used with password_hash).

---
Passwords are now handled via `PasswordHelper`, that is able to fallback to crypt() with SHA-512, when PHP is older than 5.5.

- [x] Includes unit test for hash and verify
- [x] Fallback with PHP < 5.5

Thanks to @ss23 for the contribution anyways :wink: 

Fixes #2954


lippserd:
Hi,

Thanks for the PR. Since we're having our release in a couple of days I think it's a good idea to just implement #2954. Do you want to take care of that?

Cheers,
Eric

ss23:
@lippserd Sorry, I haven't been following development much. Does this new release increase the requirements to 5.5 or higher? If not, I presume it's okay to include a shim (https://github.com/ircmaxell/password_compat) directly in the vendor folder?

lippserd:
@ss23 No worries. We'll raise our PHP dep to version 5.6 with our next release. So there's no need for a shim.

lazyfrosch:
I took the liberty to update your commit, please see my notes in the first comment.

lazyfrosch:
Since Blowfish keeps bugging me during testing (it seems to be an issue when not enough entropy is available)

New Plan: password_hash for PHP >= 5.5
And a Fallback with crypt and SHA512

lazyfrosch:
So, unfortunately my new commits replace all the code you touched @ss23, and I wasn't able to find any history of this PR.

Please have a look on the new handling that should be compatible over all PHP versions (at least >= 5.3)

Let's see how travis tests this, then it should be a good feature for 2.5.0

@lippserd please review
@ss23 Thanks for your contribution by poking us in the right direction :wink:

lazyfrosch:
Seems like my regexp in tests are a problem in PHP <5.5

Will fix that tomorrow, rest can be reviewed :)

lazyfrosch:
Tests green <sup>TM</sup>

PR #3094, opened 2017-11-16
Support and documentation changes for 2.5.0

lazyfrosch:
This is a general PR to introduce some changes in requirements for 2.5.0

Changes:
- [x] Add FPM config examples
- [x] Add FPM documentation for el6 / el7
- [x] Remove SLES 11 from docs
- [x] Upgrading documentation for 2.5.0
  * Package deprecation
  * Package changes to FPM for some OS
  * Notes for PHP upgrade on SUSE
- [x] Removed RPM spec file from this repo

Also see:
* https://github.com/Icinga/icinga-packaging/pull/60
* https://github.com/Icinga/packages.icinga.com/pull/2


lazyfrosch:
Also see https://github.com/Icinga/packages.icinga.com/pull/2

doc/02-Installation.md
```diff
@@ -104,17 +98,41 @@ apk update
...
+Please enable it prior installing the Icinga packages.
+ 
 CentOS 7/6:
-```
-yum install epel-release
-```
 
-If you are using RHEL you need enable the `optional` repository and then install
-the [EPEL rpm package](https://fedoraproject.org/wiki/EPEL#How_can_I_use_these_extra_packages.3F).
+    yum install epel-release
```
dnsmichi: Please keep the format with the three backticks in the docs.

library/Icinga/Application/Hook.php
```diff
@@ -275,21 +277,23 @@ public static function all($name)
...
      */
-    public static function first($name)
+    public static function first($name, $bypassPermission = false)
     {
         $name = self::normalizeHookName($name);
 
         if (self::has($name)) {
             foreach (self::$hooks[$name] as $key => $hook) {
-                if (self::hasPermission($hook)) {
+                if ($bypassPermission || self::hasPermission($hook)) {
```
dnsmichi: That sounds dangerous. Where is that used?

doc/02-Installation.md
lazyfrosch: Is there a reason for that?

library/Icinga/Application/Hook.php
lazyfrosch: Within the RequirementHook, reason is that `setup` doesn't use the normal role/permission model

doc/02-Installation.md
dnsmichi: The entire documentation uses that format, and we should keep it that way.

library/Icinga/Application/Hook.php
dnsmichi: Ah ok, understood, thanks 👍

lazyfrosch:
@lippserd please merge

PR #3097, opened 2017-11-16
Implement togglable flyover

### LINKED ISSUE #3024: Toggleable flyover containers

Al2Klimov:
fixes #3024

### DIFF
--- a/public/js/icinga/behavior/flyover.js
+++ b/public/js/icinga/behavior/flyover.js
@@ -1,0 +1,51 @@
+/* Icinga Web 2 | (c) 2017 Icinga Development Team | GPLv2+ */
+
+/**
+ * Icinga.Behavior.Flyover
+ *
+ * A toggleable flyover
+ */
+(function(Icinga, $) {
+
+    'use strict';
+
+    var expandedFlyovers = {};
+
+    function Flyover(icinga) {
+        Icinga.EventListener.call(this, icinga);
+
+        this.on('rendered', this.onRendered, this);
+        this.on('click', '.flyover-toggle', this.onClick, this);
+    }
+
+    Flyover.prototype = new Icinga.EventListener();
+
+    Flyover.prototype.onRendered = function(event) {
+        // Re-expand expanded containers after an auto-refresh
+
+        $(event.target).find('.flyover').each(function() {
+            var $this = $(this);
+
+            if (typeof expandedFlyovers['#' + $this.attr('id')] !== 'undefined') {
+                $this.toggleClass('flyover-expanded');
+            }
+        });
+    };
+
+    Flyover.prototype.onClick = function(event) {
+        var $flyover = $(event.target).closest('.flyover');
+
+        $flyover.toggleClass('flyover-expanded');
+
+        if ($flyover.hasClass('flyover-expanded')) {
+            expandedFlyovers['#' + $flyover.attr('id')] = null;
+        } else {
+            delete expandedFlyovers['#' + $flyover.attr('id')];
+        }
+    };
+
+    Icinga.Behaviors = Icinga.Behaviors || {};
+
+    Icinga.Behaviors.Flyover = Flyover;
+
+})(Icinga, jQuery);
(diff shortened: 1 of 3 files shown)


Al2Klimov:
As discussed w/ @nilmerg and then @lippserd (offline):

In our current use cases we can really prefer this solution over the `loadUrl(...)` foo.

Al2Klimov:
Use case branch coming soon.

public/css/icinga/flyover.less
```diff
@@ -0,0 +1,16 @@
...
+
+.flyover:not(.flyover-expanded) .flyover-content {
+  display: none;
+}
+
+.flyover {
+  position: relative;
+
+  .flyover-content {
+    position: absolute;
```
lippserd: Please move `position: absolute` to the bottom of this CSS block. I like that most of our CSS is ordered alphabetically or by its function.

public/js/icinga/behavior/flyover.js
```diff
@@ -0,0 +1,48 @@
...
+        $flyover.toggleClass('flyover-expanded');
+
+        if ($flyover.hasClass('flyover-expanded')) {
+            expandedFlyovers.push($flyover.attr('id'));
+        } else {
+            expandedFlyovers.splice(expandedFlyovers.indexOf($flyover.attr('id')), 1);
+        }
+    });
+
+}) (Icinga, jQuery);
```
lippserd: Remove the whitespace.

public/js/icinga/behavior/flyover.js
```diff
@@ -0,0 +1,48 @@
...
+                $this.toggleClass('flyover-expanded');
+            }
+        });
+    };
+
+    Icinga.Behaviors = Icinga.Behaviors || {};
+
+    Icinga.Behaviors.Flyover = Flyover;
+
+    $(document).on('click', '.flyover .flyover-toggle', function(event) {
```
lippserd: Relax the selector to just `.flyover-toggle`.

public/js/icinga/behavior/flyover.js
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Icinga.Behavior.Flyover
+ *
+ * A toggleable flyover
+ */
+(function(Icinga, $) {
+
+    'use strict';
+
+    var expandedFlyovers = [];
```
lippserd: Use object instead of array. And use `hasOwnProperty` for the isset checks.

public/js/icinga/behavior/flyover.js
```diff
@@ -0,0 +1,48 @@
...
+
+        this.on('rendered', this.onRendered, this);
+    }
+
+    Flyover.prototype = new Icinga.EventListener();
+
+    Flyover.prototype.onRendered = function(event) {
+        $(event.target).find('.flyover').each(function() {
+            var $this = $(this);
+
```
lippserd: Please add a comment here that this is for retaining state across auto-refresh.

public/js/icinga/behavior/flyover.js
lippserd: Register this event handler in the behavior's constructor with `this.on()` instead of using `$(document)` directly.

public/css/icinga/flyover.less
```diff
@@ -0,0 +1,16 @@
...
+    background-color: #ffffff;
```
lippserd: Please use `@body-bg-color` here.

PR #3108, opened 2017-11-21
Persist X-Icinga-WindowId across auto-refreshes

Al2Klimov:


library/Icinga/Web/Request.php
```diff
@@ -114,8 +114,10 @@ public function isApiRequest()
     public function protectId($id)
     {
         if (! isset($this->uniqueId)) {
-            $this->uniqueId = Window::generateId();
+            $windowId = $this->getHeader('X-Icinga-WindowId');
+            $this->uniqueId = $windowId === false ? Window::generateId() : $windowId;
```
nilmerg: Please use empty() instead of === false, the header value may be the empty string.

And, the question is not whether empty header values are valid/supported/whatever but whether we should accept non-existing window ids. Which we definitely should not.

PR #3109, opened 2017-11-21
js, layout: detect initial layout on load

### LINKED ISSUE #3107: Detect initial layout and do not switch after loading
## Expected Behavior

Icinga Web 2 loads and renders as fast as possible without flickering

## Current Behavior

DOM is re-rendered once or twice. Once when loading the page as an authenticated user because of JS depending on Logo size (should be fixed with the new layout I guess). And once if the detected layout isn't the 'default' layout. This leaves the impression of slower load times, the page is flickering on load.

Thomas-Gelf:
fixes #3107


Thomas-Gelf:
ping @lippserd: any chance we can have this for v2.5.0?

PR #3122, opened 2017-11-21
Update Travis CI to PHP >= 5.6

dnsmichi:


lazyfrosch:
No, PHP 5.6 is a recommendation, not a hard requirement as of now.

It's merely to push general platform support, and for our modules (Director, Graphite, etc.), as well as upcoming changes.

We will raise the dependency once really needed.

Issue #2790, opened 2017-03-13
Provide hooks for custom Resource type


Hipska:
This will allow modules to add new resource types, and provide forms for them.

Ex. module for connecting to a REST API needs `username`, `password`, `url`, ...

Might be related to #2649

lippserd:
Hi,

Thanks for the report. This is indeed a good feature but I can't tell you at the moment when it's going to be implemented.

Best regards,
Eric

Al2Klimov:
Me: "Shall we implement https://github.com/Icinga/icingaweb2/issues/3035#issuecomment-342116220 via #2790?"
@lippserd: "We dont' have any plan for #2790, yet."
Me: "Then I'll forge one."
@lippserd: "OK."

Al2Klimov:
So @lippserd, here you are – the plan I've forged. (I hope you like it.)

Al2Klimov:
As discussed in https://github.com/Icinga/icingaweb2/issues/3035#issuecomment-342116220:

#3035 will include "an (abstract) resource type 'HTTP'".

So that will make your life much easier, @Hipska.

Hipska:
Indeed, but still no "generic" type?

Al2Klimov:
There *will* be a custom type – as part of *this* issue.

Hipska:
And when might that be?

Al2Klimov:
Ask @bobapple, not me.

PR #3136, opened 2017-11-23
/config/resource: sort resources by name

Al2Klimov:


Al2Klimov:
Yes, this is kinda low prio *yet*, but it will become more interesting once #2790 has been done.

PR #3137, opened 2017-11-23
Donuts: State names should be displayed in uppercase letters

N-o-X:


nilmerg:
The view script is redundant now. Closing.

PR #3139, opened 2017-11-23
Implement ResourceHook

### LINKED ISSUE #3055: Hook for resource locking
## Current Behavior

* If I try to delete an SQL DB resource used by an IDO backend, nothing prevents me from doing that.
* If I rename an SQL DB resource used by an IDO backend, nothing prevents me from doing that and nothing updates the backend config.

## Expected Behavior

* If I try to delete an SQL DB resource used by an IDO backend, something prevents me from doing that.
* If I rename an SQL DB resource used by an IDO backend, something updates the backend config.

## Possible Solution

A new hook which is used by module components which use resources.

* Before every resource deletion it is asked if something uses the respective resource. If yes, the deletion shall fail w/ a message provided by the hook.
* Before every resource renaming it is asked if something uses the respective resource and whether that *something* can cascade the name change. If cascading isn't possible, the renaming shall fail w/ a message provided by the hook. After renaming it cascades the name change.

Al2Klimov:
fixes #3055

### DIFF
--- a/application/controllers/ConfigController.php
+++ b/application/controllers/ConfigController.php
@@ -4,6 +4,7 @@
 namespace Icinga\Controllers;
 
 use Exception;
+use Icinga\Application\Hook;
 use Icinga\Application\Version;
 use InvalidArgumentException;
 use Icinga\Application\Config;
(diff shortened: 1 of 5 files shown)


nilmerg:
As mentioned in #3055, closed.

PR #3170, opened 2017-11-29
remove all COLLATE latin1_general_ci from sql statements

bodsch:
hard coded COLLATE breaks a pure utf-8 database and makes modules (e.g. graphite) unusable

### DIFF
--- a/modules/monitoring/library/Monitoring/Backend/Ido/Query/ContactQuery.php
+++ b/modules/monitoring/library/Monitoring/Backend/Ido/Query/ContactQuery.php
@@ -28,16 +28,16 @@ class ContactQuery extends IdoQuery
      */
     protected $columnMap = array(
         'contactgroups' => array(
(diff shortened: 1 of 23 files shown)


dnsmichi:
This will possibly break the IDO schema which relies on latin_general_ci for several columns.

bodsch:
not so good.
but a IDO schema with full utf-8 schema breaks with `COLLATE latin1_general_ci ` in these files. :(

it is possible to use the character set from the resource configuration?
then we can make this part configurable.

dnsmichi:
A database resource type can specify the character set already. Is that what you're looking for?

bodsch:
yes. exact this.

hot needels .. i have an codefragment, but untestet yet (puhhh, my PHP skills are sleeping;) )

bodsch:
this thing is a litte bit complex, i close this PR and make later a new one.

PR #3174, opened 2017-11-29
Don't discard fullscreen when detecting the layout

### LINKED ISSUE #3169: The parameter showFullscreen opens a new window with sidebar
hello,

while using the shiny new icingaweb2 i've discovered a small issue with the paramter `showFullscreen`.

some of my shared navigation elements should open in a new window with fullscreen mode. 

example: `monitoring/list/services?service_problem=1&service_state!=3&sort=service_severity&showFullscreen&showCompact`

## Expected Behavior
the parameter showFullscreen should open a new window without all the things like tabs, big logo, sidebar and so on.

## Current Behavior
the page will opened in a new window. the sidebar and tabs are invisible but few seconds later the sidebar and the icinga2 logo appears again.

## Your Environment

* Icinga Web 2 version and modules (System - About): ebe1917b152b786cc708abe5215fd0c4be7c4237
* Version used (`icinga2 --version`): `r2.7.2-1`
* Operating System and version: Windows 10 64bit
* Browser: Chromium 62.0.3202.75

lippserd:
Fixes #3169

### DIFF
--- a/public/js/icinga/ui.js
+++ b/public/js/icinga/ui.js
@@ -34,7 +34,12 @@
         var matched;
         $.each(classList, function(index, item) {
             if (null !== (matched = item.match(/^([a-z]+)-layout$/))) {
-                _this.currentLayout = matched[1];
+                var layout = matched[1];
+                if (layout !== 'fullscreen') {
+                    _this.currentLayout = layout;
+                    // Break loop
+                    return false;
+                }
             }
         });
     };


lippserd:
@Thomas-Gelf I changed the layout detection to not include fullscreen because it is just a layout modifier. Without this change our `layoutHasBeenChanged` function thinks that the layout has been changed because `currentLayout` equals to `fullscreen`. Then, we switch to `wide` and discard the `fullscreen-layout` class.

PR #3178, opened 2017-12-03
Correct documentation about icingacli

rgevaert:
icingacli is its own package on Debian and Ubuntu too.


dnsmichi:
@lazyfrosch did we split them up lately?

lazyfrosch:
@dnsmichi We did way back: https://anonscm.debian.org/cgit/pkg-nagios/pkg-icingaweb2.git/commit/?id=672fcae6a067078e31e521bc61d1e3a1bf4591ac

lazyfrosch:
But since 2.5.0 `icingacli` is also a dependency of `icingaweb2` on RPM...

dnsmichi:
Ah ok, thanks for the clarification.

PR #3180, opened 2017-12-04
Url: prefer static over self

Al2Klimov:


Al2Klimov:
Rationale: https://github.com/Icinga/icingaweb2-module-businessprocess/blob/master/library/Businessprocess/Web/Url.php#L69

PR #3207, opened 2017-12-13
Don't interpret shell comments inside fenced code blocks as MD headers

### LINKED ISSUE #3200: Markdown doesn't ignore comments (lines beginning with '#') inside triple backtick (```) code blocks
In the Icinga Web 2 Documentation, when a code block delimited by three backticks (\`\`\`) contains a comment line beginning with a '#', it does not honor the code block and renders the comment line as a new section header.  This makes perusing the docs nearly impossible, both because the TOC becomes unmanageable and because it breaks the docs into nonexistent and often incomplete sections.

## Expected Behavior
A code block enclosed in triple backticks (\`\`\`) should be rendered as a code block and any lines within the code block that begin with a '#' should be ignored by the parser.

## Current Behavior
Lines beginning with a '#' are treated as headings even when enclosed inside code blocks that are delimited by triple backticks (\`\`\`).

## Possible Solution
Replace the triple backtick delimiters by using 4 spaces at the beginning of each line in a code block.  This alternate method of delineating a code block is parsed correctly and valid Markdown code. Alternately the markdown parser can be fixed, which while admittedly the better solution is probably more work and not something I'm familiar with.

1. Install Icinga Web 2 v2.5.0 and enable the doc module
2. In Icinga Web 2, click Documentation, then Icinga Web 2
3. Look at the TOC - note that the configuration chapter should be '3', not '13', and that chapters 3-12 consist of parts of chapter 2.  Looking at the 02-Installation.md file and note that there are lines beginning with a '#' that are enclosed in triple backticks that match the incorrect section headers.
4. Look at chapter 30 - SELinux (which should be chapter 11) and note that there is a code block with '#' characters at the beginning of the lines.  Look at the file 90-SELinux.md and note that the corresponding lines are all indented by 4 spaces.

## Context
I'm just learning how to make Icinga2 work and am constantly referring to the docs.  This makes the Icinga Web 2 docs almost unusable via the doc module.

## Your Environment
* Icinga Web 2 version and modules (System - About): 2.5.0 (via RPM)
* Version used (`icinga2 --version`): r2.8.0-1 (via RPM)
* Operating System and version: CentOS 7

Al2Klimov:
fixes #3200

### DIFF
--- a/modules/doc/library/Doc/DocParser.php
+++ b/modules/doc/library/Doc/DocParser.php
@@ -161,10 +161,19 @@ class DocParser
             $lastLine = null;
             $stack = new SplStack();
             $cachingIterator = new CachingIterator($file, CachingIterator::TOSTRING_USE_CURRENT);
+            $insideFencedCodeBlock = false;
+
             for ($cachingIterator->rewind(); $line = $cachingIterator->valid(); $cachingIterator->next()) {
                 $fileIterator = $cachingIterator->getInnerIterator();
                 $line = $cachingIterator->current();
-                $header = $this->extractHeader($line, $fileIterator->valid() ? $fileIterator->current() : null);
+                $header = null;
+
+                if (substr($line, 0, 3) === '```') {
+                    $insideFencedCodeBlock = ! $insideFencedCodeBlock;
+                } elseif (! $insideFencedCodeBlock) {
+                    $header = $this->extractHeader($line, $fileIterator->valid() ? $fileIterator->current() : null);
+                }
+
                 if ($header !== null) {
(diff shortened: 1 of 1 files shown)


modules/doc/library/Doc/DocParser.php
```diff
@@ -161,10 +161,19 @@ public function getDocTree()
             $lastLine = null;
             $stack = new SplStack();
             $cachingIterator = new CachingIterator($file, CachingIterator::TOSTRING_USE_CURRENT);
+            $insideBackTicks = false;
```
lippserd: Please rename this to `insideBackticks`.

PR #3213, opened 2017-12-15
Drop PHP 5.3 support

Al2Klimov:


Al2Klimov:
Rationale: https://github.com/Icinga/icingaweb2/pull/3186#issuecomment-351960792

Thomas-Gelf:
I agree with you that it is time to drop 5.3. We announced this with 2.5.0, but we failed to enforce it. Changing this in a patch version might break existing setups that didn't listen and still run 2.5.0 or 2.5.1 on 5.3. It's not supported, sure. But it's not a good move to break things in a minor release. When merging this I'd first create a support/2.5 branch like we did in other projects. This would allow us still backport important fixes to a 2.5.2 in a safe and compatible way while moving on with the master.

This would however influence the release policy for Icinga Web 2, and it probably also depends on the schedule for 2.6. @lippserd: it's your choice!

modules/setup/library/Setup/WebWizard.php
```diff
@@ -568,11 +568,10 @@ public function getRequirements($skipModules = false)
         $set = new RequirementSet();
 
         $set->add(new PhpVersionRequirement(array(
-            'condition'     => array('>=', '5.3.2'),
+            'condition'     => array('>=', '5.4'),
```
nilmerg: @lippserd Aren't we using 5.6 as our new requirement?

lippserd: Yes, will be in a new PR. Did not want to mix things here.

Al2Klimov: @nilmerg I agree with you (and w/ @lippserd as well), but I wouldn't like all of our customers to jump out of the window at the same time. So I'd do it in more than one step.

PR #3227, opened 2017-12-19
Separate reachability and check source info in detail views

### LINKED ISSUE #2953: Separate reachability and check source info in detail views
Quote from #2735/dnsmichi:
> check_source and is_reachable are not necessarily bound to each other.
> 
> is_reachable is determined by host/service dependencies in Icinga 2 and updated accordingly, if i.e. a parent fails. If there are no dependencies defined, it will always be "reachable".
> 
> check_source is just the node name where the plugin was executed.
> 
> I'd suggest to move Reachability into a separate line, away from the Check Source. This removes possible questions.

Al2Klimov:
fixes #2953

### DIFF
--- a/modules/monitoring/application/views/scripts/partials/object/detail-content.phtml
+++ b/modules/monitoring/application/views/scripts/partials/object/detail-content.phtml
@@ -35,6 +35,7 @@
     <tbody>
         <?= $this->render('show/components/command.phtml') ?>
         <?= $this->render('show/components/checksource.phtml') ?>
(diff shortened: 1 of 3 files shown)


dnsmichi:
Hmm, just a minor thing - could we rename the key to "Reachable", and replace the value to "yes/no" still including the green/red bulb?

screen shot 2017-12-19 at 17 44 45

dnsmichi:
Awesome, thanks ❤️ 

screen shot 2018-01-16 at 10 34 36

PR #3230, opened 2017-12-20
/application-state: ignore unauthenticated requests

### LINKED ISSUE #2882: application-state should not actively log out
When visiting a controller with `$requiresAuthentication = false` I'm forcibly being logged out once `application-state` fires next time. This should be fixed by silently ignoring unauthenticated requests, giving them an empty response or similar when visiting `application-state`.

Al2Klimov:
fixes #2882

### DIFF
--- a/application/controllers/ApplicationStateController.php
+++ b/application/controllers/ApplicationStateController.php
@@ -14,43 +14,48 @@ use Icinga\Web\Session;
  */
 class ApplicationStateController extends Controller
 {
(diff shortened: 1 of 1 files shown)


lippserd:
Hi Alex,

Thanks for the PR. I did not test this yet but the call to `$this->_helper->layout()->disableLayout();` should always happen, right? What is the response of the request if the user is not authenticated?

Cheers,
Eric

Al2Klimov:
You're right...

PR #3242, opened 2018-01-03
Unescape filter queries before parsing

### LINKED ISSUE #3239: Icingaweb2 can't handle properly escaped URIs
## Expected Behavior
Icingaweb2 recognizes `%7C` in filter queries as the pipe character.

## Current Behavior
```
Invalid filter "((service=service1&host=host1)%7C(service=service2&host=host2))", unexpected ( at pos 34

#0 /usr/share/php/Icinga/Data/Filter/FilterQueryString.php(215): Icinga\Data\Filter\FilterQueryString->parseError('(')
#1 /usr/share/php/Icinga/Data/Filter/FilterQueryString.php(176): Icinga\Data\Filter\FilterQueryString->readFilters(1, NULL)
#2 /usr/share/php/Icinga/Data/Filter/FilterQueryString.php(260): Icinga\Data\Filter\FilterQueryString->readFilters()
#3 /usr/share/php/Icinga/Data/Filter/FilterQueryString.php(40): Icinga\Data\Filter\FilterQueryString->parseQueryString('((service=servi...')
#4 /usr/share/php/Icinga/Data/Filter/Filter.php(253): Icinga\Data\Filter\FilterQueryString::parse('((service=servi...')
#5 /usr/share/icingaweb2/modules/monitoring/application/controllers/ServicesController.php(35): Icinga\Data\Filter\Filter::fromQueryString('((service=servi...')
...
```

## Possible Solution
un-escape query string using [`urldecode()`](https://secure.php.net/manual/de/function.urldecode.php) before parsing

1. Select 2 services
2. copy URL into $chat (eg. https://monitoring.example.net/dashboard#!/monitoring/services/show?((service=service1&host=host1)|(service=service2&host=host2)) )
3. open URL from $chat

## Context
Most chat client do escaping on URLs as they should. This leads to occurences of `|` to be replaced with `%7C`.

We often send each other links to multiple checks in chat. Clicking those links leads to an icingaweb2 stacktrace.

also see #1803

## Your Environment
* Icinga Web 2 version 2.5.0
* Google Chrome, Safari, Firefox

mhutter:
This is required to convert escapet special characters (e.g. `|` getting
replaced by `%7C`) back to their unescaped form.

Fixes #3239

### DIFF
--- a/test/php/regression/Bug3239Test.php
+++ b/test/php/regression/Bug3239Test.php
@@ -1,0 +1,23 @@
+<?php
+/* Icinga Web 2 | (c) 2018 Icinga Development Team | GPLv2+ */
+
+namespace Tests\Icinga\Regression;
+
+use Icinga\Test\BaseTestCase;
+use Icinga\Data\Filter\FilterQueryString;
+
+/**
+ * Icingaweb2 can't handle properly escaped URIs
+ *
+ * @see https://github.com/Icinga/icingaweb2/issues/3239
+ */
+class Bug3239Test extends BaseTestCase
+{
+    /**
+     * Ensure an encoded filter does not throw any exceptions
+     */
+    public function testParseFilter()
+    {
+        FilterQueryString::parse('((service=a)%7C(service=b))');
+    }
+}
--- a/library/Icinga/Data/Filter/FilterQueryString.php
+++ b/library/Icinga/Data/Filter/FilterQueryString.php
@@ -248,11 +248,10 @@ class FilterQueryString
 
     protected function parseQueryString($string)
     {
+        $filter = urldecode($string);
         $this->pos = 0;
-
-        $this->string = $string;
-
-        $this->length = strlen($string);
+        $this->string = $filter;
+        $this->length = strlen($filter);
 
         if ($this->length === 0) {
             return Filter::matchAll();


mhutter:
I'll try to add a test case aswell

mhutter:
Done!

mhutter:
Any more infos needed? Any fedback?

test/php/regression/Bug3239Test.php
```diff
@@ -0,0 +1,23 @@
+<?php
+/* Icinga Web 2 | (c) 2013 Icinga Development Team | GPLv2+ */
```
Al2Klimov: Please always use the current year in newly created files.

mhutter: Whops! The copy/paste was strong with me :) fixed

nilmerg:
Hi, thanks for the contribution. However, this is not the correct solution for this issue. With the proposed changes applied the following valid case breaks:

`/icingaweb2/monitoring/list/hosts?service_description=mysql%20%28master%29`

Decoding the entire query string is dangerous and not recommended. However, the issue is most certainly valid, as sharing such urls is what we like to support. Feel free to come up with another proposal or discuss it with us in #3239.

lippserd:
I'll update the issue in a second.

PR #3247, opened 2018-01-04
Improve Event Grid performance

mdetrano:
I would like to submit for consideration this pull request to improve "event grid" performance in the monitoring web 2 module.

This patch partially addresses the issue 12732 - Improve IDO query performance.

The patch works by limiting the backend query for events by date, and also selecting only hosts or services in the main subquery (according to what is requested in the form).

In some rough testing on a large system with many (too many...) events, these changes reduced the initial page load time of the Event Grid from over 5 minutes to under 20 seconds.

### DIFF
--- a/modules/monitoring/application/controllers/ListController.php
+++ b/modules/monitoring/application/controllers/ListController.php
@@ -361,13 +361,16 @@ class ListController extends Controller
         );
         $orientationBox->applyRequest($this->getRequest());
 */
+        $objectType = $this->getParam('objecttype', 'services');
+        $from = date('c', $this->getParam('from', strtotime('midnight 3 months ago')));
         $query = $this->backend->select()->from(
-            'eventgrid',
+            'eventgrid' . $objectType,
             array('day', $form->getValue('state'))
         );
         $this->params->remove(array('objecttype', 'from', 'to', 'state', 'btn_submit'));
         $this->view->filter = Filter::fromQuerystring((string) $this->params);
         $query->applyFilter($this->view->filter);
+        $query->applyFilter(Filter::fromQuerystring('timestamp >= ' . $from));
         $this->applyRestriction('monitoring/filter/objects', $query);
         $this->view->summary = $query;
         $this->view->column = $form->getValue('state');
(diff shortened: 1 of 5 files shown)


lippserd:
Hi,

Thanks for the PR. We'll have a look at this one asap.

Best regards,
Eric

mdetrano:
Ok, just updated, rebased, squashed, the commit and got it passing the Travis checks.

modules/monitoring/application/controllers/ListController.php
```diff
@@ -361,13 +361,16 @@ public function eventgridAction()
         );
         $orientationBox->applyRequest($this->getRequest());
 */
+        $objectType = $this->getParam('objecttype', 'services');
+        $from = date('c', $this->getParam('from', strtotime('midnight 3 months ago')));
```
theFeu:
```suggestion
        $from = $form->getValue('from');
```

modules/monitoring/application/controllers/ListController.php
```diff
@@ -361,13 +361,16 @@ public function eventgridAction()
         );
         $orientationBox->applyRequest($this->getRequest());
 */
+        $objectType = $this->getParam('objecttype', 'services');
```
theFeu:
```suggestion
        $objectType = $form->getValue('objecttype');
```

I would suggest using this instead, because this way you can get the value from the form directly. (It also knows the defaults and that saves you some hussle)

modules/monitoring/library/Monitoring/Backend/Ido/Query/EventgridservicesQuery.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Icinga\Module\Monitoring\Backend\Ido\Query;
+
+class EventgridservicesQuery extends EventgridQuery
+{
+
```
theFeu:
```suggestion
```

Not to be completely nitpicky, but the excess line caught my eye :)

theFeu:
Looks good to me!
And again, thanks for contributing :)

PR #3253, opened 2018-01-11
Install PHP 7.1 from RedHat SCL

### LINKED ISSUE #2989: Vagrant: upgrade to PHP 7.1
We don't require the users to have PHP 7, yet – but we'll do that one nice day.

Right now it wouldn't harm anyone if we upgrade the PHP version in our Vagrant dev box. "One nice day" we just say "you need PHP 7" and... that's it.

Al2Klimov:
fixes #2989

### DIFF
--- a/.puppet/profiles/icingaweb2_dev/manifests/init.pp
+++ b/.puppet/profiles/icingaweb2_dev/manifests/init.pp
@@ -21,7 +21,15 @@ class icingaweb2_dev (
   }
 
   # TODO(el): icinga-gui is not a icingaweb2_dev package
-  package { [ 'php-gd', 'php-intl', 'php-pdo', 'php-ldap', 'php-phpunit-PHPUnit' ]:
+  package { [
+    'rh-php71-php-gd',
+    'rh-php71-php-intl',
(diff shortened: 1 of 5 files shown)


.puppet/profiles/icingaweb2_dev/manifests/init.pp
```diff
@@ -21,7 +21,16 @@
...
-  package { [ 'php-gd', 'php-intl', 'php-pdo', 'php-ldap', 'php-phpunit-PHPUnit', 'icinga-gui' ]:
+  package { [
+    'rh-php71-php-gd',
+    'rh-php71-php-intl',
+    'rh-php71-php-pdo',
+    'rh-php71-php-mysqlnd',
+    'rh-php71-php-pgsql',
+    'rh-php71-php-ldap',
+    'php-phpunit-PHPUnit',
+    'icinga-gui'
```
dnsmichi: You can drop this package, that's the Classic UI which isn't available as meta package for Icinga 2 anymore.

Al2Klimov: #3254

PR #3262, opened 2018-01-15
Make navigation via browser history working as expected

### LINKED ISSUE #2779: Browser history handling buggy with multiple columns involved
Given Icinga Web 2 with in view (#col2 is shown) and a link in #col2 pointing to _next, navigating `back` using browser controls doesn't work as expected. It should add the full former URL to the stack only once, however it add's two single steps with the last one (the first when going back) being completely unexpected, as it shows a col1/col2 combination that has not been shown before.

Al2Klimov:
fixes #2779

### DIFF
--- a/public/js/icinga/behavior/actiontable.js
+++ b/public/js/icinga/behavior/actiontable.js
@@ -407,19 +407,15 @@
             table.select($tr);
         }
 
-        // update history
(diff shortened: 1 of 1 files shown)


lippserd:
Please also add why the steps are unnecessary in the commit description.

PR #3269, opened 2018-01-16
Vagrant: increase timeout for proxy requests to FPM

### LINKED ISSUE #3266: Increase timeout for proxy requests to fpm in the development environment
The default timeout of 60 seconds is way to low when trying to debug with xDebug.

Should be increased to a more reasonable value for developers. (e.g. 3600)

For apache2 this would be:
```
ProxyTimeout 3600
```

Currently I don't know where it's best to put this directive. (At the moment it's at the bottom of `/etc/httpd/conf.d/icingaweb.conf` for me.)

Al2Klimov:
fixes #3266


Al2Klimov:
2) Yes, it is. 1) Only if something hangs... and hangs... and... but then it's that component's problem.

PR #3275, opened 2018-01-16
Vagrant: fix missing PHP extensions

### LINKED ISSUE #3265: Missing extensions for php7 in the development environment
xDebug is not installed/configured for the php7 fpm environment in the vagrant box.

Can be manually installed with `yum install rh-php71-php-xdebug`. (Configuration still required)

Should of course be part of the provisioning instead.

Al2Klimov:
fixes #3265


Al2Klimov:
1. Maybe not. But we could do this as we encounter other missing ones.
2. Because of timeouts.

Issue #2925, opened 2017-08-21
Comment filter never matches if additional filter is added to the standard filter


darmagan:
## Expected Behavior
Additional filter is added to the search statement.

## Current Behavior
Icingaweb2 shows that nothing matched the filter.

## Steps to Reproduce (for bugs)
1. Go to comments
2. Add a additional filter ( with +)
3. Apply
4. No match is shown

## Workaround
1. Reload the URL

## Your Environment
* Icinga Web 2 version and modules (System - About):
```
MODULE         VERSION   STATE     DESCRIPTION
businessprocess 2.0.0-rc1 enabled   A Business Process viewer and modeler
director       1.3.1     enabled   Director - Config tool for Icinga 2
monitoring     2.4.1     enabled   Icinga monitoring module
```
* Version used (`icinga2 --version`):
```
icinga2 - The Icinga 2 network monitoring daemon (version: v2.7.0)

Copyright (c) 2012-2017 Icinga Development Team (https://www.icinga.com/)
License GPLv2+: GNU GPL version 2 or later <http://gnu.org/licenses/gpl2.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Application information:
  Installation root: /usr
  Sysconf directory: /etc
  Run directory: /run
  Local state directory: /var
  Package data directory: /usr/share/icinga2
  State path: /var/lib/icinga2/icinga2.state
  Modified attributes path: /var/lib/icinga2/modified-attributes.conf
  Objects path: /var/cache/icinga2/icinga2.debug
  Vars path: /var/cache/icinga2/icinga2.vars
  PID path: /run/icinga2/icinga2.pid

System information:
  Platform: Ubuntu
  Platform version: 14.04.5 LTS, Trusty Tahr
  Kernel: Linux
  Kernel version: 3.13.0-101-generic
  Architecture: x86_64

Build information:
  Compiler: GNU 4.8.2
  Build host: 053bc2a10214

```

* Operating System and version:
```
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.5 LTS
Release:	14.04
Codename:	trusty

```

* Enabled features (`icinga2 feature list`):
```
Disabled features: gelf graphite ido-pgsql influxdb livestatus logstash opentsdb perfdata redis syslog
Enabled features: api checker command compatlog debuglog ido-mysql mainlog notification statusdata

```
[...]

PR #3276, opened 2018-01-17
Support array syntax in the filter editor

### LINKED ISSUE #3040: Support array syntax for filter values in the filter editor
This is a follow up of #2925. Our framework supports parsing filter values in the array syntax, e.g. `comment_type=(comment|ack)`. Our filter editor however does not support this format which leads to invalid filters after an additional filter has been added.

nilmerg:
This changes the filter editor in the following ways:
* Parentheses are not rendered anymore for sequences of expressions in the editor's input
* Two new (FilterEditor specific) filter signs are introduced: IN and NOT IN
* Sequences of expressions may also defined with commas (for the user's convenience only, they will be transformed to pipes when applied)
* The change introduced with #2925 is reverted

The latter two changes may be dropped without interfering with this solution.

The rationale for the others is to avoid encoding issues. The parentheses are not required in the editor, so we can safely encode any parentheses put in by the user. Introducing specific filter signs allows the editor to split by `|` or `,` while the user is able to recognize that there is the possibility to input sequences of expressions *and* to input literal pipes or commas by choosing the standard filter signs.

fixes #3040

### DIFF
--- a/library/Icinga/Web/Widget/FilterEditor.php
+++ b/library/Icinga/Web/Widget/FilterEditor.php
@@ -516,7 +516,7 @@ class FilterEditor extends AbstractWidget
     {
         $value = $filter === null ? '' : $filter->getExpression();
         if (is_array($value)) {
-            $value = '(' . implode('|', $value) . ')';
+            $value = implode('|', $value);
         }
         return sprintf(
             '<input type="text" name="%s" value="%s" />',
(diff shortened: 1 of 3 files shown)


nilmerg:
Well, turns out that the FilterEditor is better to be rewritten/replaced with some new tech. Fixing the issues with this proposal doesn't seem worth to me.

PR #3285, opened 2018-01-18
module/doc: End output buffering before serving module images

### LINKED ISSUE #3223: Documentation module: Large image generates memory error
When a module documentation is viewed with the doc module, and this documentation points to a large image, then php throws this memory allocation error:
> **Fatal error**: Allowed memory size of xxx bytes exhausted (tried to allocate xxx bytes) in **/usr/share/icingaweb2/library/vendor/Zend/Controller/Response/Abstract.php** on line **582**

I noticed that a image was not loading, so i tried to access it directly using `{baseURL}/doc/module/{moduleName}/image/{image}`, which gave the previous error.

## Expected Behavior
Do not load image contents in memory .

## Current Behavior
Somewhere, the image seems to be loaded in memory before being sent to the browser.

## Context
I know I could raise the PHP memory limit, but the image should not be put into PHP's memory in the first place, and secondly, I assume that current limit has a good value.

## Your Environment
PHP memory limit is 128MiB and image size is 40MiB+.

nilmerg:
fixes #3223

### DIFF
--- a/modules/doc/application/controllers/ModuleController.php
+++ b/modules/doc/application/controllers/ModuleController.php
@@ -176,7 +176,11 @@ class ModuleController extends DocController
         } else {
             $finfo = new finfo();
             $this->getResponse()
-                ->setHeader('Content-Type', $finfo->file($imagePath, FILEINFO_MIME_TYPE));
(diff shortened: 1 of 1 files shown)


nilmerg:
Well, correct observation. I guess. :sweat_smile:

PR #3291, opened 2018-01-18
IniParser: include the file location in parsing exceptions

Al2Klimov:
fixes #3252

PR #3293, opened 2018-01-19
/error/error: show the whole exception chain

Al2Klimov:


Al2Klimov:
Rationale: https://github.com/Icinga/icingaweb2/pull/3291#discussion_r162572392

dnsmichi:
Using this branch and #3291 cherry-picked on top looks very good, thank you 👍 

screen shot 2018-01-19 at 11 02 07

dnsmichi:
Okidoki, thanks for clarification.

application/controllers/ErrorController.php
```diff
@@ -86,9 +86,20 @@ public function errorAction()
                         Logger::error("%s\n%s", $exception, $exception->getTraceAsString());
                         break;
                 }
-                $this->view->message = $exception->getMessage();
+
                 if ($this->getInvokeArg('displayExceptions')) {
-                    $this->view->stackTrace = $exception->getTraceAsString();
+                    $this->view->messages = $this->view->stackTraces = array();
```
lippserd: Please separate those lines for better readability.

application/controllers/ErrorController.php
```diff
@@ -86,9 +86,20 @@ public function errorAction()
...
+
+                    for (; $exception !== null; $exception = $exception->getPrevious()) {
```
lippserd: Please use a do-while loop here because it is the appropriate control structure to use. Your loop does of course work though. But when looking at this code, one may has to first think about what you're trying to achieve with it. With do-while it is obvious. Plus, the first check for `$exception !== null` is superfluous.

application/controllers/ErrorController.php
```diff
@@ -86,9 +86,20 @@ public function errorAction()
...
+                        $this->view->messages[] = $exception->getMessage();
+                        $this->view->stackTraces[] = $exception->getTraceAsString();
+                    }
+                } else {
+                    $this->view->messages = array();
+
+                    for (; $exception !== null; $exception = $exception->getPrevious()) {
```
lippserd: Please use a do-while loop here too.

application/views/scripts/error/error.phtml
```diff
@@ -4,9 +4,17 @@
...
 <?php endif ?>
 <div class="content">
-    <p tabindex="-1" class="autofocus error-message"><?= nl2br($this->escape($message)) ?></p>
-<?php if (isset($stackTrace)): ?>
-    <hr>
-    <pre><?= $this->escape($stackTrace) ?></pre>
-<?php endif ?>
+<?php
+if (isset($stackTraces)) {
+    for ($i = 0; isset($messages[$i]); ++$i) {
```
lippserd: Is there a reason for not using foreach?

PR #3297, opened 2018-01-19
Implement whitelist for /iframe/

### LINKED ISSUE #3296: Whitelist for /iframe/
## Current Behavior

Everyone can send me an URL (chat, shared dashboard/nav) like https://my-insecure-iw2.example.com/iframe?url=http://badguy.example.com/break-all-teh-iframez

## Possible Solution

A set of regexes of allowed URLs.

Al2Klimov:
fixes #3296


nilmerg:
Regular expressions are overkill here. iframe targets shouldn't appear in the url at all.

PR #3299, opened 2018-01-22
actiontable.js: Use `event.currentTarget` instead of the closest tr

### LINKED ISSUE #3298: HTML in the first line of plugin output may lead to JS errors
Basic example: `<table><tr><td>OK</td></tr></table>`

Results in the following when clicked upon: 

```
Uncaught TypeError: Cannot read property 'match' of undefined
    at Icinga.Loader.onResponse (icinga.dev.js:14812)
    at fire (icinga.dev.js:3047)
    at Object.fireWith [as resolveWith] (icinga.dev.js:3159)
    at done (icinga.dev.js:8183)
    at XMLHttpRequest.<anonymous> (icinga.dev.js:8530)
```

nilmerg:
This event is triggered due to delegation. If any onclick event is able to bubble up this far that it gets triggered we can safely use whatever it has been triggered for. (i.e. it can only be a tr[href])

fixes #3298

### DIFF
--- a/public/js/icinga/behavior/actiontable.js
+++ b/public/js/icinga/behavior/actiontable.js
@@ -375,7 +375,7 @@
     ActionTable.prototype.onRowClicked = function (event) {
         var _this = event.data.self;
         var $target = $(event.target);
-        var $tr = $target.closest('tr');
+        var $tr = $(event.currentTarget);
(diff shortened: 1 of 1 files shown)


nilmerg:
@lippserd Commit and pull request updated.

PR #3304, opened 2018-01-23
Integrate DetailviewExtension into multi-select views

### LINKED ISSUE #3072: Enhance DetailViewExtension hook to multi-select views
Currently there is no good way to add custom content or options to a multi-select view in monitoring.

I would add this as extra function to the DetailViewExtension.

```php
public function getHtmlForObjects($objects) {}
```

Notes:
* Can't be abstract so its compatible
* Multi select views need to be extended for hooks

lazyfrosch:
Also make sure Javascript module handling will find the container for module auto-loading.

@lippserd Please review if this makes sense.

fixes #3072

### DIFF
--- a/modules/monitoring/application/controllers/HostsController.php
+++ b/modules/monitoring/application/controllers/HostsController.php
@@ -3,6 +3,7 @@
 
 namespace Icinga\Module\Monitoring\Controllers;
 
+use Exception;
 use Icinga\Data\Filter\Filter;
 use Icinga\Data\Filter\FilterEqual;
 use Icinga\Module\Monitoring\Controller;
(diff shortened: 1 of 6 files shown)


lazyfrosch:
Here is an example that utilizes these updated hook:

bildschirmfoto von 2018-01-23 17-31-35
bildschirmfoto von 2018-01-23 17-31-26

lazyfrosch:
@lippserd please review and merge

modules/monitoring/application/controllers/HostsController.php
```diff
@@ -151,6 +153,16 @@ public function showAction()
         $this->view->commentsLink = Url::fromRequest()->setPath('monitoring/list/comments');
         $this->view->sendCustomNotificationLink = Url::fromRequest()
             ->setPath('monitoring/hosts/send-custom-notification');
+
+        $this->view->extensionsHtml = array();
+        foreach (Hook::all('Monitoring\DetailviewExtension') as $hook) {
+            /** @var DetailviewExtensionHook $hook */
+            $module = $this->view->escape($hook->getModule()->getName());
+            $this->view->extensionsHtml[] =
```
lippserd: The extension should only be added to the array if the hook does not return the empty string. Else we may render empty containers.

modules/monitoring/application/controllers/ServicesController.php
```diff
@@ -153,6 +155,16 @@ public function showAction()
         $this->view->sendCustomNotificationLink = Url::fromRequest()->setPath(
             'monitoring/services/send-custom-notification'
         );
+
+        $this->view->extensionsHtml = array();
+        foreach (Hook::all('Monitoring\DetailviewExtension') as $hook) {
+            /** @var DetailviewExtensionHook $hook */
+            $module = $this->view->escape($hook->getModule()->getName());
+            $this->view->extensionsHtml[] =
```
lippserd: The extension should only be added to the array if the hook does not return the empty string. Else we may render empty containers.

lazyfrosch:
Rebased and updated to your suggestions.

I also copied the "only output if content" check to the normal DetailviewExtension usage.

lippserd:
@nilmerg Your feedback is still pending here.

PR #3309, opened 2018-01-24
LocalFileStorageTest: restore error_reporting() level and error handler

Al2Klimov:
... not to cause side effects in other tests


Al2Klimov:
Damn! Neither I can remember anything concrete, nor I could reproduce such. Let's say: It's just because order is half of life. OK?

PR #3310, opened 2018-01-24
Add PhpCodeValidityTest

Al2Klimov:


Al2Klimov:
It helped discovering e.g. #3306.

Al2Klimov:
... and cf85e66

Thomas-Gelf:
Cool, makes sense. Please try to explain this briefly in the commit message. I'll add some more comments inline.

modules/test/application/clicommands/PhpCommand.php
```diff
@@ -155,6 +158,96 @@ public function styleAction()
...
+            T_CLASS     => 'class',
+            T_INTERFACE => 'interface'
+        );
+
+        if (version_compare(PHP_VERSION, '5.4.0') > -1) {
+            $types[T_TRAIT] = 'trait';
+        }
+
+        $files = array();
+        $baseDir = realpath(__DIR__ . '/../../../..');
```
Thomas-Gelf: Is there any reason not to base this on Icinga\Application methods?

modules/test/application/clicommands/PhpCommand.php
```diff
@@ -155,6 +158,96 @@ public function styleAction()
...
+        foreach ($iterator as $path => $info) {
+            /** @var \SplFileInfo $info */
+            if (preg_match(
+                '~\A(?:test|vendor|modules/[^/]+/test|library/Icinga/Test)(?:/|\z)~',
+                $iterator->getInnerIterator()->getSubPath()
+            ) || ! ($info->isFile() && preg_match('/\.php\z/', $path))) {
+                continue;
+            }
+
+            $content = file_get_contents("file://$path");
```
Thomas-Gelf: Do we have to expect a $path to not be a file, therefor asking for an explicit `file://` to protect us?

Al2Klimov: I did that on principle.

modules/test/application/clicommands/PhpCommand.php
```diff
@@ -155,6 +158,96 @@ public function styleAction()
...
+    /**
+     * Run code-validity checks
+     *
+     * This command checks whether icingaweb and installed modules match PHP syntax.
+     *
+     * USAGE
+     *
+     * icingacli test php validity
+     */
+    public function validityAction()
```
Thomas-Gelf: A CLI Action is like a web Controller, at best it shouldn't contain any business logic. Could you please split this method into multiple ones and move it to a correspondingly named class, maybe in the Icinga\Test namespace?

modules/test/application/clicommands/PhpCommand.php
Thomas-Gelf: Ok, then please remove it - on principle :laughing:. It absolutely makes sense when opening files based on foreign input, but it is useless and unexpected here.

Al2Klimov:
@Thomas-Gelf Do you agree w/ me on
1. this fails only on PHP 5.3
2. it's a bug in PHP 5.3
3. It's time for #3213

Thomas-Gelf:
Please keep discussion related to #3213 on that ticket, I added a comment there. What exactly breaks on 5.3 here in this issue?

Al2Klimov:
https://travis-ci.org/Icinga/icingaweb2/jobs/332709824#L347

Thomas-Gelf:
That's a bad one, I've met this friend before :laughing:. As you probably already figured out it is triggered by a class with the same name existing in the same namespace. It could be fixed by aliasing the import I guess.

Thomas-Gelf:
(and yes, it's a bug in 5.3)

Al2Klimov:
Damn. Now I understand... and provide an alternative to #3213: https://github.com/PhenX/php-svg-lib/pull/28

Thomas-Gelf:
The linked fix looks good to me. I guess `use Svg\Gradient\Stop as GradiendStop;` would also have fixed this, in case you want to see the full import in the `use` list.

Al2Klimov:
Should we patch our shipped version until we update the lib "one nice day" and the change stays the same (but then it comes from upstream)?

Thomas-Gelf:
Patch it. We should try to avoid doing so, but as we already modified this library before - go for it.

modules/test/application/clicommands/PhpCommand.php
Al2Klimov: What for? No one else uses that "business logic". (And I hope no one will.)

Al2Klimov:
Should we? Aren't our components' tests just for our components?

nilmerg:
No-one prevents me from setting just my own module path. Then also not only our own components are covered but mine as well. And since setting multiple module paths is generally supported it should be here as well.

At least in my dev environment the tests fail because the audit module is not loaded due to it being installed in a separate path.

modules/test/application/clicommands/PhpCommand.php
Al2Klimov: Yes: This is PhpUnit, not our fully-bootstrapped CLI.

Al2Klimov:
TODO:

* [ ] #3771

nilmerg:
Still doesn't work even with #3852.

nilmerg:
There are lint tests in the actions now.

PR #3323, opened 2018-01-25
Monitoring: permit disabling IDO backends no Icinga writes to

Al2Klimov:


Al2Klimov:
1. It isn't. It's just for convenience.
2. You should not have a backend with bad/no schema at all. But you I2 cluster may fail or whatever.

Issue #2209, opened 2015-12-17
[dev.icinga.com #10877] Store menu entries in database instead of local files


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/10877

**Created by dgoetz on 2015-12-17 09:51:43 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _Backlog_
Last Update: _2016-04-10 22:09:23 +00:00 (in Redmine)_

---

At the moment setting Icinga Web 2 to store user preferences into the database only stores those preferences. I think also storing custom menu entries and dashboards would be useful, but of course will need additional work for the sharing feature.

---

**Relations**:

* relates [#10877](https://dev.icinga.com/issues/10877)

nilmerg:
We should avoid making the same mistake: #2129 (Items from modules should be prefixed with `<module-name>/`)

PR #3330, opened 2018-01-29
Navigation: reduce amount of class_exists() calls

### LINKED ISSUE #2129: [dev.icinga.com #10617] Navigation is firing a massive amount of useless class_exists calls
This issue has been migrated from Redmine: https://dev.icinga.com/issues/10617

**Created by tgelf on 2015-11-12 16:46:59 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _Backlog_
Last Update: _2016-03-15 12:38:52 +00:00 (in Redmine)_

---

This results in an unbelievable amount of needless stat calls in our class loaders. One call should be enough, configured renderer names MUST allow to figure out an eventual module name.

Al2Klimov:
fixes #2129

### DIFF
--- a/library/Icinga/Application/Config.php
+++ b/library/Icinga/Application/Config.php
@@ -462,7 +462,22 @@ class Config implements Countable, Iterator, Selectable
     {
         $itemTypeConfig = Navigation::getItemTypeConfiguration();
         if (! isset($itemTypeConfig[$type])) {
-            throw new IcingaException('Invalid navigation item type %s provided', $type);
+            if (strpos($type, '/') === false) {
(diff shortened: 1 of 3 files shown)


nilmerg:
This requires a migration path. User preferences for module items are now put into subdirectories (e.g. `../preferences/jdoe/monitoring/host-action.ini`) whereas on master they're put into separate files. (e.g. `../preferences/host-action.ini`) This is not acceptable for such a small change. It even doesn't provide the user's an advantage. We should ensure this will be improved when working on #2209, but I close here now.

PR #3354, opened 2018-02-11
protocol is wrong, fcgi, not fgci

mdeguzis:


lazyfrosch:
This not affects our packages, only the config generated via `icingacli`, packages use text config files directly:

https://github.com/Icinga/icingaweb2/blob/5d91ee8b3ce52bb08f5bfef8abb5c688dadbb393/packages/files/apache/icingaweb2.fpm.conf#L53

PR #3359, opened 2018-02-13
Vagrant: Fix package name for PHP xdebug in SCLO

dnsmichi:
refs #3265


lippserd:
LGTM. Feel free to merge

dnsmichi:
Thanks :) Seems the package was moved, since @nilmerg had an old state which was working.

PR #3367, opened 2018-02-16
DateTimePicker: display default values as expected

Al2Klimov:
Rationale: default values parsed from *nix timestamps are likely not to be aware of the user's timezone.


lippserd:
Hi Alex,

I close this PR because I don't think it is necessary. If you use `(new DateTime())->setTimestamp()` the time zone will be respected. Plus, why should it be blocked to set a `DateTime` object with a time zone different to the default?

Best,
Eric

PR #3368, opened 2018-02-19
TacticalController: Add OutputFormat json and csv

theFeu:
Mimics the behaviour of the ListController

Can output json and csv format via monitoring/tactical?format=json / monitoring/tactical?format=csv

### DIFF
--- a/modules/monitoring/application/controllers/TacticalController.php
+++ b/modules/monitoring/application/controllers/TacticalController.php
@@ -12,9 +12,6 @@ use Icinga\Web\Widget\Tabextension\OutputFormat;
 
 class TacticalController extends Controller
 {
-    /**
-     * @see ActionController::init
-     */
     public function init()
     {
         parent::init();
@@ -115,13 +112,10 @@ class TacticalController extends Controller
         $this->view->statusSummary = $summary;
     }
 
-
     /**
-     * Return all tabs for this controller
-     *
-     * @return Tabs
+     * Create all tabs for this controller
      */
-    private function createTabs()
+    protected function createTabs()
     {
         $this->getTabs()->extend(new OutputFormat(['pdf']));
     }
--- a/modules/monitoring/application/controllers/TacticalController.php
+++ b/modules/monitoring/application/controllers/TacticalController.php
@@ -8,16 +8,9 @@ use Icinga\Module\Monitoring\Controller;
 use Icinga\Web\Url;
 use Icinga\Web\Widget\Tabextension\DashboardAction;
 use Icinga\Web\Widget\Tabextension\MenuAction;
-use Icinga\Web\Widget\Tabextension\OutputFormat;
 
 class TacticalController extends Controller
 {
-    public function init()
-    {
-        parent::init();
-        $this->createTabs();
-    }
-
     public function indexAction()
     {
         $this->setAutorefreshInterval(15);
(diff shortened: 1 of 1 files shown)


modules/monitoring/application/controllers/TacticalController.php
```diff
@@ -8,9 +8,19 @@
 use Icinga\Web\Url;
 use Icinga\Web\Widget\Tabextension\DashboardAction;
 use Icinga\Web\Widget\Tabextension\MenuAction;
+use Icinga\Web\Widget\Tabextension\OutputFormat;
 
 class TacticalController extends Controller
 {
+    /**
+     * @see ActionController::init
```
nilmerg: Please don't add such boilerplate documentation. If you don't have anything more to say, (for overridden methods) just omit the documentation entirely.

modules/monitoring/application/controllers/TacticalController.php
```diff
@@ -103,4 +114,15 @@ public function indexAction()
             ->render();
         $this->view->statusSummary = $summary;
     }
+
+
```
nilmerg: One empty line suffices here ;)

modules/monitoring/application/controllers/TacticalController.php
```diff
@@ -103,4 +114,15 @@ public function indexAction()
...
+    /**
+     * Return all tabs for this controller
+     *
+     * @return Tabs
+     */
+    private function createTabs()
```
nilmerg: Please use `protected` instead. `private` is only for very specific use-cases. This isn't one of these.

modules/monitoring/application/controllers/TacticalController.php
```diff
@@ -103,4 +114,15 @@ public function indexAction()
...
+    {
+        $this->getTabs()->extend(new OutputFormat(['pdf']));
```
nilmerg: Why no PDF export? Isn't that the most desired use here?

modules/monitoring/application/controllers/TacticalController.php
theFeu: I literally just copied this method from the ListContoller char by char.
So you would rather have me take out the entire docu for init() ?

modules/monitoring/application/controllers/TacticalController.php
theFeu: The same argument as above; it is private in the ListController...
Should I change both or would you rather leave it untouched there?

modules/monitoring/application/controllers/TacticalController.php
nilmerg: That's how we handle this now. The code in the ListController is simply old.

modules/monitoring/application/controllers/TacticalController.php
nilmerg: The same answer as above. The ListController is old. ;)

And no, just change your own code. The ListController is fine for now, even if ancient.

modules/monitoring/application/controllers/TacticalController.php
theFeu: That's mostly because I just wanted it to be built in quickly, and for some reason the pdf format is horribly broken here...

nilmerg: Then please investigate this further. I see absolutely no reason for this change without PDF support. (Whether it produces fancy results or not)

Or simply hide the export functionality in the UI, but only if you need the csv/json functionality for other reasons. ;)

PR #3369, opened 2018-02-19
SELinux: Add a new plugin context for icingacli

### LINKED ISSUE #3235: Icinga user should be allowed to read /etc/icingaweb2
Some checks in the ITL run `icingacli` which fails in default setups because Icinga 2 is not allowed to read `/etc/icingaweb2`.

What should be discussed is how to change permissions for that. We could simply add the `icinga` user (or its equivalent) to the `icingaweb2` group, but what about hosts only running Icinga Web 2 without Icinga 2 installation. While this is rather unlikely to happen we should cover odd setups as well.

dgoetz:
While this would resolve #3235, it would introduce a new packaging dependency for icingaweb2-selinux requiring icinga2-selinux during build and runtime. I could solve it in other way, but this would introduce the requirement simple from the other side, this way seams more logical for me.

If this gets included, I will create two further pull request. One to add the new plugin context to Icinga 2 docs and one for packaging.

### DIFF
--- a/packages/selinux/icingaweb2.fc
+++ b/packages/selinux/icingaweb2.fc
@@ -5,3 +5,5 @@
 /var/log/icingaweb2(/.*)?         gen_context(system_u:object_r:icingaweb2_rw_content_t,s0)
 /var/cache/icingaweb2(/.*)?         gen_context(system_u:object_r:icingaweb2_rw_content_t,s0)
 /var/lib/icingaweb2(/.*)?         gen_context(system_u:object_r:icingaweb2_rw_content_t,s0)
+
+/usr/bin/icingacli        gen_context(system_u:object_r:nagios_icingacli_plugin_exec_t,s0)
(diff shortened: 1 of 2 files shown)


dnsmichi:
I'm not sure if we should require icinga2-selinux here. Thoughts @lippserd ?

widhalmt:
I got the same issue, so +1 from me.

lippserd:
Hi,

To me it feels wrong to depend on `icinga2-selinux` here. How else could this be fixed? Is it sufficient to add a command to our docs that allows this maybe?

Best,
Eric

dgoetz:
No, it requires the icinga2_t domain to exists during compilation of the policy. Without it I can not use it directly or with a macro.

Other possible solutions would be moving both policies to one location and build them both at ones or move both upstream. First option was not wanted, but perhaps it would be an option now with the newest version of the build pipeline. Second one would be an long-term option, but I fear we have not enough user feedback to call the policy stable enough for it being only upstream.

lippserd:
We should extend our Icinga 2 docs as mentioned in the referenced issue in order to explain how to allow reading the icinga web 2 config dir. I think that's best because the config dir is configurable.

PR #3376, opened 2018-03-01
Fix interference between two similar namespaces

Al2Klimov:


Al2Klimov:
W/o this iplx\Sql\Select is searched in directories for the ipl namespace.

PR #3379, opened 2018-03-05
Adding available macros for url and multi urls

pandian912:
Documentation on available icingaweb2 macros and way to include multiple urls


nilmerg:
@lippserd I thought we don't support `notes_url` and `action_url` officially anymore. Navigation Items are what users are supposed to use instead. Did this change?

nilmerg:
Hi pandian912,

I'm afraid but this is not what we'd like to have in the Web 2 documentation. It's a legacy feature from the old days.

You're rather supposed to utilize the navigation item functionality in Icinga Web 2 to accomplish the same. (There you even have the possibility to use icons and titles.)

Best,
Johannes

PR #3383, opened 2018-03-07
PluginOutput: Fix text splicing for status tags in HTML output

### LINKED ISSUE #3382: PluginOutput rendering problem with state tags
We support several state tags like `[OK]` in the plugin output.

I noticed that this breaks when doing HTML in output, in a weird way.

Without tags:
image

With tags:
image

This must happen somehow inside `PluginOutput::processhtml()`

## Your Environment
* Icinga Web 2 version and modules (System - About): `2.5.1 + patches`
* Operating System and version: CentOS 7 FPM

lazyfrosch:
Trailing text was lost in processing.

Also add tests to check this behavior plus some basics.

fixes #3382


lazyfrosch:
@lippserd please review, merge and backport

PR #3392, opened 2018-03-15
Always show full date + time on history list for notifications

Mikesch-mp:
In history list we need to see always the full date + time when a notification was send and not only "on Mar 9" or similar and have to do mouseover to see the time.

View as it is now:
image

New view:
image

### DIFF
--- a/modules/monitoring/application/views/scripts/list/notifications.phtml
+++ b/modules/monitoring/application/views/scripts/list/notifications.phtml
@@ -35,7 +35,7 @@ if (! $this->compact): ?>
             <td class="state-col state-<?= $stateName ?>">
                 <div class="state-label"><?= $stateLabel ?></div>
                 <div class="state-meta">
-                    <?= $this->timeAgo($notification->notification_timestamp, $this->compact) ?>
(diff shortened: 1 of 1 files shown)


modules/monitoring/application/views/scripts/list/notifications.phtml
```diff
@@ -35,7 +35,7 @@ if (! $this->compact): ?>
             <td class="state-col state-<?= $stateName ?>">
                 <div class="state-label"><?= $stateLabel ?></div>
                 <div class="state-meta">
-                    <?= $this->timeAgo($notification->notification_timestamp, $this->compact) ?>
+                    <?= $this->formatDateTime($notification->notification_timestamp) ?>
```
dnsmichi: If we drop compact (aka timeOnly) here for compact views, does this still render in a good way on mobile screens?

Mikesch-mp: Schaut ok für mich aus
screenshot_20180322-082237

lippserd:
Hi Carsten,

Thanks for the PR. We'll include this PR in our next major release which is version 2.6.0.

Cheers,
Eric

Issue #3320, opened 2018-01-25
PHP7.0+MSPHPSQL support


chas0rde:
Since PHP 7.0 mssql-support has been removed from from PHP.
M$ started a github project to allow php-support again. [url](https://github.com/Microsoft/msphpsql)
If installed it is not recognized/available in icingaweb2 for creation of ressources.
Please include support to allow MSSQL based databases e.g. as usable resources in Icinga Director

## Expected Behavior
if MSSQL support via msphpsql is available on a system it should be detected correctly and allow usage of MSSQL databases

## Current Behavior
Platform controller test for MSSQL platform-support fails as it tests only for mssql extension based on the Zend DB Adapter PDO MSSQL

## Possible Solution
Additionally test for the M$ extensions in the platform.php
Also further support is required (tried quick-and-dirty fix but Zend framework used which might not be used then (I am not a PHP guy)

## Steps to Reproduce (for bugs)
1. install with php7.0+ 
2. install msphpsql extension and ODBC PDO drivers
3. see that MSSQL is not selectable as SQL DB ressource type while sqlcmd on cli allows access to MSSQL dbs

## Context
I try to establish a resource connection to a MSSQL server to read from DBs.
This icingaweb2-ressource should then be used as icinga director import-source

## Your Environment

* Icinga Web 2 version and modules (System - About): 2.5.1, director 1.4.2
* Version used (`icinga2 --version`): r2.8.1-1
* Operating System and version: ubuntu 16.04.3 LTS
* Enabled features (`icinga2 feature list`):
Disabled features: compatlog elasticsearch gelf influxdb opentsdb perfdata statusdata syslog
Enabled features: api checker command debuglog graphite ido-mysql livestatus mainlog notification
* Config validation (`icinga2 daemon -C`):
information/cli: Icinga application loader (version: r2.8.1-1)
information/cli: Loading configuration file(s).
information/ConfigItem: Committing config item(s).
information/ApiListener: My API identity: sv0003.corp.grimme.com
information/ConfigItem: Instantiated 1 ApiListener.
information/ConfigItem: Instantiated 3 Zones.
information/ConfigItem: Instantiated 1 Endpoint.
information/ConfigItem: Instantiated 2 ApiUsers.
information/ConfigItem: Instantiated 2 FileLoggers.
information/ConfigItem: Instantiated 1101 Notifications.
information/ConfigItem: Instantiated 6 NotificationCommands.
[...]

PR #3400, opened 2018-03-27
Platform: Correct detecting MSSQL Support

lazyfrosch:
PDO Implementation of zf1 builds on module mssql OR pdo_dblib, while dblib is preferred.

Newer implementations would support pdo_sqlsrv, but that is not available with ZF1.

### DIFF
--- a/library/Icinga/Application/Platform.php
+++ b/library/Icinga/Application/Platform.php
@@ -351,7 +351,8 @@ class Platform
      */
     public static function hasMssqlSupport()
     {
-        return static::extensionLoaded('mssql') && static::classExists('Zend_Db_Adapter_Pdo_Mssql');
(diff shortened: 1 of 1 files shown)


lazyfrosch:
Also see:
* https://blog.netways.de/?p=108789 (coming out on Friday)
* https://github.com/lazyfrosch/rpm-php-extras

Al2Klimov:
Hello @lazyfrosch!

would your change fix #3320?

Best,
AK

lazyfrosch:
We can't use the new MS module as of now, which is basically a limitation in terms of ZF1.

But we can use `pdo_dblib`, which is **included** with PHP, but not enabled on every platform.

Al2Klimov:
So it is opt-in?!

PR #3402, opened 2018-03-29
On login redirect url hook

b1zzu:
With this new method in the AuthenticationHook is possible to set a different authentication Url after the login. This is important to redirect the user back to a third party application after the login.

In our use case if the grafana session is expired the user is redirected from grafana to Icinga Web login page, and after the login, the user is redirected back to the grafana page.
This is an example of how we have implemented it:
```php
     public function onLoginRedirect(Request $request)
     {
        $redirect = $request->get('redirect');
        if (strpos($redirect, '/grafana/') === 0) {
            // the GrafanaUrl class overrides the fromPath() to set the correct path
            return GrafanaUrl::fromPath($redirect);
        }
        // is inportant to return null if the user doesn't come from grafana 
        return null;
    }
```


Thomas-Gelf:
ping @lippserd

lippserd:
Hi,

Thanks for the PR. I don't think that we should handle this in our framework because it adds complexity to the hook and login process and may produce ambiguous states. You could easily do this in `onLogin()`:

```php
use Icinga\Application\Icinga;

$app = Icinga::app();
$request = $app->getRequest();

if ($request ...) {
    $url = ...;

    $response = $app->getResponse();

    // Force HTTP redirection (See 990df95f3619859fff7063b8d69607cd40799036)
    $response->setHeader('X-Icinga-Redirect-Http', 'yes', true);

    $response->redirectAndExit($url);
}
```

Please correct me if I'm wrong.

Best regards,
Eric

PR #3421, opened 2018-04-20
Documentation for the webserver installation on Debian

winem:
The documentation was missing and it just said: 
> Debian / Ubuntu
> Your web server should be up and running after the installation.

Fixed this and added a short how-to test it.

### DIFF
--- a/doc/02-Installation.md
+++ b/doc/02-Installation.md
@@ -230,7 +230,7 @@ systemctl enable apache2.service
     
 **Debian / Ubuntu**
 
-Your web server should be up and running after the installation.
+Your web server should be up and running after the installation of Icinga Web 2.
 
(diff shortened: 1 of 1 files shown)


dnsmichi:
That's for the reason that the Debian packages pull in Apache as dependency already. Didn't you have that on your system?

winem:
Yes, you are right. It's a dependency and already installed when you hit the point in the documentation. So I just misread the documentation. 
I reverted my change and added a small addition to make it a bit more clear I think. But it's no longer that important I think. You just have to be a bit careful when you read the documentation. 

Thanks for the feedback!

lippserd:
Hi,

Thanks for the small change. I'll merge it.

Cheers,
Eric

PR #3430, opened 2018-04-26
tipsy: Don't use opacity but rgba to avoid having transparent text

### LINKED ISSUE #3418: Transparent text in tooltips may lead to contrast issues
Tooltip texts are transparent and appear in the same color as the background.

web2-transparent-tooltip-text

## Expected Behavior
Tooltip texts are not transparent.

## Current Behavior
Tooltip texts are transparent.

## Possible Solution
Make tooltip texts not transparent.

Have a tooltip overlaying the submit button for example. (As shown in the picture above)

## Your Environment
* Icinga Web 2 version and modules (System - About): master
* Browser: Chromium Version 65.0.3325.181
* Operating System and version: vagrant

nilmerg:
fixes #3418

### DIFF
--- a/public/css/icinga/widgets.less
+++ b/public/css/icinga/widgets.less
@@ -300,6 +300,7 @@ ul.tree li a.error:hover {
   font-size: @font-size-small;
   max-width: 300px;
   text-align: left;
+  background-color: rgba(0,0,0,0.8);
 }
 
 .progress-label span {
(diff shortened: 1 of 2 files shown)


lippserd:
Hi Johannes,

This is ready to merge, right?

Best,
Eric

nilmerg:
Yes. It is.

PR #3432, opened 2018-04-26
Show the last executed command in the detail view

### LINKED ISSUE #1344: [dev.icinga.com #8315] Use Icinga 2 API to expand commands
This issue has been migrated from Redmine: https://dev.icinga.com/issues/8315

**Created by bsheqa on 2015-01-29 15:29:40 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _Backlog_
Last Update: _2016-12-07 12:35:51 +00:00 (in Redmine)_

---

As the configuration of icinga2 is getting more and more complicated, it would be nice to see the resulting command of a service in the webinterface. This feature was available in icnga-classicui, but only works for icinga1.x

---

**Relations**:

* duplicates [#6879](https://dev.icinga.com/issues/6879)
* duplicates [#8315](https://dev.icinga.com/issues/8315)

Al2Klimov:
resolves #1344

### DIFF
--- a/modules/monitoring/library/Monitoring/Web/Controller/MonitoredObjectController.php
+++ b/modules/monitoring/library/Monitoring/Web/Controller/MonitoredObjectController.php
@@ -4,6 +4,9 @@
 namespace Icinga\Module\Monitoring\Web\Controller;
 
 use Exception;
+use Icinga\Data\Filter\Filter;
+use Icinga\Exception\ConfigurationError;
+use Icinga\Module\Monitoring\Backend\Icinga2Api;
 use Icinga\Module\Monitoring\Controller;
 use Icinga\Module\Monitoring\Forms\Command\Object\CheckNowCommandForm;
 use Icinga\Module\Monitoring\Forms\Command\Object\DeleteCommentCommandForm;
@@ -12,6 +15,8 @@ use Icinga\Module\Monitoring\Forms\Command\Object\ObjectsCommandForm;
 use Icinga\Module\Monitoring\Forms\Command\Object\RemoveAcknowledgementCommandForm;
 use Icinga\Module\Monitoring\Forms\Command\Object\ToggleObjectFeaturesCommandForm;
 use Icinga\Module\Monitoring\Hook\DetailviewExtensionHook;
+use Icinga\Module\Monitoring\Object\MonitoredObject;
+use Icinga\Module\Monitoring\Object\Service;
 use Icinga\Web\Hook;
 use Icinga\Web\Url;
 use Icinga\Web\Widget\Tabextension\DashboardAction;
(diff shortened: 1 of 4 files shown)


dnsmichi:
You're not fetching `last_check_result` which holds the `command` key, that's what Icinga 2 actually executed. `getCheckCommand()` instead tries to resolve static macros by parsing a check command object - leaving away runtime calculated values from functions, etc. Is that on purpose?

Al2Klimov:
I know that I could do it a lot easier, but there are env vars and protected customvars.

Btw. what does the core do in cases [like this](https://github.com/Icinga/icinga2/blob/72137b7631c8fc22fef41f342c0c3ed514cf154d/itl/command-plugins.conf#L72)? Do you resolve macros recursively until there's a string which appears not to be a macro string? `thisSECRETpassword$HASdollars$BECAUSEiCAN`

dnsmichi:
The Director implementation puts that into "inspect" which has its own permission. I would do the same for Icinga Web 2 - if you have "monitoring/executed-command" as permission, you can see the fully executed command line.

I don't see a possibility to remove sensitive custom attributes here. Since you've asked - the Icinga 2's macro resolver works in a recursive fashion. We also do have lots of functions for argument values inside the ITL itself - this is something you cannot re-implement in Icinga Web 2. If we would merge this PR, many default check commands would not highlight the exact executed command line, and bug reports will arise.

TL;DR - I'd suggest to use last_check_result.command from the host/service object and enable it on-demand via permission. That's basically what #1344 is all about.

modules/monitoring/library/Monitoring/Web/Controller/MonitoredObjectController.php
```diff
@@ -273,4 +283,107 @@ protected function setupQuickActionForms()
...
+            }
+
+            if (isset($attrs['env'])) {
+                foreach ($attrs['env'] as $key => $value) {
+                    $result[] = "$key={$this->escapeShellArg($this->resolveCommandMacros($value, $this->object, $attrs['vars']))}";
+                }
+            }
+        }
+
+        foreach ($objects['results'][0]['attrs']['last_check_result']['command'] as $arg) {
```
dnsmichi: Small not here: Checks can be pending, e.g. a fresh host/service object after a new config deployment. Then `last_check_result` returns null, and the `command` key lookup will break. You'll need another sanity check here if `last_check_result` actually is a dictionary.

Al2Klimov: No, it won't break as I've done an isset() before.

dnsmichi:
This will be provided by IcingaDB's schema inside the new monitoring module to my knowledge. I wouldn't bother with an additional roundtrip towards the REST API at this stage - @lippserd your opinion on that?

lippserd:
Yeah @dnsmichi is right. This PR is kind of overrun by time. Icinga DB and the new monitoring module will provide this information. Further, there is the Director with its inspect functionality. @nilmerg Feel free to close this one.

PR #3444, opened 2018-04-27
Json::encode(): auto-sanitize bad UTF-8 strings

### LINKED ISSUE #2635: [dev.icinga.com #13427] Improve error handling when responding to non-HTML requests
This issue has been migrated from Redmine: https://dev.icinga.com/issues/13427

**Created by medicmomcilo on 2016-12-06 15:43:52 +00:00**

Assignee: _medicmomcilo_
Status: _Feedback_ 
Target Version: _Backlog_
Last Update: _2016-12-26 00:08:41 +00:00 (in Redmine)_

---

Hi friends!

Recently in our infrastructure Nagstamon suddenly stopped working.
After some hours of troubleshooting we identified that this is due to JSON output is not being provided:
/monitoring/list/services?service\_state&gt;0&service\_state&lt;=3&service\_state\_type=1&addColumns=service\_last\_check&format=json

Our assumption is that because of number of criticals it stopped, this can also be confirmed with narrowing down the filter.

You can test this with inclusion of services in OK state if you don't have enough testing criticals. URL for that:
/monitoring/list/services?service\_state&gt;=0&service\_state&lt;=3&service\_state\_type=1&addColumns=service\_last\_check&format=json

Do let me know if there is anything else I can provide to help you investigate this issue further.

Kind regards,
Momcilo.

**Attachments**

* [patch_file.patch](https://dev.icinga.com/attachments/download/3893/patch_file.patch) aklimov - _2016-12-08 15:44:53 +00:00_
* [detect_error.patch](https://dev.icinga.com/attachments/download/3917/detect_error.patch) aklimov - _2016-12-14 12:36:45 +00:00_

Al2Klimov:
fixes #2635

### DIFF
--- a/library/Icinga/Util/Json.php
+++ b/library/Icinga/Util/Json.php
@@ -69,6 +69,7 @@ class Json
                 if ($autoSanitize) {
                     return static::encode(static::sanitizeUtf8Recursive($value), $options, $depth);
                 }
+                // Fallthrough
 
             default:
                 throw new JsonEncodeException('%s: %s', static::lastErrorMsg(), var_export($value, true));
--- a/library/Icinga/Util/Json.php
+++ b/library/Icinga/Util/Json.php
@@ -17,12 +17,42 @@ class Json
      * @param   mixed   $value
      * @param   int     $options
      * @param   int     $depth
+     *
+     * @return  string
+     * @throws  JsonEncodeException
+     */
+    public static function encode($value, $options = 0, $depth = 512)
+    {
+        return static::encodeAndSanitize($value, $options, $depth, false);
+    }
+
+    /**
+     * {@link json_encode()} wrapper, automatically sanitizes bad UTF-8
+     *
+     * @param   mixed   $value
+     * @param   int     $options
+     * @param   int     $depth
+     *
+     * @return  string
+     * @throws  JsonEncodeException
+     */
+    public static function sanitize($value, $options = 0, $depth = 512)
+    {
+        return static::encodeAndSanitize($value, $options, $depth, true);
+    }
+
+    /**
+     * {@link json_encode()} wrapper, sanitizes bad UTF-8
+     *
+     * @param   mixed   $value
+     * @param   int     $options
+     * @param   int     $depth
      * @param   bool    $autoSanitize   Automatically sanitize invalid UTF-8 (if any)
      *
      * @return  string
      * @throws  JsonEncodeException
      */
-    public static function encode($value, $options = 0, $depth = 512, $autoSanitize = false)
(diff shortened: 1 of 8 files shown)


nilmerg:
I also expect `Json::encode` and `Json::decode` to behave exactly as their functional counterparts. They may behave better in terms of quality/usability (which is their intention) but definitely not in terms of error handling.

Also, since this is a class in the `Icinga\Util` namespace, it's highly likely it's being used in the wild. Therefore suppressing exceptions a caller previously handled and expected is bad.

So, please make the sanitation optional.

dnsmichi:
Depending on where these functions are used, you'd need sanitation or not. I wouldn't move this into the Json* functions, but leave them outside. This renders the code more visible and transparent, and you don't need to know the function signature (which would be different from the inner `json_encode($value, $options, $depth)`).

library/Icinga/Web/Response/JsonResponse.php
```diff
@@ -190,7 +222,9 @@ public function outputBody()
                 $body['data'] = $this->getSuccessData();
                 break;
         }
-        echo json_encode($body, $this->getEncodingOptions());
+        echo $this->autoSanitize
```
lippserd: Please use `getAutoSanitize()`. Subclasses may override this method. Though we don't any at moment.

Al2Klimov: Always used `$this->whatever` – no one complained. Shall I do member reads like you everywhere in the future?

Al2Klimov:
I already added [a comment](https://github.com/Icinga/icingaweb2/pull/3444/files#diff-935d33cbed5a17e23b525ba7eaf5392aR67). Didn't our tests respect it?

library/Icinga/Web/Response/JsonResponse.php
lippserd: I stumbled upon this several times while working on the ipl. I'd say that we did not care about subclass overrides before. Yes, please use the getter for reads in the future.

library/Icinga/Util/Json.php
```diff
@@ -22,16 +22,57 @@ class Json
...
+        switch (json_last_error()) {
+            case JSON_ERROR_NONE:
+                return $encoded;
+
+            /** @noinspection PhpMissingBreakStatementInspection */
+            case JSON_ERROR_UTF8:
+                if ($autoSanitize) {
+                    return static::encode(static::sanitizeUtf8Recursive($value), $options, $depth);
+                }
+
```
lippserd: I think you need a `// Fallthrough` comment here.

PR #3449, opened 2018-05-02
Flippable service grid axes

### LINKED ISSUE #2640: [dev.icinga.com #13471] Servicegrid Landscape Mode
This issue has been migrated from Redmine: https://dev.icinga.com/issues/13471

**Created by pooh on 2016-12-08 10:24:07 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _Backlog_
Last Update: _2016-12-08 11:18:49 +00:00 (in Redmine)_

---

I like the servicegrid display http://whatever/icingaweb2/monitoring/list/servicegrid very much as a status summary of how all the machines are performing. I use it without "?problems" at the end, to show all machines and services, whatever the colour of the indicators.

However, I (like most people, I suspect) am monitoring more machines than I have services, so the grid ends up being tall and narrow, whereas most display monitors are wider than they are high, so it would be nice to have the facility to show the service grid with machine names across the top and services down the side, to fit more neatly onto a widescreen monitor.

nilmerg:
flipped-servicegrid-axes

resolves #2640

### DIFF
--- a/application/views/scripts/joystickPagination.phtml
+++ b/application/views/scripts/joystickPagination.phtml
@@ -2,10 +2,6 @@
 
 use Icinga\Web\Url;
 
-if ($xAxisPaginator->count() <= 1 && $yAxisPaginator->count() <= 1) {
-    return; // Display this pagination only if there are multiple pages
-}
-
 $showText = $this->translate('%s: Show %s %u to %u out of %u', 'pagination.joystick');
 $xAxisPages = $xAxisPaginator->getPages('all');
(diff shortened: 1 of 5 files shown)


nilmerg:
1. This does not seem to me being related to these changes. I'm able to reproduce the same on the master. Anyway, I'll try to fix this.

2. Well, I wanted this to behave like a boolean parameter `?flipped` just like `?problems`. But our url rendering does not handle this.. I'll change the behaviour :(

nilmerg:
Both issues fixed.

Issue #1298, opened 2015-01-13
[dev.icinga.com #8194] Monitoring: Interpret performance data ranges (thresholds)


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/8194

**Created by mjentsch on 2015-01-13 09:07:02 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _Backlog_
Last Update: _2016-07-05 15:27:31 +00:00 (in Redmine)_

---

Critical or Warning state of a perfdata set is currently only decided based on the value being smaller or bigger than the critical value and does not recognize '@' or '~' chars.

**Changesets**

_2016-04-26 14:26:06 +00:00_ by _aklimov_ 7c276a99f82b09d2cd503d338f3f2649e9ef76b7

    Implement ThresholdRange
    
    refs #8194

_2016-04-27 08:26:49 +00:00_ by _aklimov_ 7d27a079aaa0d5eb3e23814675225e768ee928dc

    ThresholdRange: trim() raw ranges
    
    refs #8194

_2016-04-27 08:30:02 +00:00_ by _aklimov_ dc6a25525134a49d86fafb76ceb70a0dcf466590

    ThresholdRange: use substr() instead of strpos() where more performant
    
    refs #8194

_2016-04-27 08:34:49 +00:00_ by _aklimov_ a0d62fdf3a685ca2f9eaaff462661654dfa1209b

    ThresholdRange: reword doc
    
    refs #8194

_2016-04-27 09:30:23 +00:00_ by _aklimov_ e76d65365704d3a1bb7e78bfe6888e46cfe701e7

    ThresholdRange::setInverted(): default to true
    
    refs #8194

_2016-04-27 09:35:44 +00:00_ by _aklimov_ f78b4ffe7424d00e9cb44a70c719239fbd540596

    Perfdata: simplify code design
    
    refs #8194

_2016-04-27 10:24:31 +00:00_ by _aklimov_ 73e404efef2afcc93e679e7b82f847a350b74d70

    ThresholdRange: trim(floatval()) -> floatval(trim())
    
    refs #8194

_2016-04-27 15:46:24 +00:00_ by _aklimov_ 1860a7c549ce2ba03cbfb055c738ff7ae632859e

    Perfdata: use ThresholdRange
    
    refs #8194

_2016-04-27 16:13:14 +00:00_ by _aklimov_ bb2a5b2d08ea818c1af1129fa3b50c6656b7504d

[...]

PR #3451, opened 2018-05-03
Interpret performance data threshold ranges

### LINKED ISSUE #2587: [dev.icinga.com #12885] Performance Data omits Warn/Crit values
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12885

**Created by rajo01 on 2016-10-10 17:58:55 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _(none)_
Last Update: _2016-10-10 17:58:55 +00:00 (in Redmine)_

---

I'm using the check\_pcmeasure2.pl script to measure some temperature values. On the command line, they display as expected:

        /usr/lib/nagios/plugins/check_pcmeasure2.pl -H 192.168.xx.xx -S com1.1 -w 16:24 -c 10:30
        PCMEASURE WARNING - value = 15.4 | value=15.4;16:24;10:30

Also, when looking into the icinga\_servicestatus table, the complete data is logged there, e.g.:

        PCMEASURE OK - Temperatur = 22.7  |             | temperatur=22.7;15.1:32.1;15:32    

But for whatever reason, icingaweb2 displays just 0 ... Any ideas?

For the record, I'm using Icinga2 2.5.4-1 and icingaweb2 2.3.4 of packages.icinga.org

**Attachments**

* [broken_perfdata.png](https://dev.icinga.com/attachments/download/3679/broken_perfdata.png) rajo01 - _2016-10-10 17:57:35 +00:00_

---

**Relations**:

* blocks [#12885](https://dev.icinga.com/issues/12885)

nilmerg:
proper-threshold-interpretation

fixes #2587

### DIFF
--- a/modules/monitoring/library/Monitoring/Plugin/Perfdata.php
+++ b/modules/monitoring/library/Monitoring/Plugin/Perfdata.php
@@ -230,21 +230,21 @@ class Perfdata
     /**
      * Return this performance data's warning treshold
      *
-     * @return  string
+     * @return  ThresholdRange
      */
     public function getWarningThreshold()
     {
-        return (string) $this->warningThreshold;
+        return $this->warningThreshold;
     }
 
     /**
      * Return this performance data's critical treshold
      *
-     * @return  string
+     * @return  ThresholdRange
      */
     public function getCriticalThreshold()
     {
-        return (string) $this->criticalThreshold;
+        return $this->criticalThreshold;
     }
 
     /**
--- a/modules/monitoring/library/Monitoring/Plugin/Perfdata.php
+++ b/modules/monitoring/library/Monitoring/Plugin/Perfdata.php
@@ -111,7 +111,7 @@ class Perfdata
     {
         if (empty($perfdata)) {
             throw new InvalidArgumentException('Perfdata::fromString expects a string with content');
-        } elseif (false === strpos($perfdata, '=')) {
+        } elseif (strpos($perfdata, '=') === false) {
             throw new InvalidArgumentException(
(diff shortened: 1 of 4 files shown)


nilmerg:
Rebased the old branch with the current master. Looks good so far.

As @Al2Klimov mentioned in the old issue #1298 only tests are missing. Since we've decided to only show the raw threshold values ( @lippserd ) we don't need any pretty printing.

The advantage of these changes is that we're able to visualize the state on pie-charts correctly. Without these any pie-chart would be green even if the state is warning/critical.

lippserd:
Could you add a test case for invalid ranges as well please? Everything else looks good.

nilmerg:
Wasn't sure how we're supposed to handle invalid ranges as they're part of a plugin's perfdata and as such already evaluated. However, there's a test now which checks if invalid notations are at least rendered as-is. Evaluation is not tested as I really don't want to define what's wrong or right in such cases.

```
s\Icinga\Module\Monitoring\Plugin\ThresholdRange
 [x] From string properly parses double exclusive ranges
 [x] Contains correctly evaluates double exclusive ranges
 [x] From string properly parses single exclusive ranges
 [x] Contains correctly evaluates single exclusive ranges
 [x] From string properly parses inclusive ranges
 [x] Contains correctly evaluates inclusive ranges
 [x] From string properly parses empty thresholds
 [x] Contains evaluates everything to true for empty thresholds
 [x] Invalid threshold notations are rendered as is
```

lippserd:
Thanks for the extensive tests. I'll merge the PR now.

PR #3455, opened 2018-05-11
WIP: Enable referencing other columns in filters

lager1:
When used in icinga-director filter expressions, another column may be referenced using php variable notation. Can be used to dynamically evaluate content when used in sync rules.

Example use in the attached picture.
example

### DIFF
--- a/library/Icinga/Data/Filter/FilterExpression.php
+++ b/library/Icinga/Data/Filter/FilterExpression.php
@@ -182,7 +182,8 @@ class FilterExpression extends Filter
         }
 
         // referencing another column using php variable notation
-        if (strlen($expression) > 2 && $expression[0] == '$' && $expression[1] == '{' && $expression[strlen($expression) - 1] == '}') {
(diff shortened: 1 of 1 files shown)


lippserd:
Hi,

Thanks for the PR. I like the idea but I don't think that it's good to make this standard behavior. Because the filter classes and filter editor are used in many places and many modules we can't guarantee that exposing all values with that feature is safe. We could guard this by only allowing access to properties which are enabled as filter columns in the filter editor for example. Further, there must be some sort of escaping for this notation. I mark this as WIP and we'll see what we can do here.

Best,
Eric

nilmerg:
Hi,

as Eric already told you, the way this PR proposes this change is not how we'd like to implement this. It's also not complete without support in our data backends, because the current way does only support this if it's Icinga Web 2 which is evaluating the filter. Which it rarely is.

So, due to the age and state of this PR and the fact that we didn't have other requests in this regard I'll close it now. Please open an issue if you still need this as an addition but with a proposal how this should also affect our data backends. Thank you!

PR #3457, opened 2018-05-14
Proper results when filtering for linked objects

### LINKED ISSUE #2934: Filter to exclude hosts/services that are member of a specific group
Hi,

I couldn't find any doc or issue on this subject, please apology if already answered out somewhere.

Thanks a lot for all the work !

## Expected Behavior
When using filters on hostgroups, hosts being member of more than one hostgroup :
When filtering out a hostgroup (eg : `"hostgroup_name!=foo"`), hosts matching this hostgroup (even if member of other hostgroups) should not be displayed.

## Current Behavior
If host `a` is member of hostgroup `foo` AND hostgroup `bar`, using filter `"hostgroup_name!=foo"`, host `a` is still displayed.
If host `a` is member of hostgroup `foo` and no other hostgroup, filtering works as expected.

## Context
I couldn't find any other way to display strictly "all hosts but hosts which are member of group foo".
Encountered both in views (eg: `/icingaweb2/monitoring/list/services?service_problem=1&hostgroup!=foo`) and roles permissions (see documentation: [06-security](https://github.com/Icinga/icingaweb2/blob/master/doc/06-Security.md#example-2-hostgroups)).

## Your Environment
* Icinga Web 2 version: 2.4.1
* Modules used: doc (2.4.1), monitoring (2.4.1), pnp (1.0.1)
* Icinga2 version used: r2.6.0-1
* Operating System and version: debian stretch (9.0), linux 4.9.0-2-amd64
* Enabled features: checker command ido-mysql mainlog notification perfdata

nilmerg:
resolves #2934

### DIFF
--- a/modules/monitoring/library/Monitoring/Backend/Ido/Query/ContactgroupQuery.php
+++ b/modules/monitoring/library/Monitoring/Backend/Ido/Query/ContactgroupQuery.php
@@ -23,6 +23,11 @@ class ContactgroupQuery extends IdoQuery
      */
     protected $groupOrigin = array('hosts', 'members', 'services');
 
+    protected $subQueryTargets = array(
+        'hostgroups'    => 'hostgroup',
+        'servicegroups' => 'servicegroup'
+    );
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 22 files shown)


nilmerg:
The following should be working and are intended as example implementations:
* `HoststatusQuery`
* `ServicestatusQuery`
* `Hostgroupquery`

The `ServicegroupQuery` is not yet finished. However, in case we really only support `hostgroup_name` and `servicegroup_name` this does not matter. That's because in this case only the "`! $correct`" branch is required. (And working.)

Also still missing is the automatic merge of similar filters by using `HAVING`.

PR #3465, opened 2018-05-29
Render multiline in custom variables correct.

Mikesch-mp:


lazyfrosch:
This should be done in `modules/monitoring/application/views/helpers/Customvar.php`

Mikesch-mp:
will send new pr then

PR #3468, opened 2018-06-02
Render multiline in custom variables correct.

Mikesch-mp:


Mikesch-mp:
As you wish master :)

lippserd:
Thank you sir 😆

nilmerg:
May I ask what the initial reason was to implement this? Just tried to reproduce this by adding a custom variable with newlines in it and Icinga 2 wrote these as real newlines into the database. If it were only about real newlines and not also `\n´ and ´\r\n´ I'd avoid to use our helper because that's what caused #3636 now.

Mikesch-mp:
I had a custom variable with some information over multiple lines, imported by director from a mssql database. But no matter what i did it was always displayed as one line even if i put it into `{{{ ... }}}`.

nilmerg:
Gnaaaah, thanks for mentioning the director. It's exporting [literal `\n` instead of `{{{ ... }}}`](https://github.com/Icinga/icingaweb2-module-director/blob/70dfc9dae98c6638259810156dd40ad3814a5882/library/Director/IcingaConfig/IcingaConfigHelper.php#L132) enclosed strings. :sob:

Thomas-Gelf:
@nilmerg: I don't see the problem, please open a related issue in case there is one. Custom variables with newlines are rendered as such by the Director. `{{{ ... }}}` is just another way of writing strings, helping to avoid very long lines. This is a real newline...

`vars.whatever = "multi\nline"`

...should end up as a newline in the IDO - and will be shown as a newline in Icinga Web 2.

@Mikesch-mp: in case you think there is a problem please share more details: what does your import look like, what does the related rendered configuration look like? If you put it into `{{{ ... }}}` then the Director is not involved.

@nilmerg: we preferred to render `" ...\n..."` with `\n` being (in theory) exactly what one wrote into that custom variable. Using `{{{ ... }}}` would involve black magic making assumptions for `\r`, `\r\n` and `\n\r`, rendering "simple" strings seemed to be the more conservative option. Please let me know if you have a better proposal.

Thanks,
Thomas

nilmerg:
There's no problem. I just thought that's the reason why we need to transform literal `\n`. Though, as you've mentioned, Icinga 2 does write real newlines into the IDO the way the director currently renders multi-line strings. I guess just utilizing `nl2br` resolves #3636 and does not interfere here.

Mikesch-mp:
The Director import was only the start to search why the newlines a not honored. When i tested/fixed it i wrote also conf files with 
```
vars.test1 = "Test1\nTest2"
vars.test2 = "Test1\r\nTest2"
vars.test3 = {{{
Test1
Test2
}}}

And none of them (if i remember correctly) worked as expected.

The Problem in the issue #3636 is because of escaping the backslash for "icinga2" and the name after the backslashes that starts with an "r". The web renderer should check if there is an "\\" before replacing well known newlines or other controll sequences.

nilmerg:
The escape sequence is unfortunately only for Icinga 2. Icinga Web 2 does not receive this anymore. The only solution would be to escape the escaped backslash by the user or to just transform real newlines by Icinga Web 2. The latter seems more reasonable to me. :thinking:

PR #3475, opened 2018-06-15
Make all grapher hooks available in the view

### LINKED ISSUE #3473: Trying to add multiple Grapher Hooks fails
I'm currently trying to create a poc for implementing the business process tile renderer as a grapher in Icingaweb2.
To achieve that I created `businessprocess/library/Businessprocess/ProvidedHook/Grapher.php`:
```
<?php

namespace Icinga\Module\Businessprocess\ProvidedHook;

use Icinga\Application\Hook\GrapherHook;
use Icinga\Module\Businessprocess\Renderer\TileRenderer;
use Icinga\Module\Businessprocess\State\MonitoringState;
use Icinga\Module\Businessprocess\Storage\LegacyStorage;
use Icinga\Module\Businessprocess\Web\Url;
use Icinga\Module\Monitoring\Object\MonitoredObject;
use Icinga\Module\Monitoring\Object\Service;

class Grapher extends GrapherHook
{

    private $storage;

    public function init()
    {
        try {
            $this->storage = new LegacyStorage(
                Config::module('businessprocess')->getSection('global')
            );
        } catch (Exception $e) {
            // Ignore and don't display anything
        }

        $this->hasPreviews = true;
    }

    public function has(MonitoredObject $object)
    {
        return $object instanceof Service && $object->check_command == 'icingacli-businessprocess';
    }


    public function getPreviewHtml(MonitoredObject $object)
    {
        if (!$this->has($object) || !$this->storage) {
            return '';
        }

        $bpName = $object->_service_icingacli_businessprocess_process;

        $bp = $this->storage->loadProcess($bpName);
        $node = $bp->getNode($bpName);

        MonitoringState::apply($bp);

        $renderer = new TileRenderer($bp, $node);
        $renderer->setBaseUrl(Url::fromPath('businessprocess/process/show?config=' . $bpName . '&node=' . $bpName));

        return '<div class="icinga-module module-businessprocess"><h2>Business Process</h2>' . $renderer->render() . '</div>';
    }
}
```

And added this line to `run.php`:
```
$this->provideHook('grapher');
```

The result works well:
grafik

## Expected Behavior
Issues arise when trying to enable the Grafana module which brings it's own Grapher:
Grafana graphs and business process tiles are both rendered.

## Current Behavior
Only the business process is rendered. Grafana is not visible even on services where my grapher's `has()` method returns false.
When renaming my `Grapher` class to anything else Grafana shows up again but I can't get my tile to render. Even when disabling Grafana again my hook only seems to work, when the class is precisely named `Grapher`.

## Possible Solution
Not sure. There seems to be an issue with how the Grapher objects are created and rendered.

1. See my initial description. Just create/edit the two files
2. Create a service with the icingacli-businessprocess command (the business process needs to have the same name as the root node. It's a poc and I haven't found out how to get the business process the node referenced in the check command belongs to).
3. Experiment with en-/disabling the Grafana module and renaming the Grapher fil.e

## Your Environment
* Business Process: 2.1.0
* Grafana: 1.2.1
* Version used (`icinga2 --version`): 2.8.4
* Operating System and version: Debian 9.4, PHP 7.0

matthenning:
Result using my PoC business process grapher:

grafik

fixes #3473

### DIFF
--- a/modules/monitoring/application/views/scripts/show/components/grapher.phtml
+++ b/modules/monitoring/application/views/scripts/show/components/grapher.phtml
@@ -1,4 +1,6 @@
-<?php if (isset($grapher)) {
-  echo $grapher->getPreviewHtml($object);
+<?php if (isset($graphers)) {
+  foreach ($graphers as $grapher) {
+    echo $grapher->getPreviewHtml($object);
+  }
 } ?>
 
--- a/modules/monitoring/library/Monitoring/Web/Controller/MonitoredObjectController.php
+++ b/modules/monitoring/library/Monitoring/Web/Controller/MonitoredObjectController.php
@@ -47,7 +47,7 @@ abstract class MonitoredObjectController extends Controller
             $this->view->tickets = Hook::first('ticket');
         }
         if (Hook::has('grapher')) {
-            $this->view->grapher = Hook::first('grapher');
+            $this->view->graphers = Hook::all('grapher');
         }
     }
 


lippserd:
Hi,

Thanks for the PR. We'll look at this one asap.

Cheers,
Eric

PR #3489, opened 2018-06-27
Don't update Parallels Tools automatically

### LINKED ISSUE #3488: Vagrant box broken on MacOS + Parallels
## Expected Behavior

A fresh IW2 Vagrant box works OOTB.

## Current Behavior

A fresh IW2 Vagrant box doesn't work at all on MacOS + Parallels.

## Possible Solution

To be done.

1. Install Parallels on MacOS
2. Clone this repository
3. Run `vagrant box update`
4. Run `vagrant up`
5. Run `vagrant reload` two times

## Context

Renewal of my Vagrant box for running some unit tests on an up-to-date platform.

## Your Environment

c633c86

## Commentary

On the first run the VM is initialized and successfully provisioned via Puppet. Then [`.puppet/manifests/parallels-upgrade-guest-tools.sh`](https://github.com/Icinga/icingaweb2/blob/c633c86db727289a5c85270eddf3559065baf77f/.puppet/manifests/parallels-upgrade-guest-tools.sh) behaves as expected and updates the kernel due to the failed installation of the Parallels Guest Tools due to the failed installation of `kernel-devel-$(uname -r)` due to a mismatch of the installed/running kernel version w/ the only version of `kernel-devel-...` available from the CentOS package repo.

On the first `vagrant reload` the Parallels Guest Tools fail to be installed (by `.puppet/manifests/parallels-upgrade-guest-tools.sh`) and restoring the previous version also fails: `Failed to restore previous version '13.3.0.43321' (retcode 143)`

On the second `vagrant reload` Vagrant finally (too late) tells prominently that something is broken:

```
Failed to mount folders in Linux guest. This is usually because
the "prl_fs" file system is not available. Please verify that
Parallels Tools are properly installed in the guest and
can work properly. If so, the VM reboot can solve a problem.
The command attempted was:

mount -t prl_fs -o uid=`id -u vagrant`,gid=`getent group vagrant | cut -d: -f3` vagrant /vagrant
mount -t prl_fs -o uid=`id -u vagrant`,gid=`id -g vagrant` vagrant /vagrant
```

Al2Klimov:
fixes #3488


Al2Klimov:
We've updated the guest tools automatically "since day 1" (f3df8db) and not due to an error caused by not updating them. So I expect no errors due to not updating them at all.

dnsmichi:
AFAIK the bento boxes have the parallel tools installed already, I don't see any problems without updating the tools inside the "icinga-vagrant" project.

PR #3490, opened 2018-06-27
Update LoginForm.php - Add failed auth logging

willhseitz:
Adds logging for failed authentications with source IP. Useful for IDS such as Fail2Ban.

### DIFF
--- a/application/forms/Authentication/LoginForm.php
+++ b/application/forms/Authentication/LoginForm.php
@@ -86,11 +86,15 @@ class LoginForm extends Form
      */
     public function onSuccess()
     {
-        $ip = $_SERVER['REMOTE_ADDR'];
+        $ip = null;
+        if (Config::app()->get('logging', 'log_auth_ip')) {
+                $ip = $_SERVER['REMOTE_ADDR'];
+        }
         $auth = Auth::getInstance();
         $authChain = $auth->getAuthChain();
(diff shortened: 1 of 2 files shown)


dnsmichi:
Logging of IP addresses might be forbidden in terms of data retention laws in several countries. Such a thing should be explicitly enabled by the admin imho, and opt-out for those who cannot use this addition.

willhseitz:
Understood, I'll update request to log entries without IP info by default, and add a flag in Logging options to enable IP logging

nilmerg:
Hi,

sorry for this very late response. I guess I also have no good news for you here, because this implementation suffers from three issues:

* Having this as part of class `LoginForm` has no effect in case repeated attempts to authenticate with the API fail
* Logging this with level `warning` won't probably be ever seen because the default log level is `error`
* Since the release of the [audit module](https://github.com/Icinga/icingaweb2-module-audit) this is especially nothing which should be logged with the application log (and logging logins is already supported now)

If you are still interested to have failed login attempts being logged, please open a new PR and implement it using the `AuditHook::logActivity` static method. Place this in method `AuthChain::authenticate` at the end and it also covers API authentication attempts. The configuration shouldn't be necessary as well as this is supposed to be part of the audit module. (https://github.com/Icinga/icingaweb2-module-audit/issues/7)

Thanks,
Johannes

PR #3491, opened 2018-06-27
WIP: Implement transiflex

lazyfrosch:
I wanted to tryout Transifex for translations for Icinga Web and created an account:

https://www.transifex.com/icinga/icingaweb2

If anyone wants to help, join Transifex above and ping me for questions.

Actions:
- [x] Include and handle POT files
- [x] Upload current state to transifex
- [ ] Use relative paths for files (no `/usr/share/...`)
- [ ] Update local documentation

Handling in the future:
* No more manual fiddling with po files
* Update POT files when code changes (can be automated)
* Work on the webserver
* Pull PO files before and releases and compile them

Opinions?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 !.puppet*
 !.travis.yml
 !.mailmap
+!.tx/
 
 # Testing - created by test/setup_vendor.sh
 /vendor/
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,10 @@ php:
   - '7.2'
   - nightly
 
+env:
+  global:
+    - secure: "HPrbKBgJTctoebpG/lBxPQ/6Jt0nqAYPG+G/9Qkvzbo5LS7XyCjfkARNY2aRvaOcaDo/RmyBx3OdxVXXVY/DxwlF0J/E1VoxD5w/lQOnOHGRK5CrH+L1ztLyp/mvQcI091oZgkAsJeeb7bQzyLXb/5ayFz2UKiYdIn2E7MpgjqI="
+
 matrix:
   allow_failures:
     - php: nightly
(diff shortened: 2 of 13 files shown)


nilmerg:
Had this much traction in the past months? Doesn't seem so to me, but I don't have an account.

**If** we're integrating this, we should consider guarding all translation calls against malicious strings (as I've already seen and used too many non-escaped `$this->translate()` calls) And having our internationalization this open makes this even more likely to be a serious problem.

lazyfrosch:
It was an experiment, but makes absolutely sense for use to have better translations...

There should be a way to check for escaped strings.

nilmerg:
Of course it makes sense, I'm with you here. But if it's not used I don't see any advantage having it. Our previous translation maintainers did not have much interest here on Github. Adding another platform just to see the same result doesn't make any sense to me, unfortunately. So if there really was *no activity at all* in the past months on transiflex I'd see this experiment as failed.

lazyfrosch:
It is a way better UI than any local software and will allow non-technical people to help us translating.

We told nobody we even use Transifex, so why should anyone? It see it merely as a benefit that translations doesn't need to be done by developers.

nilmerg:
> We told nobody we even use Transifex, so why should anyone?

Then we should do so?

To be clear: I like the idea to have more recent/complete translations. Though what I don't like is the idea of a platform that is either inactive or requires a full time job to maintain. (And that's what it looks like now, previously I had the impression this is some kind of crowdsource platorm)

lazyfrosch:
This is up for adoption when someone wants to implement it

nilmerg:
Transifex is unfortunately not an option for us (anymore). Regarding their terms of use we're not open-source enough to make use of the free plan.

The automation however is still on track, though now in a separate project: https://github.com/Icinga/L10n

PR #3504, opened 2018-07-05
Fix collapsible sidebar issues

### LINKED ISSUE #3187: Collapsible Sidebar Issues
This is a collection of some smaller issues i discovered regarding the collapsible sidebar implementation.

1) When the sidebar is collapsed, the navigation items scroll out of the visible area, when the search input is focussed. This could be reproduced in Safari. Expanding and collapsing again corrects this temporarily until the search input will be focussed again.

bildschirmfoto 2017-12-04 um 17 57 18

2) Active first-level nav item should be better distinguishable from non active ones (When first level index is selected).

artboard 2

3) The toggle-sidebar "lug" could be spaced wider and taller to better align with the nav icons, especially in collapsed state. It should also get a little offset to bottom to align with the status bar, when available.

artboard 3

4) The second level fly out menu appears after a slight delay after hovering. While the delay seems necessary for addressing the "scrollbar issue" for hiding the fly out, it should appear immediately when hovering.

5) In the original draft the flyout menu featured a caret on the left pointing to the according first level menu item. This doesn't harm usability too much, but should improve visual clarity, especially when another level-2 item is active.

artboard 4

## Your Environment
Icinga Web 2 2.5.0, screenshot taken in Safari 10.11.1 on macOS

theFeu:
fixes #3187

### DIFF
--- a/library/Icinga/Web/Navigation/NavigationItem.php
+++ b/library/Icinga/Web/Navigation/NavigationItem.php
@@ -38,6 +38,13 @@ class NavigationItem implements IteratorAggregate
     protected $active;
 
     /**
+     * Whether this item is selected
+     *
+     * @var bool
+     */
+    protected $selected;
+
+    /**
      * The CSS class used for the outer li element
      *
      * @var string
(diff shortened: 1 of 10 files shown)


theFeu:
Problems that still need fixing:

1. In Safari: When the sidebar is collapsed and the search input is clicked the menu bar icons slip to the left

2. Add icon with first letter of the li if there is no icon set

theFeu:
2. Add icon with first letter of the li if there is no icon set

theFeu:
Found a new issue that came up with the flyout caret introduced with commit 327a1298530268cb4f9e940f8dc8149ac297612b:
it's always body-bg-color (white), and it shows when one of the items are hovered (light-blue)

- [ ] done

theFeu:
- [ ] done

When hovered the background of an item will change to light blue.
If the sidebar is not collapsed then the item will hide the 1px border on the left,
if collapsed it is positioned one pixel further to the right so the border is still visible (expected behaviour).

theFeu:
About
> Also I don't think that it is necessary to add the class active-nav-level-1.

We need some classe inserted by the JS to indicate that no children ( => nav-items) are active.
Since it is not possible to do that in pure CSS we need some class, the name "active-nav-level-1" is a bit bloated, yes - but we do need some class for this.

theFeu:
> The new active style for first level nav items is odd and should not be JS only.
If you click on a first level nav item it is styled the "old" way and gets changed when JS kicks in.

About this: With the current 'active'  state it is impossible to distinguish between it being active (the menu section is open) and it being active (this nav-item is selected).
With only CSS it is not possible to check wether any of the lvl-2 nav-items are clicked or not, since they are down the cascade.

If you turn the JS off, it is still pretty clear where you are though, since the sections are all unfolded and the currently active one is in a darker shade of grey without an item being highlighted even further in blue. imho this is sufficient.

lippserd:
Well done. Thanks feu!

dnsmichi:
Woah, awesome ❤️

PR #3511, opened 2018-07-06
WIP: Redo the way we generate PDF files

theFeu:
What still needs to be done:

**1) The generated PDFs' layout is still a bit wonky...**
The command doesn't support removing the generated headers and footers from the PDF, so there is always the date in the header and the temporary files' name and pagination in the footer. This needs to be removed with the Chrome Devtools. As discussed offline we might need some JS to remove those.
We still use the same CSS for the PDF gen, what causes a 'Page 0' footer on every page.

**2) The efficiency is not optimal either**
I think this might be because we always open a new browser, instead of tabs. We should probably do some research on how to improve this.

**3) //todos are still in place**

### DIFF
--- a/library/Icinga/File/Pdf.php
+++ b/library/Icinga/File/Pdf.php
@@ -4,10 +4,9 @@
 namespace Icinga\File;
 
 use Dompdf\Autoloader;
-use Dompdf\Dompdf;
-use Dompdf\Options;
 use Icinga\Application\Icinga;
 use Icinga\Exception\ProgrammingError;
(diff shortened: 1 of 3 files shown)


theFeu:
Potentially interesting repo to look into: https://github.com/chrome-php/headless-chromium-php

theFeu:
Good for the understanding documentation: https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-printToPDF

nilmerg:
This is obsolete since https://github.com/Icinga/icingaweb2-module-pdfexport?

PR #3519, opened 2018-07-10
Pdo\Abstract: Properly handle incomplete error messages

### LINKED ISSUE #3249: MySQL SSL connection error brings up no explanation
When connecting to a MySQL Server via SSL, the error message can be empty

Set in config:
* use_ssl
* ssl_ca
* host uses IP instead of a hostname

Of course it works when the FQDN host is used...

## Expected Behavior
Error should be something like:
```
SQLSTATE[HY000] [2002] SSL error, hostname does not match ...
```

## Current Behavior
Error is:
```
SQLSTATE[HY000] [2002]
```

## Your Environment
* Icinga Web 2 version and modules (System - About): 2.5.0
* Version used (`icinga2 --version`): 2.7.0
* Operating System and version: CentOS 7 + PHP 7.1 SCL

nilmerg:
fixes #3249

### DIFF
--- a/library/vendor/Zend/Db/Adapter/Pdo/Abstract.php
+++ b/library/vendor/Zend/Db/Adapter/Pdo/Abstract.php
@@ -133,10 +133,16 @@ abstract class Zend_Db_Adapter_Pdo_Abstract extends Zend_Db_Adapter_Abstract
             $this->_connection->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
 
         } catch (PDOException $e) {
+            $message = $e->getMessage();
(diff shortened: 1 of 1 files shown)


lippserd:
Hi Johannes,

Thanks for the analysis and the provided fix. Could you please also push this one to https://github.com/Icinga/icingaweb2-vendor-zf1

Cheers,
Eric

nilmerg:
@lippserd I'll happily do that, once I've got the permission to at least push branches.

nilmerg:
Nope
> ERROR: Permission to Icinga/icingaweb2-vendor-zf1.git denied to nilmerg.

PR #3529, opened 2018-07-18
Introduce PdfexportHook

lippserd:


library/Icinga/Application/Hook/PdfexportHook.php
```diff
@@ -0,0 +1,25 @@
...
+ * Base class for the PDF Export Hook
+ */
+abstract class PdfexportHook
+{
+    /**
+     * Get whether PDF export is supported
+     *
+     * @return  bool
+     */
+    abstract public function isSupported();
```
Thomas-Gelf: This could default to true, only implementations with third-party dependencies might want to override the default.

library/Icinga/File/Pdf.php
```diff
@@ -61,12 +62,29 @@ public function renderControllerAction($controller)
             'src="' . Icinga::app()->getBootstrapDirectory() . '/img/',
             $html
         );
+
+        $request = $controller->getRequest();
+
+        if (Hook::has('Pdfexport')) {
```
Thomas-Gelf: The combination of Hook::has and (any, not first) `Hook->isSupported` could be moved to a static method in the base abstract Hook class - like `PdfExportHook::implementationIsAvailable()` or similar

library/Icinga/File/Pdf.php
```diff
@@ -61,12 +62,29 @@ public function renderControllerAction($controller)
...
+            $pdfexport = Hook::first('Pdfexport');
+
+            if ($pdfexport->isSupported()) {
+                $pdfexport->streamPdfFromHtml($html, sprintf(
```
Thomas-Gelf: What about moving this to a dedicated `PdfExport` class? `File/Pdf` has similar logic right now and IMO that's also not the right place. But this Hook-based logic is one more logical component you wouldn't expect here.

library/Icinga/Application/Hook/PdfexportHook.php
lippserd: This hook is for our chrome-headless pdfexport module in the first place and this one has deps.

library/Icinga/File/Pdf.php
lippserd: Again, this hook is only for our own purpose right now. I expect that there is exactly one hook which provides pdfexort. The dompdf rendering is stuff of our framework and will die once we've finished our pdfexport module. If there are more pdfexport hooks available the user would need a configuration option as well. Far out of scope at the moment.

library/Icinga/File/Pdf.php
lippserd: You already said that `File/Pdf` is already full of logic which you would not expect there. So, why bother?

library/Icinga/Application/Hook/PdfexportHook.php
Thomas-Gelf: Others may not, so why force implementors to implement a useless method?

library/Icinga/File/Pdf.php
Thomas-Gelf: See below. Btw, I wouldn't make it die but move it to a dedicated module once deprecated - for those unable or unwilling to install Chromium for whatever reason. Compared to DomPDF it's a "monster" dependency. The configuration option is a good point for not leaving such logic in `File\Pdf`, I also had that in my mind but didn't want to clutter this comment with too many details.

library/Icinga/File/Pdf.php
Thomas-Gelf: Because it is a Hook, it's per definition not internal. If you want it to be internal, do not create a Hook ;-)

library/Icinga/Application/Hook/PdfexportHook.php
lippserd: If there's a pdfexport module other than our own on the horizon which does not have deps for rendering PDFs (good luck on that one), I'll happily change the method.

library/Icinga/File/Pdf.php
lippserd: Come on, dompdf is a monster dep as well with all the libs it requires. You know all the bugs we have with it and that it's slow. It will die. Again, if there's a module other than our own in the future or if our module requires changes in this hook, we'll change it. For now, there's no point in doing so.

library/Icinga/File/Pdf.php
lippserd: I don't get it. `File\Pdf` is a useless library class for others. It would make sense to rewrite all that stuff, sure. The hook could introduce a new method which just renders HTML to PDF though. The current implementation is for Web 2's PDF export only. And there's no need for more at the moment.

library/Icinga/File/Pdf.php
Thomas-Gelf: Then there is no point for a hook, file_exists for chromium in a bunch of well-known dirs should suffice.

PR #3532, opened 2018-07-18
Fetch check_interval property of monitored objects, too

Al2Klimov:


modules/monitoring/library/Monitoring/Backend/Ido/Query/HoststatusQuery.php
```diff
@@ -38,6 +38,7 @@ class HoststatusQuery extends IdoQuery
             'host_address'          => 'h.address',
             'host_address6'         => 'h.address6',
             'host_alias'            => 'h.alias',
+            'host_check_interval'   => '(h.check_interval * 60)',
```
Al2Klimov: @dnsmichi Why the heck does Icinga 2 divide these by 60?!

dnsmichi: That's the old IDO schema. It only stores intervals as minutes, similar to 1.x. Icinga 2 has seconds as default base, and needs to divide these values to properly store them for addons and web interfaces. Which then calculates it into seconds again.

Al2Klimov: *Sigh*.

PR #3533, opened 2018-07-18
Add command audit

lippserd:


dnsmichi:
Imho the REST API only transport is a cool feature to soft force users to change. Then we can finally deprecate the command pipe at some point in the future.

Issue #1594, opened 2015-04-09
[dev.icinga.com #9028] Add support for Shibboleth as SSO auth provider


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/9028

**Created by mfriedrich on 2015-04-09 08:24:38 +00:00**

Assignee: _(none)_
Status: _New_ 
Target Version: _Backlog_
Last Update: _2015-07-03 14:45:03 +00:00 (in Redmine)_

---

Feature request from SIG-NOC meeting.

https://wiki.shibboleth.net/confluence/display/SHIB2/Home

Generally speaking the SSO provider handling is done with the webserver, passing only the 'entitlement' env var which contains filter restrictions. We had that implemented in a similar fashion for Classic UI in [\#909.](https://dev.icinga.com/issues/909)

PR #3544, opened 2018-07-26
POC: external attribute based authorization

wferi:
Hi,

This is a proof-of-concept patchlet I'm currently experimenting with in the hope of "fixing" #1594.
I don't know the internals of IcingaWeb2, so I've got no idea if this approach is sound, but it makes external entitlements available in `roles.ini` as group names, which is pretty much adequate for our purposes. Do IcingaWeb sessions expire eventually? That would be required for (delayed) revocation of group memberships in this simplistic approach. Comments welcome!

Thanks,
Feri.

### DIFF
--- a/library/Icinga/Authentication/User/ExternalBackend.php
+++ b/library/Icinga/Authentication/User/ExternalBackend.php
@@ -116,6 +116,9 @@ class ExternalBackend implements UserBackendInterface
             }
 
             $user->setUsername($username);
+            if (array_key_exists('entitlement', $_SERVER)) {
+                $user->setGroups(explode(';', $_SERVER['entitlement']));
+            }
             return true;
         }
 


Thomas-Gelf:
@wferi: my little [extragroups](https://github.com/Thomas-Gelf/icingaweb2-module-extragroups) module could provide what you're looking for

wferi:
@Thomas-Gelf: From reading its documentation, it could if it used semicolon for separator. Can you see a way?

ip1981:
FYI: [Sproxy Backend](https://github.com/ip1981/nixsap/blob/b0eb86793ee1913aa908f9b24950866be507fa85/modules/pkgs/icingaweb2/sproxy.patch).

nilmerg:
Hi, I guess what Thomas and Igor showed already made it clear that there are alternative ways of implementing this without patching Icinga Web 2. If this doesn't suit you or you think this should definitely a part of Icinga Web 2 please open a new issue and explain what your requirements are. We'll then may discuss a final solution worth adding.

wferi:
@nilmerg I checked again, and it looks like both the extragroups module and the Sproxy Backend has the comma separator baked in, so they aren't directly usable behind my web server module, which provides semicolon-separated values. But I certainly can patch them instead of Icinga Web 2, which will be easier to maintain.

PR #3553, opened 2018-08-08
Enable `$user:local_name$` placeholder in restrictions

### LINKED ISSUE #3493: Allow roles to filter for the currently logged in user
In nagios it was possible to limit the hosts/services visible to a logged in user based on contact information. e.g. if userA was logged into the webfrontend he would only see hosts and services with her/him assigned as a contact.
This allows users to see only systems they are actually responsible for instead of having to somehow filter away issues of system out of their responsibility.

It would be nice to have something similar in icingaweb2.

## Context

e.g. we assign hosts AND services custom variables via icinga director, that respresent 

- responsible <-- single string value
- deputy <-- single string value
- additional informational users <-- array of string values 

These vars are then pulled (among other vars) by notfication assign-where's to map the required notifications thus allowing easy deployment.

It would be great to have a functionality that allows e.g. filtering like 
`host.vars.responsible=="$currentUser$" || host.vars.deputy=="$currentUser$"||"$currentUser$" in host.vars.addition_users`

## Your Environment

* Version used (`icinga2 --version`): r2.8.4-1
* Operating System and version: ubuntu 16 LTS
* Enabled features (`icinga2 feature list`): 
Disabled features: compatlog elasticsearch gelf icingastatus influxdb livestatus opentsdb perfdata statusdata syslog
Enabled features: api checker command debuglog graphite ido-mysql mainlog notification
* Config validation (`icinga2 daemon -C`):
* If you run multiple Icinga 2 instances, the `zones.conf` file (or `icinga2 object list --type Endpoint` and `icinga2 object list --type Zone`) from all affected nodes.

oliverrahner:
resolves #3493

### DIFF
--- a/doc/06-Security.md
+++ b/doc/06-Security.md
@@ -216,13 +216,13 @@ results of this query instead:
 
 #### Username placeholder <a id="username-placeholder"></a>
 
-The string `$currentUser` is replaced by the local username (without the domain part) of the logged on user while evaluating restrictions.
+The string `$user:local_name$` is replaced by the local username (without the domain part) of the logged on user while evaluating restrictions.
 This can come in handy if you have some kind of attribute on host or service level defining which user is responsible for a certain host or service.
 
-#### Example <a id="username-placeholder-example"> </a>:
+**Example**
 
 ```
-monitoring/filter/objects = (__responsible=$currentUser$|__deputy=$currentUser$)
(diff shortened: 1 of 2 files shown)


lippserd:
Hi,

Thanks for the PR. We consider this for our next feature release.

Best,
Eric

oliverrahner:
Comments by @nilmerg have been worked in.

PR #3573, opened 2018-09-06
CSV timestamps can be exported human readable

Megatherium:
Timestamps in CSVs both from the WebUI and CLI can be exported as ISO8601 strings. The timezone can be configured under "My Account" or with the --timezone parameter.

Also minor annotation fixes.

### DIFF
--- a/application/forms/PreferenceForm.php
+++ b/application/forms/PreferenceForm.php
@@ -13,7 +13,6 @@ use Icinga\User\Preferences;
 use Icinga\User\Preferences\PreferencesStore;
 use Icinga\Util\TimezoneDetect;
 use Icinga\Util\Translator;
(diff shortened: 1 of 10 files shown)


Megatherium:
Hi,
I just saw that you commented on the PR. Thank you for all the criticism. I knew the code I submitted wasn't great but that didn't mean that I really knew how to make it better. Thanks for providing concrete pointers as to what needs fixing.
I'll take another stab at it although that basically means starting from zero as I haven't looked at the code for almost 9 months which is probably a good thing as I'll be less tainted by my old approach.

Thanks again.

nilmerg:
I'll close it then. Don't hesitate to open a new PR. :smiley:

PR #3578, opened 2018-09-24
Fix uninitialized array and module in Cli\Command

mxhash:
Can not do this in cli context:

    $this->Config('redis')

ERROR: ErrorException in /vagrant/icingaweb2/library/Icinga/Cli/Command.php:81 with message: array_key_exists() expects parameter 2 to be array, null given

### DIFF
--- a/library/Icinga/Cli/Command.php
+++ b/library/Icinga/Cli/Command.php
@@ -48,14 +48,15 @@ abstract class Command
     public function __construct(App $app, $moduleName, $commandName, $actionName, $initialize = true)
     {
         $this->app = $app;
-        $this->moduleName  = $moduleName;
-        $this->commandName = $commandName;
-        $this->actionName  = $actionName;
-        $this->params     = $app->getParams();
-        $this->screen     = Screen::instance($app);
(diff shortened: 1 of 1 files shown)


library/Icinga/Cli/Command.php
```diff
@@ -94,7 +95,7 @@ private function getMainConfig($file = null)
             return $this->config;
         } else {
             if (! array_key_exists($file, $this->configs)) {
-                $this->configs[$file] = Config::module($module, $file);
+                $this->configs[$file] = Config::module($this->moduleName, $file);
```
nilmerg: I know, you didn't touch this, but `getMainConfig` is not about module configs. You've faced this issue probably because the namespace of your CLI command is/was wrong and as such `Config` decided to call `getMainConfig`. Which is indeed bugged.

The actual required fix would probably be:
```suggestion
                $this->configs[$file] = Config::app($file);
```

mxhash:
The change in getMainConfig was only to fix the uninitialized variable. Had only issued with the array_key_exists. Applied your suggestion.

PR #3585, opened 2018-10-02
Feature/enhanced ldap logging

nilmerg:
Makes the LdapConnection a litte bit more verbose.


nilmerg:
Now also the logger is set up earlier in the bootstrap process, because:

```
Previously only ERROR got logged (to syslog only) up until
setupLogger() was called. This prevented any other message
level from being logged. It's now required to be able to
log in the Auth singleton. It also fixes the problem that
during setupSession() some warnings were never visible
due to the same reason.
```

PR #3588, opened 2018-10-09
Fix problems the world has and ever will

nilmerg:


lippserd:
Hi Johannes,

Thanks for the changes. Could you please create dedicated forms instead of extending the controller?

Best,
Eric

nilmerg:
Done.

PR #3592, opened 2018-10-16
LdapConnection: Add timeout setting with a useful default value

lazyfrosch:


dnsmichi:
Please update the docs too :) 
https://github.com/Icinga/icingaweb2/blob/master/doc/04-Resources.md

PR #3638, opened 2018-11-29
Persistent Collapsible Containers

flourish86:


flourish86:
#34307d3 fixes collapsible container id handling.

#f07e3c0 adds demo output for collapsible containers in detail view.

nilmerg:
So... I've reworked this now quite a bit. The previous implementation has been plagued by inconsistencies regarding the used jQuery functions and made too many assumptions how to make containers collapsible. Having an extra data attribute just for id's didn't seem necessary to me as well.

The general usage of this behavior is now as follows:

```html
<div class="collapsible">
    <!-- This will collapse the entire content to the default height of 36px. The height can be customized with `data-visible-height` -->
</div>

<table class="collapsible">
    <!-- This will collapse `tbody` to the default amount (2) of `tr`s. The amount can be customized with `data-visible-rows` -->
</table>

<ul class="collapsible"> <!-- may also be an ol -->
    <!-- This will collapse to the default amount (2) of `li`s. The amount can be customized with `data-visible-rows` -->
</ul>
```

Note that none of these have an id. `collapsible.js` utilizes internally `ìcinga.utils.getCSSPath` now and automatically identifies an appropriate id in the container's vicinity.

On shutdown (window unload, controlled by `icinga.js` which now *destroys* behaviors just like modules) the recently expanded collapsibles are preserved in the browsers local storage.

Possible issues:
* The behavior assumes now that every id is unique. This will cause conflicts with elements having the same id but different content. (e.g. `#plugin-output`) Though, this may very well be of advantage as well since someone expanding plugin output may want to do this everywhere else as well.
* Protected ids are useless at the moment. A window's id is not stored in local storage thus will definitely change and the container's id will get invalid.
* There's no cleanup of the local storage currently. Containers expanded at shutdown are written into storage and not removed until either sometime nothing is expanded at shutdown (storage is cleared) or they'll get manually collapsed.

modules/monitoring/application/views/scripts/partials/object/detail-content.phtml
```diff
@@ -31,7 +31,7 @@
     </table>
 
     <h2><?= $this->translate('Check execution') ?></h2>
-    <table class="name-value-table">
+    <table class="name-value-table collapsible" data-visible-rows="2" id="<?= $this->protectId('boo') ?>">
```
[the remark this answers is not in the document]
nilmerg: That's an example, of course. This PR is only about the behavior. Where it's utilized then is subject to other issues/PRs.

nilmerg:
I've rebased the branch, are you sure you've got a working copy? For me everything works as expected, Chromium though. The state saving.. no that can't really cope with multiple simultaneously open tabs. I just had multiple sessions in mind. The local storage only allows simple key-value pairs so the behavior utilizes a single value to store its state. Though, adjusting the storage transparently/instantly, while making it less efficient, would also save us the need to *destroy* the behavior as we wouldn't need the unload event.

nilmerg:
Fixed js style issues. Didn't found the css style issues you've mentioned.

Made it work with multiple tabs by directly speaking to the localstorage when updating the state. I'm well aware of the fact that this is highly inefficient, but unless we make use of a dedicated web-worker or such this will not be an easy thing to optimize. If you think this is reasonable, I'll happily do so. Though, since that's another thing that might require a more sophisticated implementation, we should probably wait for it until v2.8.

lippserd:
I still have the style and functionality issues. Safari has no style issues but it does not expand the container.

Screenshot 2019-06-27 at 15 01 01

lippserd:
@flourish86 Could you please test this?

flourish86:
I can't reproduve any of the mentioned issues.

@lippserd: The reason why it might not collapse in your case is, because there aren't sufficient rows.

I also discovered another issue: The collapsible state isn't calculated correctly when you switch from two column to one column layout.

See here (from Safari):

Screen Recording 2019-06-27 at 17 38 24 2019-06-27 17_39_32

nilmerg:
@flourish86 I'd rather say that for only text content a specific height (i.e. 36px) does not work since fonts tend to vary across devices :wink: I guess counting the lines is the better option here :thinking:

flourish86:
> @flourish86 I'd rather say that for only text content a specific height (i.e. 36px) does not work since fonts tend to vary across devices 😉 I guess counting the lines is the better option here 🤔

Sounds realtistic. But shouldn't one line be exactly `font-size * line-height`?

nilmerg:
> Sounds realtistic. But shouldn't one line be exactly `font-size * line-height`?

*Should* be, yes. Though, internet said this is not reliable as not all browsers give a unit-less value. Or even `normal` may be a case. And calculating all cases correctly on-demand I figured this is a nightmare since this may differ based on where text is displayed. (i.e. A dummy text content somewhere hidden can't be used as basis.)

I've committed de2744542c30ff093440da020d97c953c022e981 as a work-around.

public/js/icinga/behavior/collapsible.js
```diff
@@ -0,0 +1,216 @@
...
+        this.defaultVisibleHeight = 36;
+
+        this.state = new Icinga.Storage.StorageAwareMap.withStorage(
+            Icinga.Storage.BehaviorStorage('collapsible'),
+            'expanded'
+        )
+            .on('add', this.onExternalExpansion, this)
+            .on('delete', this.onExternalCollapse, this);
+    };
+    Collapsible.prototype = new Icinga.EventListener();
```
lippserd: I would add a newline here. Else the prototype initialisation is kind of hidden.

public/js/icinga/behavior/collapsible.js
```diff
@@ -0,0 +1,216 @@
...
+        }
+    };
+
+    /**
+     * A collapsible got collapsed in another window, try to apply this here as well
+     *
+     * @param   {Event}     event
+     * @param   {string}    collapsiblePath
+     */
+    Collapsible.prototype.onExternalCollapse = function(collapsiblePath) {
```
lippserd: `onExternalCollapse` sounds like the world is going down 😆 Maybe we should stick to verbs, e.g. `onCollapseExternal` and `onExpandExternal`?

public/js/icinga/behavior/collapsible.js
```diff
@@ -0,0 +1,216 @@
...
+     *
+     * @returns {boolean}
+     */
+    Collapsible.prototype.canCollapse = function($collapsible) {
+        var rowSelector = this.getRowSelector($collapsible);
+        if (!! rowSelector) {
+            return $(rowSelector, $collapsible).length > ($collapsible.data('visibleRows') || this.defaultVisibleRows);
+        } else {
+            var actualHeight = $collapsible[0].scrollHeight,
+                maxHeight = $collapsible.data('visibleHeight') || this.defaultVisibleHeight;
```
lippserd: Please add `var` here.

public/js/icinga/storage.js
```diff
@@ -0,0 +1,498 @@
+/*! Icinga Web 2 | (c) 2019 Icinga GmbH | GPLv2+ */
+
+;(function (Icinga) {
```
lippserd: Please remove the space after the `function` keyword.

public/js/icinga/behavior/collapsible.js
```diff
@@ -0,0 +1,216 @@
...
+
+        if ($collapsible.length && $collapsible.is('.can-collapse')) {
+            this.expand($collapsible);
+        }
+    };
+
+    /**
+     * A collapsible got collapsed in another window, try to apply this here as well
+     *
+     * @param   {Event}     event
```
lippserd: There is no `event` param.

public/js/icinga/behavior/collapsible.js
```diff
@@ -0,0 +1,216 @@
...
+                    _this.collapse($collapsible);
+                }
+            }
+        });
+    };
+
+    /**
+     * A collapsible got expanded in another window, try to apply this here as well
+     *
+     * @param   {Event}     event
```
lippserd: There is no `event` param.

public/js/icinga/storage.js
```diff
@@ -0,0 +1,498 @@
...
+         * @param   {object}    thisArg
+         *
+         * @returns {void}
+         */
+        forEach: function(callback, thisArg) {
+            if (typeof thisArg === 'undefined') {
+                thisArg = this;
+            }
+
+            return this.data.forEach(function(value, key) {
```
lippserd: `forEach()` should return nothing.

PR #3652, opened 2018-12-18
Auth: do not ask for unrelated group membership

Thomas-Gelf:
If a specific User-Backend has been assigned to a Group Backend, and
the User has been authenticated by another User-Backend, then there is
no need to ask the unrelated Groups Backend for membership.

### DIFF
--- a/library/Icinga/Authentication/Auth.php
+++ b/library/Icinga/Authentication/Auth.php
@@ -139,7 +139,20 @@ class Auth
         $this->getResponse()->setReloadCss(true);
         $user->setPreferences($preferences);
         $groups = $user->getGroups();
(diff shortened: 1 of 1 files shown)


lippserd:
Hi Tom,

PR looks good. Just one question: Where does `none` come from?

Best,
Eric

Thomas-Gelf:
@lippserd: it's in the [form](https://github.com/Icinga/icingaweb2/blob/master/application/forms/Config/UserGroup/LdapUserGroupBackendForm.php#L67) and being used [here](https://github.com/Icinga/icingaweb2/blob/master/library/Icinga/Authentication/UserGroup/LdapUserGroupBackend.php#L803). Don't know why this isn't `null`.

Cheers,
Thomas

PR #3666, opened 2019-01-17
Feature/check all services for host

SaschaDoering:
This PR adds another quick-action for hosts to check all services at once. We also added a German translation. While we were doing this, the following problem arose. The command `icingacli translation refresh module monitoring de_DE` didn't find the used translation, we believe it's because we translate a variable instead of a string.

Please let us know if everything is fine with the PR, we're open for feedback.


ekeih:
Anything we can do to get this merged?

nilmerg:
Hi, thanks for the PR, I'll have a look at it now.

> we believe it's because we translate a variable instead of a string.

Correct :+1: 

> Please let us know if everything is fine with the PR, we're open for feedback.

A proper review follows, but at first glance, the merge conflict needs attention. :wink:

nilmerg:
At second glance, were you aware of the case that the same is already possible by accessing the *Services* tab and selecting all services?

I just wonder what the advantage of this solution is. Because **if** I'll accept this I'd want to have this quick action as part of the *Services* tab anyway. All other quick actions in the *Host* tab are only meant for this single host. This new action doesn't exactly suit there.

Then it's only a difference of pushing a button and two additional clicks. Sure, still way more than a single click but all quick actions involving multiple hosts/services are currently only available on our multi-selection views. I guess I just want another justification than the shorter navigation alone. :thinking:

ekeih:
@nilmerg The "only" advantage is the shorter navigation path.

I think that the quick action makes more sense in the Hosts tab because at this point the user does not really care about the services itself. She or he just wants to recheck everything on the host with as few clicks as possible.
Anyway, if you still want the quick action only in the Services tab we could move it there. (But it would require one click more than in the Host tab.)

While it may sound like a small change it is a regular requested feature from my coworkers. So we will definitively add this feature to our own IcingaWeb2 instances, we just thought that it would be nice to have this in the upstream code instead of our local patch.

ekeih:
A few other points that came up in an offline discussion:
* We could also add the quick action to the Host tab AND the Services tab.
* The quick action is helpful when you many services and a small display.
* If you select many services the URL gets to long and the webserver refuses the request. The exact limit depends on the webserver. With the quick action this does not happen.

dnsmichi:
I kind of like the feature, though some thoughts about it:

- `check now` is a forced re-check, async to the specified check interval. If users click that often, followed by a web page refresh, this may cause master or satellite check schedulers to overload. With badly written plugins, this may cause load. And it can of course trigger notifications you'd not expect at this point.
- Having `check now` with a single click for 1..n services invites users to click it, while with the selection and multiple action it becomes harder to even do it. What exactly is the use case to immediately check all the services ... a server reboot, ended downtime maintenance window, anything else?
- Not sure if the host tab really is the right context, the first line of quick actions should solely catch the host actions. The services tab currently doesn't have actions, but may a candidate to a) re-check all b) schedule downtime for all. Still my objection against multiple forced re-checks from above still stands.

Cheers,
Michael

modules/monitoring/application/forms/Command/Object/CheckNowCommandForm.php
```diff
@@ -21,6 +21,26 @@ public function init()
         $this->setAttrib('class', 'inline');
     }
 
+    /**
+     * SubmitButton Label
+     *
+     * @var SubmitButtonLabel
+     */
+    protected $submitButtonLabel = 'Check now';
```
nilmerg: This must be somehow marked as translatable. Consider doing so in method `init()`.

modules/monitoring/application/forms/Command/Object/CheckNowCommandForm.php
```diff
@@ -21,6 +21,26 @@ public function init()
...
+
+    /**
+     * Set the submit button label
+     *
+     * @param   String $submitButtonLabel
+     *
+     * @return  $this
+     */
+    public function setSubmitButtonLabel(String $submitButtonLabel)
```
nilmerg: The type hint for strings does only work for PHP version 7+. Since we still support 5.6+ this must be removed.

modules/monitoring/application/forms/Command/Object/CheckNowCommandForm.php
```diff
@@ -21,6 +21,26 @@ public function init()
         $this->setAttrib('class', 'inline');
     }
 
+    /**
+     * SubmitButton Label
+     *
+     * @var SubmitButtonLabel
```
nilmerg: `@var` takes a datatype name. (i.e. `@var string`)

nilmerg:
Hi,

so, I'll close this now. No update happened over the last year, Icinga DB has already appeared on the horizon (finally) and I still have my reservations regarding the place of this button.

Icinga DB Web already provides a new small widget at a host's details which shows the states and count of services of a host. Clicking on the amount shows the services in a list and with the new *Continue With* widget it's also no more than a single click to reach the multi-selection view. Yep, it's more than a single click, but since it doesn't involve any *shift-click* acrobatics anymore this has definitely improved.

Thanks for your understanding.

PR #3670, opened 2019-01-25
FilterExpression#matches(): save a few CPU cycles

Al2Klimov:


library/Icinga/Data/Filter/FilterExpression.php
```diff
@@ -163,6 +163,10 @@ protected function strtolowerRecursive($var)
 
     public function matches($row)
     {
+        if (! isset($row->{$this->column})) {
```
nilmerg: This serves obviously two purposes. Checking that the property exists and that it's not `null`. The first is already verified by the try-catch. The second case is probably better verified after the property has already been accessed. This may be a dynamic property which this way may be resolved twice. Please just check for `null` below.

Al2Klimov: Oh yes... almost forgot PHP's omnipresent magic. Isn't this PR anyway a breaking change then?

nilmerg: No, not that I'm aware of. The tests are also not complaining. Though, as the tests have been mentioned now, please create one for this change :wink:

Al2Klimov: I mean: Now we would check for NULL – that would change the behavior.

Al2Klimov:
This breaking change is not worth to "save a few CPU cycles".

POC:

```php
Filter::where('x', [null])->matches(null)
```

PR #3672, opened 2019-01-26
Use https:// links on login page

andreaso:
These url changes are all the result of following 301 redirects.


lippserd:
Thank you 👍

Issue #3416, opened 2018-04-11
Improve UI


lippserd:
- [x] Improve the looks of our forms.
- [x] Improve backend configuration dialogs, especially the role management.
- [x] Allow to collapse custom variables and performance data.
- [x] Add a date and time picker control.
- [ ] Allow to collapse and expand columns.
- [ ] Provide list inputs with paging capability.
- [ ] Provide autocomplete dropdowns.
- [x] Evaluate how to improve the service list views to quickly identify which services belong to which host(s).
- [ ] Evaluate whether we should support resizable columns and implement this if we want it.
- [ ] Collapse too wide tabs into an expandable "..." menu.
- [ ] Improve fonts, colors and contrasts.
- [ ] Move away from fontello.
- [ ] Provide more differentiation between rows in list views.
- [ ] Expand dashboard columns to fill the remaining space.
- [ ] **Mobile layout:** Move menu icon into the tab bar.
- [ ] **Mobile layout:** Tooltips are cut off at screen edge.

jmroth:
I've got another one: No progress shown on click

Unless you're running the web developer tools of the browser you don't know what's going on.

Click for example on "Event Overview" in the sidebar. No feedback. When the DB is slow or anything else happens, you don't know whether the screen you wanted to go to is loading or if you should click again, potentially making the wait even longer since it was already loading. This is especially true if you don't know what the screen you should be getting looks like in the first place.

sinyb:
I don't know if this is the right place, but let me make a suggestion: reduce the height of list items. There is no need for them to take up 3 lines of text (in Icinga Classic they used to fit just fine within one line). 
Make this an option,  maybe? 
I am monitoring 100's of hosts and services and I hate to have to scroll the page to see more than 12, and never more than 12 at a time, when there is space for at least 3 times that number. Also, that way there won't be half of the screen empty...

karlisup:
Hi @lippserd ,
Could you tell me just briefly what is the future vision of this task:
- Move away from fontello.
Are you thinking to move to SVG icons or to use another tool that would manage font-icons?

PR #3676, opened 2019-01-31
Improve look of our forms

### LINKED ISSUE #3560: Announcements should be more visible
The announcements look like just another bookmark bar in the Browser. Since Announcements are more less important information for the user they should be more visible. 

## Possible Solution
- Colorize
- Maybe something like that: https://css-tricks.com/examples/PopFromTopMessage/

1. Create an announcement
2. Ask your colleagues if they have noticed new the annoucment ;)

theFeu:
refs #3416 
fixes #3560

### DIFF
--- a/public/css/icinga/forms.less
+++ b/public/css/icinga/forms.less
@@ -1,5 +1,14 @@
 /*! Icinga Web 2 | (c) 2019 Icinga Development Team | GPLv2+ */
 
+/**
+ Rules found in here are structured with two layers:
+
+  1) form.icinga-forms, that's what defines the general structure of our single/individual forms. It's not
+     supposed to be used for any other forms that are not the only content on the page (e.g. inline-forms)
+  2) .icinga-controls, this defines the design of our controls. Any input that's part of a container with
+     this class gets our design applied
+ */
+
 // Used colors
 
 @disabled-gray: #9a9a9a;
@@ -10,26 +19,25 @@
 
 // General form layout
 
-form:not(.inline):not([role="search"]) {
+form.icinga-forms {
   max-width: 70em;
   width: 80%;
-}
 
-.form-controls {
-  display: flex;
-  margin-left: 16em;
-  justify-content: flex-end;
-}
+  .control-group {
+    display: flex;
+    flex-wrap: wrap;
+    margin: 0 -.5em;
+    align-items: flex-start;
 
+    > * {
+      margin: .5em;
+    }
+  }
 
-
-.control-group {
-  display: flex;
-  margin: 0 -.5em;
-  align-items: flex-start;
-
-  > * {
-    margin: .5em;
+  .form-controls {
+    display: flex;
+    margin-left: 16em;
+    justify-content: flex-end;
   }
 }
 
@@ -37,7 +45,7 @@ form:not(.inline):not([role="search"]) {
 
 #layout.minimal-layout,
 #layout.twocols:not(.wide-layout) {
-  form:not(.inline) {
+  form.icinga-forms {
     width: 100%;
 
     .control-label-group {
@@ -47,24 +55,24 @@ form:not(.inline):not([role="search"]) {
       padding-left: 0;
       margin-bottom: 0;
     }
-  }
 
-  .toggle-switch ~ .control-info:before {
-    margin-left: 0;
-  }
+    .toggle-switch ~ .control-info:before {
+      margin-left: 0;
+    }
 
-  .control-info {
-    opacity: .5;
-  }
+    .control-info {
+      opacity: .5;
+    }
 
-  .errors > li {
-    margin-left: .5em;
+    .errors > li {
+      margin-left: .5em;
+    }
   }
 }
 
 // Label styles
 
-.control-group .control-label-group {
+form.icinga-forms .control-group .control-label-group {
   display: flex;
   flex-direction: column;
   justify-content: center;
@@ -75,7 +83,7 @@ form:not(.inline):not([role="search"]) {
   width: @name-value-table-name-width;
 }
 
-.control-info {
+form.icinga-forms .control-group .control-info {
   line-height: 2.25em;
   margin-right: 1em;
   opacity: .6;
@@ -86,18 +94,9 @@ form:not(.inline):not([role="search"]) {
   }
 }
 
-.toggle-switch ~ .control-info {
+form.icinga-forms .control-group .toggle-switch ~ .control-info {
   position: relative;
   margin-left: .5em;
-}
-
-.spinner {
-  line-height: 2.5em;
-
-  &:before {
-    margin-left: .5em;
-    opacity: 0.4;
-  }
 }
 
 form.inline {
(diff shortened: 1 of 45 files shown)


theFeu:
What's still to do here is to go through our current forms and check if the new styles apply properly.

Add classes to the inputs, so we don't have to use the tags in the CSS and then adapt that.

public/css/icinga/forms.less
```diff
@@ -1,618 +1,231 @@
...
+/**********************************************************************************************************************
+                                             specific input styles
+***********************************************************************************************************************/
 
-  &:focus {
-    background-color: @body-bg-color;
-    background-image: url('../img/icons/search_icinga_blue.png') !important;
-  }
+.control-group textarea {
+  background-image: url(../static/img?file=textarea-corner-2x.png&module_name=reporting);
```
lippserd: The URL is wrong.

public/css/icinga/forms.less
```diff
@@ -1,618 +1,231 @@
...
-  margin-left: 0.5em;
+.form-controls input:not(:last-child),
+.form-controls button:not(:last-child) {
+  margin-right: 1em;
 }
 
-button:hover .icon-cancel {
-  color: @color-critical;
+.control-group select {
+  background-image: url(../static/img?file=select-icon-2x.png&module_name=reporting);
```
lippserd: The URL is wrong.

public/css/icinga/mixins.less
```diff
@@ -6,26 +6,34 @@
           box-shadow: @arguments;
 }
 
-.button(@background-color: @body-bg-color, @color: @icinga-blue) {
+.button(@border-font-color: @icinga-blue, @color-dark: @icinga-blue-dark, @background-color: @body-bg-color) {
```
lippserd: Please don't change this mixin because we're not the only ones using it.

public/css/icinga/forms.less
theFeu: fixed in c2b886598579c3293727954afff8f8e383cdd73e

public/css/icinga/forms.less
theFeu: fixed in c2b886598579c3293727954afff8f8e383cdd73e

theFeu:
> **Current Forms:**
> * Think about auto-submit spinner
> * Think about control-info

@lippserd 
We might need to discuss this before I change something up :)

public/css/icinga/forms.less
```diff
@@ -78,6 +78,10 @@ form:not(.inline) {
   //todo(jem): maybe solve width with flexbox flex-growth: 1
 }
 
+.control-info {
+  width: 0;
```
theFeu: @flourish86 is this too dirty?

flourish86:
Here are some details to fix:

- [ ] Please add some opacity to the refresh indicators (~.4)
- [ ] Use the default cursor, when hovering disabled checkboxes
- [ ] Add negative margin to asterisks, if necessary wrap label text in another `<span>` to make sure asterisks won't get ellipsisized. This alignes the text borders
- [ ] Make sure to inset buttons by the label width to align with the inputs

theFeu:
Maybe rethink the checkbox styles again and potentially solve the position with a 1 px margin or a transparent border for the :before

theFeu:
Spinner width 0 makes them unusable in the module activation in safari
-> find different solution

theFeu:
Replace the ellipsis for the form labels with a 2 line high box that just cuts off if the label is too long.

theFeu:
Hover on spinner doesn't center the tooltip correctly

theFeu:
input type number still has browser style

theFeu:
should the select arrows be icingablue?

theFeu:
form description has no styling yet - how do we want to display this?

theFeu:
how do we display form validation / info messages?

flourish86:
For Checkbox `:focus` state the outline isn't rounded. Please remove the outline for checkboxes only and use `box-shadow` instead.
- [ ] Remove outline for 'input[type="checkbox"]:focus` and replicate effect using `box-shadow`

theFeu:
- [ ]  make checkboxes IE / Edge compatible

theFeu:
So, I just found an old pen I made where I attempted to improve the handling of the role form
https://codepen.io/the_Feu/pen/zEeLdv

If you could have a look at it and tell me what you think @flourish86 that'd be neat!

theFeu:
Where should the tooltips be for checkboxes?

Right next to them?
image

or

In a col with the others?
image

theFeu:
- [ ] form description has no styling yet - how do we want to display this?

My first idea was to have rows with light gray background looking like this:
desktop:
image

mobile:
image

theFeu:
- [ ] Should the select arrows be icinga blue?
Since it is an interaction and we want to let the user know this is clickable I would say: yes?

flourish86:
> [x] Should the select arrows be icinga blue?
> Since it is an interaction and we want to let the user know this is clickable I would say: yes?

Yes, pls leave them as they are.

theFeu:
- [ ] Add style for disabled checkboxes

theFeu:
- [ ] style form errors

put the error message under the input field

theFeu:
Still has no fluent layout switch - when window is pulled smaller there is a period of awkwardness...

image

flourish86:
A side note: line-height is somehow odd on `select` elements, especially cross-browser wise.

I'm trying to fix this at the moment, with a reasonable amount of style code.

flourish86:
I worked it all through and it seems like it's pretty done now.

There are still tiny imperfections with the filter editor. Nothing critical, though, so I think the current state is reasonably well in general and appropriate for merge. We should have the filter editor overworked in a separate issue.

A Review would be nice @lippserd.

lippserd:
@flourish86 I had a look at the changes and noticed the following:
* Form controls have a submit animation
* The search bar in the menu has not the same height as the other menu entries. Though that's not new I guess. Should we change that? They have the same height when the sidebar is collapsed. Should we show the "search clear" button as well when the sidebar is collapsed?
* The filter editor has some style issues. See screenshot [1]
* The cancel button for the bottom notifications is kind of lost. See screenshot [2]
* The same goes for announcements. See screenshot [3]

[1]
Screenshot 2019-07-03 at 10 38 47

[2]
Screenshot 2019-07-04 at 11 40 32

[3]
Screenshot 2019-07-04 at 13 13 01

flourish86:
Can't help, but having the nav search input being as tall as the nav items seems somehow odd.

Screenshot 2019-07-04 at 17 46 11

flourish86:
> Form controls have a submit animation

Should be fixed by 33fbe30

> The search bar in the menu has not the same height as the other menu entries. Though that's not new I guess. Should we change that? They have the same height when the sidebar is collapsed. Should we show the "search clear" button as well when the sidebar is collapsed?

As pointed out above, I wouldn't change as it looks more balanced now.

> The filter editor has some style issues. See screenshot [1]

Should be fixed by 0cd54c8

> The cancel button for the bottom notifications is kind of lost. See screenshot [2]

Fixed by 620ae0a

> The same goes for announcements. See screenshot [3]

Fixed by 456c48f

flourish86:
@nilmerg Still to do:

1) The `<em>`s wrapping the highlighted elements in the labels are still missing.
2) The permission and restriction subheading is contained in a `.control-group`, which also contains a `.control-label-group`. I couldn't find a way to make the headings left-aligned this way. Can you please remove the `.control-label-group`?

lippserd:
@flourish86 Could please fix the following issues:

* The search input height is still too low in my opinion
* The bottom notifications look displaced

localhost_8080_icingaweb2_dashboard

lippserd:
Plus, if the side bar is collapsed and you hover over the search icon, the changed background does not fill the full tile height.

flourish86:
@lippserd: fixed by 4b88dd8

lippserd:
Thx! Any chance to fix the reset button flickering upon auto-refresh of the menu?

flourish86:
> Thx! Any chance to fix the reset button flickering upon auto-refresh of the menu?

Commit 9b5b604 removes fade effect. It also hides the button, when the input is empty (not `:valid`)

lippserd:
@flourish86 Did you fix the bottom notifications?

flourish86:
Now ccd4d4d

public/css/icinga/widgets.less
```diff
@@ -66,6 +68,10 @@
 
   position: relative;
 
+  p {
+    margin-bottom: 0;
```
nilmerg: @flourish86 There's no bottom margin even without this. Why is this necessary?

nilmerg: Never mind, I bet you've fixed the bottom notifications without knowing I've done this already with ef7bdf807ff75533e91c82ecf47758809bb12e3b

lippserd: I guess that's true 😆

lippserd:
I can't expand the modules in the role form.

nilmerg:
Should be fixed now :sweat:

lippserd:
Looks good!

sinyb:
With a little help from Firefox debugging, I have changed a few files (relative to icingaweb2):
modules/monitoring/application/views/scripts/list/hosts.phtml
modules/monitoring/application/views/scripts/list/services.phtml
modules/monitoring/public/css/module.less
modules/monitoring/public/css/tables.less
public/css/icinga/base.less
and managed to squeeze much more useful info on a single screen, as can be seen down in screenshots. I needed this because out of 2500 hosts, there are always at least 20 with problems, and I want to see them all at once.
Now, I wonder if it could be possible to pack this into a Theme, so I don't have to change everything again after an update?
I suppose that I can override .less files via public/css/themes  (will try that next time I have some spare time), but what about .phtml? I needed to delete some div's, and reorder a bit of php/html code (I can attach diffs if anybody is interested). This is what I was writing about in issue #3416  ...

Before
Screenshot_20200127_213203

After (not on the same system):
Screenshot_20200127_213208

nilmerg:
@sinyb Please head over to our community forums https://community.icinga.com where you'll certainly get more help sooner.

PR #3679, opened 2019-02-06
 monitoring/hostgroup: Add service badges to grid view

lazyfrosch:
This adds more detail information to the hostgroup grid.

bildschirmfoto von 2019-02-06 11-52-29

ref/NC/590161
ref/NC/575665


lippserd:
The host group grid has been introduced in order to show tons of groups. This PR kind of works against this because it reduces the available space and disturbs the UI. I'll check this with @flourish86 and @nilmerg.

lazyfrosch:
Actual grid size is more or less the same, it just contains more information.

nilmerg:
It's not only about the amount of space that is available to groups but also about this view's accessibility. Note how cluttered the view is now, despite the fact that it requires nearly half the previous space more. The hostgroup grid was meant only for the groups themselves, throwing in the service states does not quite match this purpose.

default-view
new-view

flourish86:
This seems pretty odd to me.

Firstly, services are already listed in the *Service Grid* list view. So if this is the information you're looking for, why not use this view.

Secondly the addition of the services state not only wastes space, but also makes the view desperately messy. On one hand the layout itself gets unnecessarily noisy. On the other hand the additional coloured badges distract from the primary state badges of the host group.

The service badges in the grid view won't add information, that isn't already there. But it makes it harder to grasp the information, for which the grid view is intended.

lazyfrosch:
Let's close this, it will be moved to a customer specific module...

PR #3680, opened 2019-02-06
Setup: don't require date.timezone

### LINKED ISSUE #3626: It is required that a default timezone has been set using date.timezone
## Expected Behavior

I don't have to set date.timezone as [Icinga Web 2 falls back anyway](https://github.com/Icinga/icingaweb2/blob/9f15bda698952a369a2bcf949842548a60e447ac/library/Icinga/Application/ApplicationBootstrap.php#L624) in the rare case that my browser doesn't provide any TZ offset.

## Current Behavior

I have to configure it every damn time.

## Possible Solution

Replace the error in the setup wizard with a warning.

Run the setup wizard as usual.

## Context

Testing something with a fresh IW2.

## Your Environment

The latest bleeding edge hipster stuff from you.

Al2Klimov:
fixes #3626

### DIFF
--- a/modules/setup/library/Setup/WebWizard.php
+++ b/modules/setup/library/Setup/WebWizard.php
@@ -575,15 +575,6 @@ class WebWizard extends Wizard implements SetupWizard
             )
         )));
 
-        $set->add(new PhpConfigRequirement(array(
-            'condition'     => array('date.timezone', true),
-            'title'         => mt('setup', 'Default Timezone'),
-            'description'   => sprintf(
-                mt('setup', 'It is required that a default timezone has been set using date.timezone in %s.'),
-                php_ini_loaded_file() ?: 'php.ini'
-            ),
-        )));
-
         $set->add(new OSRequirement(array(
             'optional'      => true,
             'condition'     => 'linux',
--- a/doc/02-Installation.md
+++ b/doc/02-Installation.md
@@ -16,7 +16,6 @@ chapter.
 * A web server, e.g. Apache or Nginx
 * PHP version >= 5.6.0
 * The following PHP modules must be installed: cURL, gettext, intl, mbstring, OpenSSL and xml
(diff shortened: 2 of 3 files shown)


modules/setup/library/Setup/WebWizard.php
```diff
@@ -576,6 +576,7 @@ public function getRequirements($skipModules = false)
         )));
 
         $set->add(new PhpConfigRequirement(array(
+            'optional'      => true,
```
nilmerg: Please only do this for PHP versions lower than 7.0. v5.6 is still affected.

nilmerg:
And please also update the requirement in the installation documentation.

modules/setup/library/Setup/WebWizard.php
Al2Klimov: I've fixed this in #3747. POC module:

```
[root@07b1f18d5654 /]# cat icingaweb2/modules/badtz/application/clicommands/ProofCommand.php
<?php

namespace Icinga\Module\Badtz\Clicommands;

use Icinga\Cli\Command;

class ProofCommand extends Command
{
    public function fallbackAction()
    {
        var_dump(@date_default_timezone_get());
        var_dump(date_default_timezone_get());
    }
}
[root@07b1f18d5654 /]# cat icingaweb2/modules/badtz/application/controllers/ProofController.php
<?php

namespace Icinga\Module\Badtz\Controllers;

use Icinga\Web\Controller;

class ProofController extends Controller
{
    protected $requiresAuthentication = false;

    public function fallbackAction()
    {
        var_dump(@date_default_timezone_get());
        var_dump(date_default_timezone_get());
    }
}
[root@07b1f18d5654 /]# cat icingaweb2/modules/badtz/application/views/scripts/proof/fallback.phtml
[root@07b1f18d5654 /]#
```

Issue #3527, opened 2018-07-16
Vagrant: Initial provision fails with PDO class not found, needs PHP-FPM restart


dnsmichi:
## Expected Behavior

http://localhost:8080/icingaweb2 works

## Current Behavior

The PDO class cannot be loaded.

```
Fatal error: Class 'PDO' not found in /vagrant/library/Icinga/Data/ResourceFactory.php on line 103
```

## Possible Solution

Restart PHP-FPM at the end, this fixes the problem.

```
systemctl restart rh-php71-php-fpm.service
```

I'm doing the same for icinga-vagrant already.
https://github.com/Icinga/icinga-vagrant/blob/master/.puppet/scripts/shell_provisioner_post.sh#L5

## Steps to Reproduce (for bugs)

1. vagrant up
2. http://localhost:8080/icingaweb2

## Your Environment

* Icinga Web 2 version and modules (System - About): HEAD, 84e0c0c4fbd1ce82c2b9926bb9ff7e04d3aefd54

PR #3686, opened 2019-02-13
Fix missing PHP-FPM restart after Puppet provisioning finished

mrimann:
refs #3527

### DIFF
--- a/.puppet/modules/php/manifests/init.pp
+++ b/.puppet/modules/php/manifests/init.pp
@@ -30,6 +30,7 @@ class php {
   -> service { 'rh-php71-php-fpm':
     ensure => running,
     enable => true,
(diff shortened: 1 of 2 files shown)


lippserd:
Hi,

Thanks for the PR. Though I think that we should fix that in the Puppet manifests. Could you try setting an alias for the `rh-php71-php-fpm` service in `.puppet/modules/php/manifests/init.pp` and notify that alias in `.puppet/modules/php/manifests/extension.pp` instead of the apache service? `php-fpm` as alias sounds good to me.

Best,
Eric

mrimann:
Hi Eric
I hope I understood your proposal the right way - please review. I couldn't 100% test the change - a second test would help, too.
Best,
Mario

Issue #3690, opened 2019-02-20
Use auto-detected binary prefix (SI/IEC) for perf data display


lippserd:
We always use the IEC prefixes for perf data display. This is wrong for disk checks for example. But we could automatically detect whether to use SI or IEC prefixes from the perf data units.

lippserd:
Stalled until Icinga 2 supports writing UOMs with the IEC prefix.

nilmerg:
Superseded by https://github.com/Icinga/icingadb-web/issues/71

PR #3695, opened 2019-02-21
Detect binary prefix and use for perf data display

theFeu:
refs #3690

### DIFF
--- a/modules/monitoring/library/Monitoring/Plugin/Perfdata.php
+++ b/modules/monitoring/library/Monitoring/Plugin/Perfdata.php
@@ -163,7 +163,27 @@ class Perfdata
      */
     public function isBytes()
     {
-        return in_array($this->unit, array('b', 'kb', 'mb', 'gb', 'tb'));
+        return $this->isSi() || $this->isIec();
+    }
+
+    /**
+     * Return whether this performance data's value is in a SI unit
+     *
+     * @return  bool    True in case it's in a SI unit, otherwise False
+     */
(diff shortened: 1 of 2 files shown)


nilmerg:
What's been the rationale for this? Most checks I know of return SI units. Are there any returning IEC ones? If not, does a fix to correct the transformation suffice?

lippserd:
Stalled until Icinga 2 supports writing UOMs with the IEC prefix.

nilmerg:
That's https://github.com/Icinga/icinga2/issues/7225, for reference.

nilmerg:
@lippserd Just noticed.. does it make sense finishing this for the monitoring module? Yes, Icinga DB still depends on monitoring and also utilizes this class but at the time Icinga 2.13 is a thing aren't we also ready/finished having this natively in Icinga DB?

lippserd:
> @lippserd Just noticed.. does it make sense finishing this for the monitoring module? Yes, Icinga DB still depends on monitoring and also utilizes this class but at the time Icinga 2.13 is a thing aren't we also ready/finished having this natively in Icinga DB?

We don't have this automatically in Icinga DB. The performance data is stored unparsed still.
We should implement this together with icinga/icinga2#7225. It is ok though to have this in Icinga DB Web only.

nilmerg:
Then this gets closed.

lippserd:
> Then this gets closed.

Do we have a new issue for Icinga DB Web?

PR #3699, opened 2019-02-22
[WIP] Modal

theFeu:


theFeu:
What is still to do here?

- clean up and document class names and structures
- more testing with variable content

PR #3703, opened 2019-02-27
host downtimes: set the all_services flag default to true

pogii123:
I realized, that in our company nobody wants to set a downtime only for a host. We set only downtime for hosts with all services.

So I think other company's and Icingaweb2 users have the same problem. We have a lot of admins which forget to set this flag and then the monitoring notifies falsely.

Since we have changed this flag locally, we save a lot of time. I think it would be helpful for other users too, so I thought it would be nice to share this with others.


dnsmichi:
You can change the settings in your config.ini too, no need to patch the source code.

pogii123:
Hi @dnsmichi,
which parameter do I have to set for this?

I searched in the Icinga Web 2 documentation under the "Configuration" and "Advanced Topics" sections and I cant find any description which include this type of changes.

And also not in the source code.

Could you help me with that?

Regards,
pogii123

pogii123:
Ok, I'm sorry. I found it.

Its in `/etc/icingaweb2/modules/monitoring/config.ini`

I added:
```ini
[settings]
hostdowntime_all_services = true
```

Thank you again :)

PR #3706, opened 2019-03-04
Fix multiline values in ini files broken

### LINKED ISSUE #3705: Multiline values in ini files broken
Since 86ae8c12d multiline values in ini files cause the UI to badly crash.

There's at least one native way (announcements) with which to get newlines into ini files using the UI.

nilmerg:
fixes #3705

### DIFF
--- a/library/Icinga/File/Ini/Dom/Directive.php
+++ b/library/Icinga/File/Ini/Dom/Directive.php
@@ -157,11 +157,10 @@ class Directive
     {
         $str = trim($str);
         $str = str_replace('\\', '\\\\', $str);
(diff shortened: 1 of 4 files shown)


nilmerg:
I'm not a huge fan of this solution, but it's our only chance to solve the referenced bug:
* It's PHP itself having issues so we can't simply postprocess it's output
* `INI_SCANNER_TYPED` has the same issues as `INI_SCANNER_NORMAL`

We have two ways to natively get multiline values into INI files (announcements, roles) so we have to choose between disabling these or fix the issue.

I've chosen the latter.

Well, there's another option: Not using PHP to parse INI files. But that's something for far more severe issues...

PR #3711, opened 2019-03-07
Don't hardcode limits overriding user preferences

sebastic:
The 'Default page size' preference configured by users is not respected everywhere.

Especially on the 'Current Incidents' dashboard it is very useful to display more result than the hardcoded 10.

### DIFF
--- a/modules/monitoring/configuration.php
+++ b/modules/monitoring/configuration.php
@@ -104,10 +104,10 @@ $this->provideSetupWizard('Icinga\Module\Monitoring\MonitoringWizard');
 /*
  * Available Search Urls
  */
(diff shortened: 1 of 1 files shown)


nilmerg:
I'm afraid but this can't be done the proposed way. By doing so it's showing 25 (instead of 10) items per dashlet if the user does not configure a limit on his/her own. It's also not possible to dynamically add a limit of 10 just for these dashlets, since the dashboard has no notion about the type of content that is displayed.

I'd suggest you duplicate this dashboard in your environment (with the limit removed from the urls) and hide the default one. This way you get the desired result as well.

Issue #3687, opened 2019-02-18
Show 'Check Source' in Service History


ConstantinMatheis:
The 'Check Source' information of a Check Command should also be shown when looking at entries in the history. Currently, it is only shown for the most recent execution of the check.
![not-enough-info](https://user-images.githubusercontent.com/26505965/52954369-4e14cc00-338a-11e9-879a-f2cd98adc39e.png)

## Context
In a clustered environment it is helpful to know the which server issued the command.

## Your Environment
![environment](https://user-images.githubusercontent.com/26505965/52954359-4523fa80-338a-11e9-8cdd-60d9da61530a.png)

PR #3712, opened 2019-03-07
Add check source to history details view.

ConstantinMatheis:
refs #3687


dnsmichi:
That's the event history view, not the detail view, as the topic suggests, right?

nilmerg:
The detail view of an history entry, actually.

PR #3714, opened 2019-03-08
use the correct LDAPS port

### LINKED ISSUE #3713: LDAPS authentication ignores custom port setting
## Expected Behavior
LDAP resource with LDAPS encryption uses the port that is configured in the 'Port' field.

## Current Behavior
Because of how ldap_connect() works, the port used is always 636 (and the configured port is ignored). See "elsint at yahoo dot com" comment on http://php.net/manual/en/function.ldap-bind.php

## Possible Solution
When LDAPS is used, the $hostname must be concatenated with $port. The line
https://github.com/Icinga/icingaweb2/blob/d00fdf4d19839c916d811bc5caeee63f1299f2ec/library/Icinga/Protocol/Ldap/LdapConnection.php#L1195
should be changed to 
`$hostname = 'ldaps://' . $hostname . ':' . $this->port;`

1. in the icingaweb2 GUI, create a new LDAP resource, choose LDAPS and random server (9.9.9.9) and radom port (8888) and random pw as connection parameters
2. on the icingaweb2 instance start tcpdump with `tcpdump -i any host 9.9.9.9 -nn`
3. click "Validate configuration" and watch how tcpdump logs this (notice the wrong port)
```
16:52.100727 IP 10.xx..xx.xx.38266 > 9.9.9.9.636: Flags [S], seq 2868288110, win 26883, options [mss 8961,sackOK,TS val 1034756882 ecr 0,nop,wscale 7], length 0
```
although this is (wrongly) logged by icingaweb2:
```
Mar 08 18:16:57 host icingaweb2[28849]: LDAP bind (uid=xxx / ***) to 9.9.9.9 with default port 8888 failed: Can't contact LDAP server
```

After the code change the connection looks and works like expected:
```
18:29:48.853087 IP 10.xx.xx.xx.60032 > 9.9.9.9.8888: Flags [S], seq 4193613384, win 26883, options [mss 8961,sackOK,TS val 1035533634 ecr 0,nop,wscale 7], length 0
```

## Context
This ticket is similar but it seems it did not fix the problem: https://github.com/Icinga/icingaweb2/issues/2812

## Your Environment
Icinga Web 2 Version     2.6.2
Git commit     63cb9d728338225833884fda2c312a5a9f431fd6
Git commit date    2018-11-21
* Operating System and version: Centos 7
* PHP:
[root@host ~]# /opt/rh/rh-php71/root/usr/bin/php --version
PHP 7.1.8 (cli) (built: Aug  9 2017 13:20:06) ( NTS )

ignasr:
fixes #3713

### DIFF
--- a/library/Icinga/Protocol/Ldap/LdapConnection.php
+++ b/library/Icinga/Protocol/Ldap/LdapConnection.php
@@ -1192,7 +1192,7 @@ class LdapConnection implements Selectable, Inspectable
 
                 $hostname = implode(' ', $ldapUrls);
             } else {
-                $hostname = 'ldaps://' . $hostname;
+                $hostname = 'ldaps://' . $hostname . ':' . $this->port;
             }
         }
 
--- a/library/Icinga/Protocol/Ldap/LdapConnection.php
+++ b/library/Icinga/Protocol/Ldap/LdapConnection.php
@@ -1185,6 +1185,9 @@ class LdapConnection implements Selectable, Inspectable
             $ldapUrls = explode(' ', $hostname);
             if (count($ldapUrls) > 1) {
                 foreach ($ldapUrls as & $uri) {
+                    if (preg_match('/:\d+$/', $uri) === 0) {
+                        $uri = $uri . ':' . $this->port;
+                    }
                     if (strpos($uri, '://') === false) {
                         $uri = 'ldaps://' . $uri;
                     }


ignasr:
Thanks for your comment @nilmerg 
I think that modification of the multiple hostnames block above is not needed. 

https://github.com/Icinga/icingaweb2/blob/9b1f49413db9c4a4772599399576db564778eed0/library/Icinga/Protocol/Ldap/LdapConnection.php#L1186

You see, if one uses multiple LDAPS server hostnames, for example

`host1:3333 host2`

 the string is exploded on space, "ldaps://" is added and a hostname string is created:

`ldaps://host1:3333 ldaps://host2`

which is a valid string for `ldap_connect()`. 

The Port text field is not used and has no meaning when multiple hostnames are entered in the Hostname field.

nilmerg:
And what about this?:

```ini
[abcd]
host = "host1 host2"
port = 1234
ssl = "1"
```

That's transformed to `ldaps://host1 ldaps://host2` which then still causes `ldap_connect` to connect to port 636. Not 1234.

ignasr:
True. Pull request updated.

Issue #3722, opened 2019-03-20
ctrl-click does not open links in new tab


theFeu:
Default browser behaviour would be opening a new tab when crtl-clicking (or cmd-click for mac) a link.

Current behaviour is that it doesn't matter whether you hold the key down or not.

PR #3723, opened 2019-03-20
Make ctrl-click open new tab

theFeu:
refs #3722

### DIFF
--- a/public/js/icinga/events.js
+++ b/public/js/icinga/events.js
@@ -457,8 +457,8 @@
                 return true;
             }
 
-            // Check for ctrl or cmd click to open new tab
-            if (event.ctrlKey || event.metaKey) {
+            // Check for ctrl or cmd click to open new tab unless clicking on a multiselect row
+            if ((event.ctrlKey || event.metaKey) && ! $a.parent().parent('table.multiselect').length > 0) {
                 window.open(href, linkTarget);
                 return false;
             }
--- a/public/js/icinga/events.js
+++ b/public/js/icinga/events.js
@@ -458,7 +458,7 @@
             }
 
             // Check for ctrl or cmd click to open new tab unless clicking on a multiselect row
-            if ((event.ctrlKey || event.metaKey) && ! $a.parent().parent('table').length > 0 && href !== '#') {
+            if ((event.ctrlKey || event.metaKey) && href !== '#' && $a.is('a')) {
                 window.open(href, linkTarget);
                 return false;
             }
--- a/public/js/icinga/events.js
+++ b/public/js/icinga/events.js
@@ -458,7 +458,7 @@
             }
 
             // Check for ctrl or cmd click to open new tab unless clicking on a multiselect row
-            if ((event.ctrlKey || event.metaKey) && ! $a.parent().parent('table.multiselect').length > 0) {
+            if ((event.ctrlKey || event.metaKey) && ! $a.parent().parent('table').length > 0 && href !== '#') {
                 window.open(href, linkTarget);
                 return false;
             }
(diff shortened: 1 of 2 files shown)


public/js/icinga/events.js
```diff
@@ -457,6 +457,12 @@
                 return true;
             }
 
+            // Check for ctrl or cmd click to open new tab
+            if (event.ctrlKey || event.metaKey) {
```
nilmerg: What's the `metaKey`? Is this an apple specific thing?

theFeu: Form the [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey)
On Macintosh keyboards, this is the ⌘ Command key. On Windows keyboards, this is the Windows key (⊞ Windows).

theFeu:
> Works very well in the menu. Though, list entries behave weird now. They open in a new tab in addition to the detail view. It also interferes with the selection of multiple rows.

I'll look into this ASAP :)

theFeu:
@nilmerg 
What behaviour would you expect on clicking the list-item?
A new browser tab opening but no detail view?
No tab opening?

This interferes with it being possible to make multiselects with the ctrl/cmd key.

So I see two possibilities on how to handle this: we keep the behaviour as it is (normal multiselect + new tab opeing) or we remove the crtl/cmd keys from triggering multiselect.

nilmerg:
I generally only expect the ctrl+click behavior to work when directly aiming at an anchor. Doing this on text or empty areas I wouldn't expect this to work. To select multiple elements I rather tend to click anywhere but the anchor.

theFeu:
should be fixed now :)

nilmerg:
Are you sure? Shows the same behavior for me.

theFeu:
Fixed me being an idiot too now ^-^'

nilmerg:
Yep, the host list and services list work fine even with multiselection. Though, some lists still open in a new tab *and* open a column. (e.g. `/icingaweb2/user/list`)

theFeu:
> Yep, the host list and services list work fine even with multiselection. Though, some lists still open in a new tab and open a column. (e.g. /icingaweb2/user/list)

And you think it should not open a new column?
Or no new tab?

Since the multiselect is not possible here, I don't see the problem with it opening a new tab - even when clicking on the row instead of the anchor.
The meta key does not serve any other function in this (except for open new tab, that is).

Plus: I personally really like having a little preview over what i just opened in a new tab :)

nilmerg:
Well, the default behavior of a browser in this case is to open the target in a new window (for browsers nowadays in a new tab) while the page where the link was clicked stays the same. Consider for a moment someone clicking on a link like this in the right-most column. The left column is lost, what's probably not what the user wanted.

theFeu:
> The left column is lost, what's probably not what the user wanted.

This is something I definitely didn't think about!

public/js/icinga/events.js
```diff
@@ -457,10 +457,16 @@
                 return true;
             }
 
+            // Check for ctrl or cmd click to open new tab unless clicking on a multiselect row
+            if ((event.ctrlKey || event.metaKey) && ! $a.parent().parent('table').length > 0 && href !== '#') {
```
nilmerg: Please use the same condition as on line no. 500 to check for multiselect tables. (And outsource this to a variable for re-use) The current condition also matches the tactical overview's fake-links and still opens a new column due to this.

theFeu: If I use the multiselect condition other tables and lists (e.g. the `/icingaweb2/user/list` that you mentioned will no longer work as intended.
The tactical overview does not really fit in with the way we display in other views in general - maybe that should change?

nilmerg: The condition I've mentioned also verifies that it's a `tr` tag. Links shouldn't be affected by this.

theFeu: So, what if instead i just verify that it is an anchor?
Just `(event.ctrlKey || event.metaKey) && $a.is('a') && href !== '#'`

nilmerg: That's also fine. Though, then I'd rather want to see the check at the very end. The `$a.is('a')` is probably the most expensive one.

theFeu:
The Tactical overview would also be less of a special case if we would change the legend to behave like other 'badge collections'.
See: #3797

PR #3735, opened 2019-04-05
Support temperatures (°C, °F) in perfdata

dasJ:
Allows for something like that:

image

### DIFF
--- a/modules/monitoring/library/Monitoring/Plugin/Perfdata.php
+++ b/modules/monitoring/library/Monitoring/Plugin/Perfdata.php
@@ -147,6 +147,16 @@ class Perfdata
     }
 
     /**
+     * Return whether this performance data's value is a temperature
+     *
+     * @return  bool    True in case it's temperature, otherwise False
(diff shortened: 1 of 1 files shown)


nilmerg:
Thanks for the PR. Sounds like a reasonable idea. Though, is there a reason why the `c` and `f` are lowercase in the perfdata source? It's not an UOM from the official guides, so don't we have the opportunity to just use `°C` and `°F` directly?

Saves us the `isTemperature` method and call to `strtoupper`. Of course, that's nitpicking, but perfdata parsing and rendering is a core part and to support temperatures the only thing really required is the `preg_match` adjustment.

dasJ:
@nilmerg Our perfdata supplies the perfdata with °C, but in parse(), icingaweb2 calls strtolower on the unit so strtoupper is required ;)

nilmerg:
Well, you're right. What a mess, I hope this will clear up with *monitoring's* new incarnation. :pray:

PR #3741, opened 2019-04-10
Fix tactical overview legend badges not properly affected by themes

theFeu:
The legend badges used the same classes as the donuts - so the background colour could be set via ```@state-xxx``` but the font colour could not be overwritten unless the non-standard class ```.slice-state-xxx``` was overwritten.

Now it should have the normal theming behaviour.


nilmerg:
Is this ready?

theFeu:
Yep!
Ready to merge :)

PR #3743, opened 2019-04-12
Add colorblind theme

theFeu:


theFeu:
As discussed with @lippserd :

We should probably take care of it, that the modules work well with themes in general - in the modules themselves.
Themes should only replace the colours and not add any specific code for each module. (As seen in 'solarized dark' already)

theFeu:
done!

Issue #2529, opened 2016-08-19
Integrate the tactical overview into the "global" search


icinga-migration:
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12492

**Created by mnardin on 2016-08-19 15:13:14 +00:00**

Hi,
when I do a search I would like to get additional information for the filtered objects at one glance (like the information bar in overview -&gt; host + services).

Best regards
Mirko

PR #3745, opened 2019-04-12
Add compact mode for Tactical Overview in monitoring

lazyfrosch:
So it will fit into a dashlet on the dashboard.


lazyfrosch:
Screenhots below, non-compact view should be unchanged.

Bildschirmfoto von 2019-04-12 13-53-08
Bildschirmfoto von 2019-04-12 13-53-19

nilmerg:
Improved this now with #3845 while working on #2529. This is now obsolete. Though, thanks for the attempt!

PR #3747, opened 2019-04-15
ApplicationBootstrap#setupTimezone(): ensure a vaild default timezone to be set in any case

Al2Klimov:


Al2Klimov:
Just before this one we check `@date_default_timezone_get()` for either a configured timezone or a fallback timezone, but we suppress errors... and future callers of `date_default_timezone_get()` may not suppress any errors:

```
[root@07b1f18d5654 /]# php -v
PHP 5.6.31 (cli) (built: Jul  6 2017 05:27:03)
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies
[root@07b1f18d5654 /]# php -a
Interactive shell

php > var_dump(@date_default_timezone_get());
string(3) "UTC"
php > var_dump(date_default_timezone_get());
PHP Warning:  date_default_timezone_get(): It is not safe to rely on the system's timezone settings. You are *required* to use the date.timezone setting or the date_default_timezone_set() function. In case you used any of those methods and you are still getting this warning, you most likely misspelled the timezone identifier. We selected the timezone 'UTC' for now, but please set date.timezone to select your timezone. in php shell code on line 1
string(3) "UTC"
php > date_default_timezone_set(@date_default_timezone_get());
php > var_dump(date_default_timezone_get());
string(3) "UTC"
php >
```

nilmerg:
> future callers of date_default_timezone_get() may not suppress any errors

Good. We didn't require setting `date.timezone` without reason after all. Anyone who doesn't care about this requirement should be reminded about it as often as possible.

Al2Klimov:
We're already trying to set a timezone if not set. But we miss it on PHP 5.6 where `@date_default_timezone_get()` returns just a fallback tz. `date_default_timezone_get()` (w/o `@`) would still produce an error w/o `date_default_timezone_set()`.

PR #3766, opened 2019-04-30
loader.js: Return in onComplete upon handled redirects

nilmerg:
Got changed with f12a5741b8a4c53b7e5d94e43094e18d5a946934. Triggering
rendered events for stuff that's never displayed doesn't seem right.


nilmerg:
This has the potential to break a few things. Especially with handlers which should fire upon finished redirects. This needs to be tested against modules which utilize the `rendered` event before merge.

nilmerg:
Modules checked and not affected:

* Doc
* Monitoring
* Graphite
* Businessprocess
* Director

PR #3780, opened 2019-05-14
MonitoredObjectController: Add handleFormatRequest() for history queries

lazyfrosch:
Especially to get SQL from history...

### DIFF
--- a/modules/monitoring/library/Monitoring/Web/Controller/MonitoredObjectController.php
+++ b/modules/monitoring/library/Monitoring/Web/Controller/MonitoredObjectController.php
@@ -108,7 +108,10 @@ abstract class MonitoredObjectController extends Controller
     public function historyAction()
     {
         $this->getTabs()->activate('history');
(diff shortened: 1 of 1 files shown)


modules/monitoring/library/Monitoring/Web/Controller/MonitoredObjectController.php
```diff
@@ -160,6 +163,8 @@ protected function handleFormatRequest($query = null)
                 ->setSuccessData($payload)
                 ->setAutoSanitize()
                 ->sendResponse();
+        } else {
+            parent::handleFormatRequest($query);
```
nilmerg: Please change this so that the base implementation is called everytime `$query` is passed. At the moment we're checking the parameters without need and don't provide a output for the given query if they match.

```php
if ($query !== null) {
    parent::handleFormatRequest($query);
} elseif ($this->params->...
```

nilmerg:
Hi, any update?

Note that in Icinga DB Web this is already a thing.

PR #3783, opened 2019-05-19
Added info about icingaweb endpoint into "System -> Monitoring Health" section

Punkoivan:
We have infrastructure like this:
Screenshot  о 21 28 05

And when I'm gonna to explore IcingaWeb logs it becomes a real pain to find proper node.

So I added info about icingaweb endpoint when using IcingaWeb in HA mode and it works for us.

In UI it looks like this:
Screenshot  о 21 33 52

I hope it is usefull. 
Please let me know if you think it should be another page to display that info.

Thanks,
Ivan

### DIFF
--- a/modules/monitoring/application/views/scripts/health/info.phtml
+++ b/modules/monitoring/application/views/scripts/health/info.phtml
@@ -57,11 +57,9 @@ if (! $this->compact): ?>
                             : $this->translate('N/A') ?></td>
                 </tr>
                 <tr>
-                    <th><?= $this->translate('Active IcingaWeb Endpoint'); ?></th>
-                    <td><?= gethostname()
-                            ? gethostname()
-                            : $this->translate('N/A') ?></td>
-                    </tr>
+                    <th><?= $this->translate('Active Icinga Web 2 Endpoint'); ?></th>
+                    <td><?= gethostname() ?: $this->translate('N/A') ?></td>
+                </tr>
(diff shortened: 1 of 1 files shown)


modules/monitoring/application/views/scripts/health/info.phtml
```diff
@@ -56,6 +56,12 @@ if (! $this->compact): ?>
                             ? $this->programStatus->endpoint_name
                             : $this->translate('N/A') ?></td>
                 </tr>
+                <tr>
+                    <th><?= $this->translate('Active IcingaWeb Endpoint'); ?></th>
+                    <td><?= gethostname()
+                            ? gethostname()
+                            : $this->translate('N/A') ?></td>
+                    </tr>
```
nilmerg: * Please ensure your indentation is correct
* `IcingaWeb` should read `Icinga Web 2`
* If the value you're comparing in a ternary operation is about to be used as truthy result anyway you can shorten the statement by just using `?:`

```php
                <tr>
                    <th><?= $this->translate('Active Icinga Web 2 Endpoint'); ?></th>
                    <td><?= gethostname() ?: $this->translate('N/A') ?></td>
                </tr>
```

Punkoivan: Thanks @nilmerg 
I fixed with your suggestion and checked indentation.
Please see updated code.

Issue #3461, opened 2018-05-18
Keep closing "x" for panes available at all times


widhalmt:
When viewing Icinga Web 2.5.1 with a very low resolution (due to mirroring to a legacy beamer) the "x" button to close a pane disappears on the left pane when I have "Dashboards" on the left and some detailview on the right.

I have the three default dashboards plus two custom ones in a testing environment. 

![missing-x](https://user-images.githubusercontent.com/504849/40225389-3151eb3c-5a89-11e8-8313-3425c86d04db.PNG)

I'm totally ok with some dashboards being invisible when I have to use a minimal resolution but I'd really like to keep the "x" to close this pane.

Workaround: Rewrite the URL so the left pane is removed.

lippserd:
Hi Thomas,

Thanks for the report. Yep, we should fix that.

Cheers,
Eric

theFeu:
Everything that falls off due to the narrow pane width should be available under a '...' menu entry.

nilmerg:
Dependent of #3810 -> v2.8

tbauriedel:
ref/NC/804873

Saeid-Abadi:
ref/NC/824374

PR #3787, opened 2019-05-21
Adapt pane controls to available width with dropdown

theFeu:
in regard to #3461

### DIFF
--- a/public/js/icinga/ui.js
+++ b/public/js/icinga/ui.js
@@ -678,6 +678,7 @@
             // Enable this only in case you want to track down UI problems
             //this.icinga.logger.debug('Fixing controls for ', $container);
 
+            var _this = this;
             $container.find('.controls').each(function() {
                 var $controls = $(this);
                 var $fakeControls = $controls.prev('.fake-controls');
@@ -689,6 +690,8 @@
                     top: $container.offsetParent().position().top,
                     width: $fakeControls.outerWidth()
                 });
+
+                _this.ensureTabVisibility($controls);
             });
 
             var $statusBar = $container.children('.monitoring-statusbar');
@@ -699,6 +702,43 @@
                 });
                 $statusBar.prev('.monitoring-statusbar-ghost').height($statusBar.outerHeight(true));
             }
+        },
+
+        ensureTabVisibility: function ($controls) {
+            var $tabContainer = $controls.find('.tabs', $controls);
+            var $tabs = $tabContainer.children('li');
+            var $dropdown = $tabContainer.find('.tabs-dropdown');
+            var $visibleCount = $tabs.has(':visible').size();
+
+            if ($tabContainer[0].scrollHeight > $tabContainer[0].clientHeight) {
+                if ($dropdown.is(":visible")) {
+                    $tabs.slice($visibleCount - 5, -4).css('display', 'none');
+                } else {
+                    $tabs.slice($visibleCount - 4, -3).css('display', 'none');
+                }
+                $dropdown.css('display', 'inline-block');
+            } else if ($tabContainer.find('.tabs-dropdown').is(':visible') && $tabContainer.children().is(':hidden')) {
+                var $visibleWidth = 0;
+                $.each($tabs.has(':visible'), function(){
+                    $visibleWidth += $(this).width();
+                });
+                for (var count = 0; count < $tabs.size() - 4; count++) {
+                    if($($tabs[count]).is(":hidden")) {
+                        if(($($tabs[count]).width()) < ($tabContainer.width() - $visibleWidth - $visibleCount * 5)) {
+                            $($tabs[count]).css('display', 'inline-block');
+                            if ($($tabs[count + 1]).is(":visible")) {
+                                $dropdown.css('display', 'none');
+                            }
+                        }
+                        break;
+                    }
+                }
+            }
+
+            var $dropdownList = $dropdown.children('ul').children('li');
+            $visibleCount = $tabs.has(':visible').size();
+            $dropdownList.slice(0, $visibleCount - 4).css('display', 'none');
+            $dropdownList.slice($visibleCount - 4).css('display', 'block');
         },
 
         toggleFullscreen: function () {
--- a/library/Icinga/Web/Widget/Tabs.php
+++ b/library/Icinga/Web/Widget/Tabs.php
@@ -23,6 +23,7 @@ class Tabs extends AbstractWidget implements Countable
     private $baseTpl = <<< 'EOT'
 <ul class="tabs primary-nav nav">
   {TABS}
+  {TABSDROPDOWN}
   {DROPDOWN}
   {REFRESH}
   {CLOSE}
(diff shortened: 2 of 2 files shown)


flourish86:
Please replace the icon for the menu with the `chevron-double-right`, as it is more common for things like that. Alternatively an ellipsis `…` would also work, but the one in FontAwesome is too heavy in comparison to the other icons.

theFeu:
Problem: when the active tab is in the dropdown, no tab is shows as active
Proposed solution: mark the dropdown as .active

flourish86 (dismissed):
Works fine! 👍

public/js/icinga/ui.js
```diff
@@ -701,6 +704,43 @@
             }
         },
 
+        ensureTabVisibility: function ($controls) {
+            var $tabContainer = $controls.find('.tabs', $controls);
+            var $tabs = $tabContainer.children('li');
+            var $dropdown = $tabContainer.find('.tabs-dropdown');
+            var $visibleCount = $tabs.has(':visible').size();
```
nilmerg: `$tabs.filter(':visible')` has better performance as per the [documentation](https://api.jquery.com/visible-selector/). And since we're only interested in the tabs themselves (not their content) this suffices.

public/js/icinga/ui.js
```diff
@@ -701,6 +704,43 @@
...
+
+            if ($tabContainer[0].scrollHeight > $tabContainer[0].clientHeight) {
+                if ($dropdown.is(":visible")) {
+                    $tabs.slice($visibleCount - 5, -4).css('display', 'none');
+                } else {
+                    $tabs.slice($visibleCount - 4, -3).css('display', 'none');
+                }
```
nilmerg: Please use `.hide()` instead of `.css('display', 'none')`.

public/js/icinga/ui.js
```diff
@@ -701,6 +704,43 @@
...
+                $dropdown.css('display', 'inline-block');
```
nilmerg: Here as well: `.show()` please

public/js/icinga/ui.js
```diff
@@ -701,6 +704,43 @@
...
+            } else if ($tabContainer.find('.tabs-dropdown').is(':visible') && $tabContainer.children().is(':hidden')) {
+                var $visibleWidth = 0;
+                $.each($tabs.has(':visible'), function(){
```
nilmerg: Another case of `$tabs.filter(':visible')`, I guess

public/js/icinga/ui.js
```diff
@@ -701,6 +704,43 @@
...
+                    $visibleWidth += $(this).width();
+                });
+                for (var count = 0; count < $tabs.size() - 4; count++) {
+                    if($($tabs[count]).is(":hidden")) {
+                        if(($($tabs[count]).width()) < ($tabContainer.width() - $visibleWidth - $visibleCount * 5)) {
+                            $($tabs[count]).css('display', 'inline-block');
```
nilmerg: `.show()`

public/js/icinga/ui.js
```diff
@@ -701,6 +704,43 @@
...
+                            if ($($tabs[count + 1]).is(":visible")) {
+                                $dropdown.css('display', 'none');
```
nilmerg: `.hide()`

public/js/icinga/ui.js
```diff
@@ -701,6 +704,43 @@
...
+                            }
+                        }
+                        break;
+                    }
+                }
+            }
+
+            var $dropdownList = $dropdown.children('ul').children('li');
+            $visibleCount = $tabs.has(':visible').size();
```
nilmerg: `$tabs.filter(':visible')`

public/js/icinga/ui.js
```diff
@@ -701,6 +704,43 @@
...
+            $dropdownList.slice(0, $visibleCount - 4).css('display', 'none');
+            $dropdownList.slice($visibleCount - 4).css('display', 'block');
```
nilmerg: `.hide()` and `.show()` :wink:

public/js/icinga/ui.js
```diff
@@ -701,6 +704,43 @@
...
+                            if ($($tabs[count + 1]).is(":visible")) {
+                                $dropdown.css('display', 'none');
+                            }
+                        }
+                        break;
+                    }
+                }
+            }
+
+            var $dropdownList = $dropdown.children('ul').children('li');
```
nilmerg: Is it necessary to first select the ul? Can't you directly select the li's?

theFeu:
#### It's not possible to close the dropdown by clicking on the arrows again

This is intended, the other dropdowns work the same way.

#### The dropdown overlaps with the menu 

While trying to reproduce this I found that the dropdown also does not adapt to long items: 
image
But it doesn't overlap with the menu (for me)...


#### The close button may still be missing

Still researching how this happens...


#### Sometimes a tab is hidden and sometimes not while in the same view

Has something to do with the unfolding - I made it so, that it needs some extra space to unfold in order to prevent twitching when you get close to where the fold break is.

#### When trying to select an entry in the menu it disappears too easily

This is also the case with the image dropdown. I think this should be raised in another issue.

nilmerg:
* It's not possible to close the dropdown by clicking on the arrows again
> This is intended, the other dropdowns work the same way.

That's strange. I mentioned this only because the dropdown didn't disappear on mouse-leave for me and I had to click somewhere else to close it. When trying to reproduce this now it works fine. :thinking: 

* The dropdown overlaps with the menu
> But it doesn't overlap with the menu (for me)...

I noticed this while emulating an iPad with the dev-tools (768x1024px)

* The close button may still be missing
> Still researching how this happens...

I noticed this also while emulating an iPad. Occurred only when choosing an entry from the dropdown.

flourish86:
> While trying to reproduce this I found that the dropdown also does not adapt to long items

Please use `text-overflow: ellipsis` for this

theFeu:
@nilmerg 
@flourish86 and I decided to leave the dropdown overlap as it is for now - it should be fixed with dropping the fixed positioning - ref #3810

theFeu:
Found out why the close button is sometimes still missing:

It happens when you have to many tabs that need hidin - it is only called 2(?) times and that might not be enough to hide all tabs necessary

theFeu:
**in single col views the number of `<li>`s is wrong**
this is due to counting on a fixed amount of controls - where of course the closing control is missing.

**the selected `<li>` is not visibly active when hidden**
should look like this: 
image

nilmerg:
Just tried it again and got this:

```
Uncaught RangeError: Maximum call stack size exceeded
    at String.replace (<anonymous>)
    at Sizzle (icinga.dev.js:838)
    at Function.Sizzle.matches (icinga.dev.js:1367)
    at Function.jQuery.filter (icinga.dev.js:2619)
    at winnow (icinga.dev.js:2599)
    at jQuery.fn.init.filter (icinga.dev.js:2651)
    at Icinga.UI.ensureTabVisibility (icinga.dev.js:14137)
    at Icinga.UI.ensureTabVisibility (icinga.dev.js:14134)
    at Icinga.UI.ensureTabVisibility (icinga.dev.js:14148)
    at Icinga.UI.ensureTabVisibility (icinga.dev.js:14134)
```

Reproducible by:
* Reloading the window
* Opening a link that causes a second column to open (or just have it already open)
* Narrowing the window by making it smaller on the right
* Notice that the right column gets cut off
* Wait, upon auto-refresh the error appears

Oh, and I just noticed how I've got the dropdown to stay permanently. That's the case when emulating a touch screen as I have to tap the arrow then which puts the focus on it. Sure, that's also the case for the other (export) dropdown. Inconvenient nonetheless. But may very well be fixed in another issue.

PR #3789, opened 2019-05-21
Fix main layout for use without js

### LINKED ISSUE #3629: Main layout is broken without JS
## Expected Behavior
Basic functionality should be available without JS, page should look good.

## Current Behavior
It looks ugly, we broke this with v2.5

Disable JavaScript, load the page

flourish86:
fixes #3629

### DIFF
--- a/application/forms/Control/LimiterControlForm.php
+++ b/application/forms/Control/LimiterControlForm.php
@@ -15,7 +15,7 @@ class LimiterControlForm extends Form
      *
      * @var string
      */
-    const CSS_CLASS_LIMITER = 'limiter-control icinga-controls';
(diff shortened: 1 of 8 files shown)


nilmerg:
Also, seems to still have problems with announcements:

Expected:
announcements-js

Actual:
announcements-no-js

Rebase with master once #3792 is merged to get multiple announcements working.

flourish86:
Just for the protocol. There are better solutions for a js-less ui.

Since we run into problems with the fixed tabs all the time, maybe a `position: fixed`-less layout would be the better option in the long run.

See how WordPress has done it with their backend. This is much more stable and less prone to errors and oddities.

Screen Recording 2019-05-23 at 15 04 55 2019-05-23 15_06_49

PR #3790, opened 2019-05-21
Update documentation to fetch keys securely

jugglinmike:
Instruct users to retrieve APT keys via a secure connection to avoid
man-in-the-middle attacks.


nilmerg:
Thanks! I'll just use this as an opportunity to make all the other links go over `https` :no_mouth:

jugglinmike:
Great! Thanks for Icinga :)

PR #3801, opened 2019-05-27
Improve host-service distinction in lists

### LINKED ISSUE #3615: Better distinction between hosts and services (more clickable looking)
When I watch certain users use the web interface, I can see they struggle to distinguish between hosts and services most often. 
This is apparent for example in the Service Problems column as per this picture:

Service Problems

The tendency is for them to click on the hostname and then get lost in the interface a little as it didn't show the service they were expecting.

Even in recently recovered services, they often click on the hostname, and either don't see the services tab, nor the little icon they can click to show the services. Perhaps the interface elements don't appear as 'clickable' as they could, or perhaps they need a larger hint as to where services are. I often have to guide new users as to where to click to see the services correctly.

I'm not sure what the solution is, but I'm hoping someone can suggest a neat way to make the distinction  little broader, so users can find services more easily.
In fact, the word '12 Services' in this image doesn't even look clickable.

Where is services again
by contrast the 'Check Now' is clearly clickable, due to it being a different font and colour.

Perhaps just making "12 Services" more similar to "Check Now", with a larger font and an icon would make it clearer for people?

theFeu:
fixes #3615

### DIFF
--- a/modules/monitoring/application/views/scripts/list/services.phtml
+++ b/modules/monitoring/application/views/scripts/list/services.phtml
@@ -61,33 +61,56 @@ if (! $this->compact): ?>
 
             <td>
                 <div class="state-header">
-                    <?= $this->iconImage()->service($service) ?>
-                    <?php if ($this->showHost): ?><?= $this->qlink(
-                        $service->host_display_name
-                        . ($service->host_state != 0 ? ' (' . Host::getStateText($service->host_state, true) . ')' : ''),
-                        $hostLink,
-                        null,
-                        array(
-                            'title' => sprintf(
-                                $this->translate('Show detailed information for host %s'),
-                                $service->host_display_name
-                            )
-                        )
-                    ) ?>&#58;
-                    <?php endif ?><?= $this->qlink(
-                        $service->service_display_name,
-                        $serviceLink,
-                        null,
-                        array(
-                            'title' => sprintf(
-                                $this->translate('Show detailed information for service %s on host %s'),
-                                $service->service_display_name,
-                                $service->host_display_name
-                            ),
-                            'class' => 'rowaction'
-                        )
-                    ) ?>
-                    <span class="state-icons"><?= $this->serviceFlags($service) ?></span>
+                  <span class="service-on">
+                  <?= $this->iconImage()->service($service) ?>
+                  <?php
+                  if ($this->showHost) {
+                      echo sprintf(
+                          $this->translate('%s on %s', 'service on host'),
+                          $this->qlink(
+                              $service->service_display_name,
+                              $serviceLink,
+                              null,
+                              array(
+                                  'title' => sprintf(
+                                      $this->translate('Show detailed information for service %s on host %s'),
+                                      $service->service_display_name,
+                                      $service->host_display_name
+                                  ),
+                                  'class' => 'rowaction'
+                              )
+                          ),
+                              $this->qlink(
+                                  $service->host_display_name
+                                  . ($service->host_state != 0 ? ' (' . Host::getStateText($service->host_state, true) . ')' : ''),
+                                  $hostLink,
+                                  null,
+                                  [
+                                      'title' => sprintf(
+                                          $this->translate('Show detailed information for host %s'),
+                                          $service->host_display_name
+                                      )
+                                  ]
+                              )
+                          );
+                  } else {
+                      echo $this->qlink(
+                          $service->service_display_name,
+                          $serviceLink,
+                          null,
+                          array(
+                              'title' => sprintf(
+                                  $this->translate('Show detailed information for service %s on host %s'),
+                                  $service->service_display_name,
+                                  $service->host_display_name
+                              ),
+                              'class' => 'rowaction'
+                          )
+                      );
+                  }
+                  ?>
+                  </span>
+                  <span class="state-icons"><?= $this->serviceFlags($service) ?></span>
                 </div>
                 <div class="overview-plugin-output-container">
                     <div class="overview-performance-data">
--- a/modules/monitoring/application/views/scripts/partials/comment/comment-detail.phtml
+++ b/modules/monitoring/application/views/scripts/partials/comment/comment-detail.phtml
@@ -10,7 +10,6 @@
                 'host' => $comment->host_name,
                 'service' => $comment->service_description
             ],
-            null,
             [
                 'title' => sprintf(
                     $this->translate('Show detailed information for service %s on host %s'),
@@ -22,11 +21,8 @@
         ),
         $this->qlink(
             $comment->host_display_name,
-            $this->href(
-                'monitoring/host/show',
-                ['host' => $comment->host_name]
-            ),
-            null,
+            'monitoring/host/show',
+            ['host' => $comment->host_name],
             [
                 'title' => sprintf(
                     $this->translate('Show detailed information for host %s'),
--- a/modules/monitoring/application/views/scripts/partials/comment/comment-detail.phtml
+++ b/modules/monitoring/application/views/scripts/partials/comment/comment-detail.phtml
@@ -1,23 +1,42 @@
 <div class="comment-author">
 <?php if ($comment->objecttype === 'service') {
-    echo $this->icon('service', $this->translate('Service'));
-    echo $this->qlink(
-        $comment->host_display_name . ': ' . $comment->service_display_name,
-        'monitoring/service/show',
-        array(
-            'host' => $comment->host_name,
-            'service' => $comment->service_description
+    echo '<span class="service-on">';
+    echo sprintf(
+        $this->translate('%s on %s', 'service on host'),
+        $this->qlink(
+            $comment->service_display_name,
+            'monitoring/service/show',
+            [
+                'host' => $comment->host_name,
+                'service' => $comment->service_description
+            ],
+            null,
+            [
+                'title' => sprintf(
+                    $this->translate('Show detailed information for service %s on host %s'),
+                    $comment->service_display_name,
+                    $comment->host_display_name
+                ),
+                'class' => 'rowaction'
+            ]
         ),
-        array(
-            'title' => sprintf(
-                $this->translate('Show detailed information for this comment about service %s on host %s'),
-                $comment->service_display_name,
-                $comment->host_display_name
-            )
+        $this->qlink(
+            $comment->host_display_name,
+            $this->href(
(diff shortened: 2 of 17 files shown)


flourish86:
Please also apply to comment list

modules/monitoring/application/views/scripts/list/services.phtml
```diff
@@ -80,6 +68,19 @@ if (! $this->compact): ?>
                             'class' => 'rowaction'
                         )
                     ) ?>
+                    <?php if ($this->showHost): ?> on <?= $this->qlink(
```
nilmerg: Please translate `on`

modules/monitoring/application/views/scripts/partials/comment/comment-detail.phtml
```diff
@@ -1,17 +1,34 @@
...
-            'service' => $comment->service_description
+            'title' => sprintf(
+                $this->translate('Show detailed information for service %s on host %s'),
+                $comment->service_display_name,
+                $comment->host_display_name
+            ),
+            'class' => 'rowaction'
+        )
+    );
+    echo ' on ';
```
nilmerg: Please translate this

modules/monitoring/application/views/scripts/partials/comment/comment-detail.phtml
```diff
@@ -1,17 +1,34 @@
 <div class="comment-author">
 <?php if ($comment->objecttype === 'service') {
-    echo $this->icon('service', $this->translate('Service'));
```
nilmerg: Why is the icon removed? A host event still got an icon.

modules/monitoring/application/views/scripts/partials/downtime/downtime-header.phtml
```diff
@@ -10,21 +10,41 @@
 <td>
   <div class="comment-author">
     <?php if ($isService): ?>
-        <?= $this->icon('service', $this->translate('Service')); ?> <?= $this->qlink(
```
nilmerg: Again, the icon. Why? Also here hosts got still an icon.

modules/monitoring/application/views/scripts/partials/downtime/downtime-header.phtml
```diff
@@ -10,21 +10,41 @@
...
+        array(
+            'title' => sprintf(
+                $this->translate('Show detailed information for service %s on host %s'),
+                $downtime->service_display_name,
+                $downtime->host_display_name
+            ),
+            'class' => 'rowaction'
+        )
+    );
+        echo ' on ';
```
nilmerg: Translation, pls

modules/monitoring/application/views/scripts/partials/downtime/downtime-header.phtml
```diff
@@ -10,21 +10,41 @@
...
-                'service' => $downtime->service_description
+    <?php
+        echo $this->qlink(
+        $downtime->service_display_name,
+        'monitoring/service/show',
+        [
+            'host' => $downtime->host_name,
+            'service' => $downtime->service_description
+        ]
+        ,
```
nilmerg: Huh?

modules/monitoring/application/views/scripts/partials/downtime/downtime-header.phtml
```diff
@@ -10,21 +10,41 @@
...
   <div class="comment-author">
     <?php if ($isService): ?>
-        <?= $this->icon('service', $this->translate('Service')); ?> <?= $this->qlink(
-            $downtime->host_display_name . ': ' . $downtime->service_display_name,
-            'monitoring/service/show',
-            array(
-                'host'    => $downtime->host_name,
-                'service' => $downtime->service_description
+    <?php
+        echo $this->qlink(
```
nilmerg: If you really don't want to use `<?= ... ?>` (I fail to see why) please pay attention to indentation.

modules/monitoring/public/css/tables.less
```diff
@@ -209,8 +209,13 @@
 }
 
 // Wraps links, icons and meta in overviews
-.state-header {
+.state-header,
+.comment-author {
   .clearfix();
+
+  > a {
+    font-weight: bold;
```
nilmerg: Is there room for a compromise? Maybe a less bold look?

modules/monitoring/application/views/scripts/partials/comment/comment-detail.phtml
flourish86: I think the icon won't work as intended. Especially because, there isn't any consistency in other views, such as service~ and host~.

@theFeu: Of course, we should remove the icon consistently, namely downtime list as well.

modules/monitoring/application/views/scripts/partials/downtime/downtime-header.phtml
flourish86: see above

modules/monitoring/public/css/tables.less
flourish86: Since we use different OS fonts, the only way to ensure to have bold fonts is to use the descriptive value `bold`. If you use numeric values, some fonts behave different than others.

theFeu:
- Translate
- Wrap 'on' in span with a class for styling like
`sprintf('%s<span class="service-on">%s</span>%s', $service, 'on', $host)`

nilmerg:
Thanks for the changes. Though, I've noticed another issue with how `on` is translated now.

We usually follow the principle that full sentences are marked as translatable if possible. This allows translators to slightly change the structure as they see fit. So marking single words is only a thing if they really stand by themselves eventually. This also makes it instantly obvious what semantic such a word has. Especially `on` may have multiple meanings depending on how it's used.

In this case the translatable string should be `%s on %s`. A context of `service on host` makes its semantic even more obvious. (`$this->translate('%s on %s', 'service on host')`)

The styling of course then needs to be inverted. The base style on the `a` and special styles on e.g. `span`s by which the service description and host name are wrapped. ( :thinking: With the side-effect of being able to target these individually in a custom theme :sunglasses: )

modules/monitoring/application/views/scripts/list/services.phtml
```diff
@@ -54,33 +54,39 @@ if (! $this->compact): ?>
...
+                          array(
+                              'title' => sprintf(
+                                  $this->translate('Show detailed information for service %s on host %s'),
+                                  $service->service_display_name,
+                                  $service->host_display_name
+                              ),
+                              'class' => 'rowaction'
+                          )
+                      ),
+                      ! $this->showHost ? '' :
```
lippserd: `showHost` is set to `false` for the services list in the host detail areas. There you now have "ping on"  for example. How does an empty string for the hostname ever makes sense in the "service on host" format?

modules/monitoring/application/views/scripts/list/services.phtml
```diff
@@ -54,33 +54,39 @@ if (! $this->compact): ?>
...
+                      $this->qlink(
+                          $service->host_display_name
+                          . ($service->host_state != 0 ? ' (' . Host::getStateText($service->host_state, true) . ')' : ''),
+                          $hostLink,
+                          null,
+                          array(
```
lippserd: Use short array syntax.

modules/monitoring/application/views/scripts/partials/comment/comment-detail.phtml
```diff
@@ -1,23 +1,42 @@
...
+        $this->translate('%s on %s', 'service on host'),
+        $this->qlink(
+            $comment->service_display_name,
+            'monitoring/service/show',
+            [
+                'host' => $comment->host_name,
+                'service' => $comment->service_description
+            ],
+            null,
+            array(
```
lippserd: Use short array syntax.

modules/monitoring/application/views/scripts/partials/comment/comment-detail.phtml
```diff
@@ -1,23 +1,42 @@
...
-                $comment->service_display_name,
-                $comment->host_display_name
+        $this->qlink(
+            $comment->host_display_name,
+            $this->href(
+                'monitoring/host/show',
+                ['host' => $comment->host_name]
+            ),
+            null,
+            array(
```
lippserd: Use short array syntax.

modules/monitoring/application/views/scripts/partials/downtime/downtime-header.phtml
```diff
@@ -9,24 +9,45 @@
...
-                'host'    => $downtime->host_name,
-                'service' => $downtime->service_description
+    <?php if ($isService) {
+        echo '<span class="service-on">';
+        echo sprintf(
+            $this->translate('%s on %s', 'service on host'),
+            $this->qlink(
+                $downtime->service_display_name,
+                'monitoring/service/show',
+                array(
```
lippserd: Use short array syntax.

modules/monitoring/application/views/scripts/partials/downtime/downtime-header.phtml
```diff
@@ -9,24 +9,45 @@
...
+                    'host' => $downtime->host_name,
+                    'service' => $downtime->service_description
+                ),
+                null,
+                array(
```
lippserd: Use short array syntax.

modules/monitoring/application/views/scripts/partials/downtime/downtime-header.phtml
```diff
@@ -9,24 +9,45 @@
...
-                    $downtime->service_display_name,
-                    $downtime->host_display_name
+            $this->qlink(
+                $downtime->host_display_name,
+                $this->href(
+                    'monitoring/host/show',
+                    ['host' => $downtime->host_name]
+                ),
+                null,
+                array(
```
lippserd: Use short array syntax.

flourish86:
Wouldn't this be nice to have in 2.7.0?

nilmerg:
@lippserd Are we done here?

lippserd:
Yep, done. Made the host and service names bold in their details areas too. Though the names are not clickable in some scenarios, it just makes sense from a design point of view to have them bold as well. Because else the names are prominent in the list views but not in the detail areas.

flourish86:
Sorry, but the `on` looks super odd. Can we at least remove the `letter-spacing`?

Screen Recording 2019-07-29 at 13 49 20 2019-07-29 13_57_44

Issue #3793, opened 2019-05-23
Service grid problem toggle


nilmerg:
## Is your feature request related to a problem? Please describe.
The service grid shows only problems by default. By removing the `problems` parameter the grid shows everything though this is not obvious enough. (#3788)

## Describe the solution you'd like
A toggle to switch between *only problems* and *everything*. Similar to what we have for group overviews.

theFeu:
In all of the other views in the 'Problems' tab there is a filter, that you can just remove:
![image](https://user-images.githubusercontent.com/22217062/58249244-9d3e1480-7d5e-11e9-806a-14071ffa0ed0.png)

Why don't we have this in the service grid?

nilmerg:
Because it's a rather special filter. Any other filter is utilized with a comparison. (e.g. `something=else`) and applied on the result directly. `problems` however is not applied directly, it's processed differently and so putting it in the filter editor does not work. That's because I think we should provide a toggle switch as well here. (Just like the service grid already has one for `?flipped`)

PR #3806, opened 2019-05-29
Add checkbox to toggle problems parameter in service grid

theFeu:
refs #3793

### DIFF
--- a/modules/monitoring/application/views/scripts/list/servicegrid-flipped.phtml
+++ b/modules/monitoring/application/views/scripts/list/servicegrid-flipped.phtml
@@ -10,6 +10,26 @@ if (! $this->compact): ?>
         <?= $this->sortBox ?>
     </div>
     <?= $this->filterEditor ?>
+    <?php
+    $toggleFilter = Url::fromRequest();
+    if ($toggleFilter->getParam('problems')) {
+        $toggleFilter->remove('problems');
+    } else {
+        $toggleFilter->setParam('problems');
+    }
+    ?>
+  <label class="relevance-filter">
+    <input type="checkbox" <?= (bool) $toggleFilter->getParam('problems') ?: "checked" ?>>
+      <?= $this->qlink(
+          $this->translate('relevance filter'),
+          $toggleFilter,
+          null,
+          array(
+              'data-base-target'    => '_self',
+              'title'               => $this->translate('Toggle relevance filter')
+          )
+      ) ?>
+  </label>
 </div>
 <?php endif ?>
 <div class="content" data-base-target="_next">
--- a/modules/monitoring/application/views/scripts/list/servicegrid-flipped.phtml
+++ b/modules/monitoring/application/views/scripts/list/servicegrid-flipped.phtml
@@ -9,27 +9,26 @@ if (! $this->compact): ?>
     <div class="sort-controls-container">
         <?= $this->sortBox ?>
     </div>
-    <?= $this->filterEditor ?>
     <?php
     $toggleFilter = Url::fromRequest();
     if ($toggleFilter->getParam('problems')) {
+        $title = 'Show entire environment';
         $toggleFilter->remove('problems');
     } else {
         $toggleFilter->setParam('problems');
+        $title = 'Show problem context';
     }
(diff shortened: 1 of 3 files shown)


modules/monitoring/application/views/scripts/list/servicegrid-flipped.phtml
```diff
@@ -10,6 +10,26 @@ if (! $this->compact): ?>
...
     <?= $this->filterEditor ?>
+    <?php
+    $toggleFilter = Url::fromRequest();
+    if ($toggleFilter->getParam('problems')) {
+        $toggleFilter->remove('problems');
+    } else {
+        $toggleFilter->setParam('problems');
+    }
+    ?>
+  <label class="relevane-filter">
```
dnsmichi: This should be called `relevanCe`, right? There are other locations too.

theFeu: That is most definitely a typo!
Thanks for pointing it out!

modules/monitoring/application/views/scripts/list/servicegrid-flipped.phtml
```diff
@@ -9,6 +9,25 @@ if (! $this->compact): ?>
...
+    if ($toggleFilter->getParam('problems')) {
+        $title = 'Show entire environment';
+        $toggleFilter->remove('problems');
+    } else {
+        $toggleFilter->setParam('problems');
+        $title = 'Show problem context';
+    }
+    ?>
+    <?= $this->qlink(
+        $this->translate($title),
```
nilmerg: Translating a variable causes gettext to not identify this as translatable. Please pass a string here directly.

modules/monitoring/application/views/scripts/list/servicegrid-flipped.phtml
```diff
@@ -9,6 +9,25 @@ if (! $this->compact): ?>
...
+        $toggleFilter,
+        null,
+        array(
+            'data-base-target'    => '_self',
+            'title'               => $this->translate('Toggle relevance filter')
```
nilmerg: The `data-base-target` is not required here. `_self` is the default behavior and not adjusted otherwise by a parent. Also the `title` is, since the link got a nice label now, rather obsolete. Thus the `qlink` call can be shortened I guess.

modules/monitoring/application/views/scripts/list/servicegrid.phtml
```diff
@@ -9,7 +9,27 @@ if (! $this->compact): ?>
...
+        $this->translate($title),
+        $toggleFilter,
+        null,
+        array(
+            'data-base-target'    => '_self',
+            'title'               => $this->translate('Toggle relevance filter')
+        )
+    ) ?>
     <?= $this->filterEditor ?>
+  </label>
```
nilmerg: That's a remnant, I guess.

modules/monitoring/public/css/service-grid.less
```diff
@@ -51,6 +51,13 @@
   }
 }
 
+.relevance-filter {
```
nilmerg: Since the checkbox is now gone this is not required anymore, please remove.

lippserd:
@theFeu Could you please redo your commit message? I know that this may be a subject to squash but it still makes sense to summarize what you've done.

theFeu:
> @theFeu Could you please redo your commit message? I know that this may be a subject to squash but it still makes sense to summarize what you've done.

Apologies, I will pay more attention to that!

lippserd:
A link feels wrong to me here. I would expect a checkbox labelled "Show only problems" next to the search bar. With our new form implementation, this does even look good 😆

nilmerg:
> With our new form implementation, this does even look good

Hard to believe. Good looking checkboxes :thinking: :rofl:

theFeu:
Apart from the wording being odd (since it does not only show problems) I feel like neither the checkbox nor the link are fitting...

I honestly feel like it should have the flair of a filter (because that's what it is?)

nilmerg:
Superseded by #3871

PR #3814, opened 2019-06-04
Make notes, comments and announcements markdown aware

### LINKED ISSUE #2831: Allow HTML in host and service notes
In the notes attribute of host and service objects it is currenty not possible to add html tags which are correctly rendered in web2.

The possibility to show rendered html here would be enough.

### LINKED ISSUE #3441: Transform any URL in a Comment to a clickable Link
Hi,

I think it would be good if any url in a comment is transformed to a clickable Link within icingaweb2, like in my module: https://github.com/cstegm/icingaweb2-module-commenturls

### LINKED ISSUE #3684: Make notes markdown-aware
## Expected Behavior
Every Service we monitor gets a very detailed description of what has to be done if the service enters warning oder critical states. These descriptions can contain links to other resources.

## Current Behavior
Currently there is no way to provide links in notes to services or even highlight important things about the description to the service.

## Possible Solution
The Director should be able to render markdown (for example).

nilmerg:
Adds a new view helper `Markdown` which allows to transform markdown.

`$this->markdown($text)`

At the moment markdown is transformed for notes, comments and announcements. (yep, I thought this suits here as well)

resolves #3684 
resolves #3441 
resolves #2831

### DIFF
--- a/modules/monitoring/application/views/scripts/show/components/notes.phtml
+++ b/modules/monitoring/application/views/scripts/show/components/notes.phtml
@@ -38,7 +38,7 @@ if (($navigation->isEmpty() || ! $navigation->hasRenderableItems()) && $notes ==
     <td>
         <?= $navigation->getRenderer() ?>
     <?php if ($notes !== ''): ?>
-        <p><?= $this->nl2br($this->escape($notes)) ?></p>
+        <section class="notes"><?= $this->markdown($notes) ?></section>
     <?php endif ?>
     </td>
 </tr>
--- a/library/Icinga/Web/FileCache.php
+++ b/library/Icinga/Web/FileCache.php
@@ -155,6 +155,27 @@ class FileCache
     }
 
     /**
+     * Prepare a sub directory with the given name and return its path
+     *
+     * @param   string  $name
+     *
+     * @return  string|false    Returns FALSE in case the cache is not enabled or an error occurred
+     */
+    public function directory($name)
+    {
+        if (! $this->enabled) {
+            return false;
+        }
+
+        $path = $this->filename($name);
+        if (! is_dir($path) && ! @mkdir($path, octdec('1750'), true)) {
+            return false;
+        }
+
+        return $path;
+    }
+
+    /**
      * Whether the given ETag matches a cached file
      *
      * If no ETag is given we'll try to fetch the one from the current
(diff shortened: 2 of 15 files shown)


lippserd:
I don't see why we should limit the allowed tags or CSS attributes for either text or line. We had kind of the same problem with the the plugin output. We added more and more allowed tags and CSS attributes over time. I think users would like the full flexibility. The default configuration should be sufficient in terms of security. Maybe we should consider setting Parsedown's safe mode to `true` with `Parsedown:: setSafeMode()`. The `Markdown` class should cache the `Parsedown` and `HtmlPurifier` instances.

nilmerg:
Tags are not limited anymore. I've just omitted the esoteric ones such as e.g. `ruby`, `details` and `tt`. Basically everything parsedown may possibly produce is allowed.

Enabling safe mode is a good idea, yes. Though, prior allowing all attributes I'd verify that things like `onclick` or `onhover` don't work.

The Markdown class doesn't need to cache anything as the Parsedown class already caches its own instance and caching the HtmlPurifier does not help much. I'd rather introduce a [cache directory](https://github.com/Icinga/icingaweb2/blob/master/library/Icinga/Web/Helper/HtmlPurifier.php#L35).

nilmerg:
Removed the attribute limitations. HtmlPurifier is already smart enough to remove all risky ones. (e.g. `onclick`)

Parsedown's safe mode however also escapes any raw markup and only allows markdown syntax. And now switching over to only markdown doesn't seem right. So safe mode is not an option.

lippserd:
Sounds good. But the HTML tags are still limited. Why not just allow them all?

nilmerg:
Removed tag limitations and dropped the distinction between phrasing and block level content. It's all everything now.

Also, HtmlPurifier now got its own cache directory in our file cache.

modules/monitoring/application/views/scripts/show/components/notes.phtml
```diff
@@ -38,7 +38,7 @@ if (($navigation->isEmpty() || ! $navigation->hasRenderableItems()) && $notes ==
     <td>
         <?= $navigation->getRenderer() ?>
     <?php if ($notes !== ''): ?>
-        <p><?= $this->nl2br($this->escape($notes)) ?></p>
+        <section class="notes"><?= $this->markdown($notes) ?></section>
```
lippserd: Please use just a `div` here. We are far away from using semantic HTML 😆

nilmerg: It's a `<section>` because of `#` == `<h1>` and multiple `<h1>` are only valid if separated by `<section>`.

lippserd: That makes sense.

PR #3816, opened 2019-06-04
Change announcement color to yellow

### LINKED ISSUE #3560: Announcements should be more visible

theFeu:
fixes #3560


lippserd:
We should find a better solution here. The yellow background gives you the feeling that this is a warning.

PR #3823, opened 2019-06-16
Order the programdata table by last status update

dasJ:
This ensures the master with the latest state is used instead of the
master that wrote its row first.


nilmerg:
Closing due to inactivity.

PR #3826, opened 2019-06-18
Create SECURITY.md

Crunsher:
Same text as proposed in the Icinga 2 core project.


lippserd:
This should be just a section in our README.md.

dnsmichi:
#3840 is similar to how I've done it in Icinga 2.

nilmerg:
Okay, then I guess this can be closed.

PR #3828, opened 2019-06-23
Add Ukrainian language support

p4k8:
Ukrainian language support for main application and modules doc, setup and monitoring.


nilmerg:
Woohoo :tada:

Just recently a Japanese translation and now an Ukrainian one. How did you come up with the idea of doing this? Just due to personal need or some sort of collaboration?  :thinking:  Just asking because this can't be just coincidence? :smile:

p4k8:
Haha, it is actually just a coincidence.
Sharing thing I did for my own environment due to personal need I guess :smile: . Would probably do director module translation later as well.

PR #3835, opened 2019-06-26
ui: Show problem badges in collapsed sidebar

### LINKED ISSUE #3568: Collapsed sidebar does not highlight problems
## Expected Behavior

I want to see that there are unhandled problems when collapsing the sidebar

## Current Behavior

Full-with sidebar shows System Issues (monitoring backend not running), module warnings (pending Director changes), unhandled issues (monitoring module). Collapsed sidebar does not provide this information.

## Possible Solution

Collapsed sidebar could either colorize the related icon with the corresponding problem-color or show a mini-badge in the upper/lower right corner, slightly overlapping the icon.

nilmerg:
fixes #3568

### DIFF
--- a/library/Icinga/Web/Navigation/Renderer/BadgeNavigationItemRenderer.php
+++ b/library/Icinga/Web/Navigation/Renderer/BadgeNavigationItemRenderer.php
@@ -113,6 +113,12 @@ abstract class BadgeNavigationItemRenderer extends NavigationItemRenderer
     protected function renderBadge()
     {
         if ($count = $this->getCount()) {
+            if ($count > 1000000) {
+                $count = round($count, -6) / 1000000 . 'M';
+            } elseif ($count > 1000) {
+                $count = round($count, -3) / 1000 . 'k';
+            }
+
             $view = $this->view();
             return sprintf(
                 '<span title="%s" class="badge state-%s">%s</span>',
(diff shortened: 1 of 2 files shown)


flourish86:
This would be the css for `text-overflow: ellipsis`;

```
#layout:not(.minimal-layout).sidebar-collapsed #menu .nav-level-1 > .nav-item.active > a > .badge {
    max-width: 4.5em;
    overflow: hidden;
    text-overflow: ellipsis;
}
```

nilmerg:
Looks fine to me on chromium Bildschirmfoto von 2019-07-03 07-44-01. Just with firefox it looks cropped somehow Bildschirmfoto von 2019-07-03 07-44-39. But we're seeing a million problems here. I guess if the number gets cropped in this case that's the least thing you're worried about. :thinking: :rofl:

flourish86:
> I guess if the number gets cropped in this case that's the least thing you're worried about. 🤔 🤣

Maybe. But the UI should be able to handle that anyway. Especially since the only thing it requires 3 LoC.

lippserd:
And we support high numbers of problems in the tactical overview as well.

nilmerg:
So, I think I've got us covered:  Bildschirmfoto von 2019-07-03 11-28-17 Bildschirmfoto von 2019-07-03 11-28-32 Bildschirmfoto von 2019-07-03 11-28-55 Bildschirmfoto von 2019-07-03 11-29-19

PR #3837, opened 2019-06-26
Don't translate pane and dashlet names in configs

### LINKED ISSUE #3542: Changes in dashboard.ini needed due to updated german localization
As there is a updated german localization fpr IcingaWeb2 2.6.0 existing dashboard.ini files for users need to be updated to use the new words.
The default dashboard "Aktuelle Incidents" was renamed to "Aktuelle Vorfälle".

I have fixed this for my using "vi" and ```:%s/Incidents/Vorfälle/g``` 

Maybe add a shot mention to the update docs?

nilmerg:
Introduces a major breaking change. All non-english users need to adjust their dashboard.ini after this change. (In case they've adjusted dashlets provided by modules)

Missing:
- [x] Upgrade documentation

fixes #3542

### DIFF
--- a/doc/80-Upgrading.md
+++ b/doc/80-Upgrading.md
@@ -14,6 +14,9 @@ v2.3 to v2.5 requires to follow the instructions for v2.4 too.
   for log messages emitted by jquery-migrate. (https://github.com/jquery/jquery-migrate) Your javascript code will still
   work, though jquery-migrate will notify you if you're utilizing deprecated/removed functions. jquery-migrate will be
   removed with Icinga Web v2.8 and code not adjusted accordingly will stop working.
+* If you're using a language other than english and you've adjusted or disabled module dashboards, you'll need to
+  update all of your `dashboard.ini` files. A CLI command exists to assist you with this task. Enable the `migrate`
+  module and run the following on the host where these files exist: `icingacli migrate dashboard sections --verbose`
 
 ## Upgrading to Icinga Web 2 2.6.x <a id="upgrading-to-2.6.x"></a>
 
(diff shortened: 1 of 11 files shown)


library/Icinga/Web/Widget/Dashboard.php
```diff
@@ -77,15 +77,15 @@ public function load()
         foreach ($navigation as $dashboardPane) {
             /** @var DashboardPane $dashboardPane */
             $pane = new Pane($dashboardPane->getLabel());
```
lippserd: My quick fix for this also included the following here:
```
$pane = (new Pane($dashboardPane->getName()))->setTitle($dashboardPane->getLabel());
```
Maybe this helps? With this change only, panes won't get translated upon saving.

nilmerg: Tried it, doesn't affect dashlets though. And honestly, my solution works. Maybe there's a simpler one, but any additional second used to look at this is a wasted one.

doc/80-Upgrading.md
```diff
@@ -6,6 +6,14 @@ Icinga Web 2 database.
 Specific version upgrades are described below. Please note that version updates are incremental. An upgrade from
 v2.3 to v2.5 requires to follow the instructions for v2.4 too.
 
+## Upgrading to Icinga Web 2 2.7.x <a id="upgrading-to-2.7.x"></a>
+
+**Breaking changes**
+
+* If you're using a language other than english and you've adjusted or disabled module dashboards, you'll need to
+  update all of your `dashboard.ini` files. Either replace all section names with their english equivalent or just
+  remove the sections (Though, only those for module dashboards!) and apply all changes again.
```
dnsmichi: Maybe we can add an example here - old vs new :)

lippserd:
We should provide a migrate command for this instead of having users to do this on their own.

PR #3838, opened 2019-06-27
Update php vendor libraries

### LINKED ISSUE #3785: Update vendor PHP libraries

nilmerg:
resolves #3785


nilmerg:
@lippserd Skipped dompdf due to https://github.com/Icinga/icingaweb2-module-pdfexport. Or should we update to v0.8.3 anyway? (Currently at v0.8.2, more or less)

lippserd:
I do think we should update anyway. Everyone without reporting may have no idea about the pdfexport module.

PR #3846, opened 2019-07-08
Add contact and contactgroup filters

### LINKED ISSUE #2708: Add contact and contactgroup filters (not for restrictions)
It would be great if you could filter for hosts/services which alert to a specific contactgroup. For example something like:

/icingaweb2/monitoring/list/services?service_hard_state=1&service_contactgroup=\*team1\*

lippserd:
fixes #2708

### DIFF
--- a/modules/monitoring/library/Monitoring/Backend/Ido/Query/HostgroupQuery.php
+++ b/modules/monitoring/library/Monitoring/Backend/Ido/Query/HostgroupQuery.php
@@ -29,6 +29,9 @@ class HostgroupQuery extends IdoQuery
         'contacts' => [
             'host_contact' => 'hco.name1'
         ],
+        'contactgroups' => [
+            'host_contactgroup' => 'hcgo.name1'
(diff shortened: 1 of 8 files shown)


lippserd:
Combined group membership filters are not yet implemented for contacts and contact groups.

lippserd:
Decided against supporting combined group membership filters for the moment. This would require changes to our contact and contactgroup queries for which we don't have the time.

lippserd:
Had a typo in the service_contactgroup column. Though it is intended that service_contact and service_contactgroup filters only work for services and service groups. I wanted to keep that as minimal as possible. Hosts and Host groups only support host_contact and hostgroup_contact.

nilmerg:
Choosing `Service Contactgroup` now fails with this error: `SQLSTATE[42S22]: Column not found: 1054 Unknown column 'scg.contact_object_id' in 'on clause'`

lippserd:
Done.

PR #3852, opened 2019-07-11
CLI: Automatically load enabled modules if not disabled

### LINKED ISSUE #3771: Load enabled modules in CLI context
Loading hook implementations in CLI context will only succeed for loaded modules. Since no other module is loaded than that of the command (if at all) most hook implementations are missing.

The same is true for all dependencies a module may have with other modules. Without loading these dependencies *manually* nothing works.

Load enabled modules in CLI context, by default, possibly with a switch to turn it off or enable it only for dependencies.

## Additional context
Loading only dependencies may not suffice. Take the audit module as an example, it provides a hook that may be necessary just because a core part of Web 2 logs something to the audit log.

lippserd:
fixes #3771

### DIFF
--- a/library/Icinga/Cli/Command.php
+++ b/library/Icinga/Cli/Command.php
@@ -3,14 +3,12 @@
 
 namespace Icinga\Cli;
 
+use Icinga\Application\ApplicationBootstrap as App;
+use Icinga\Application\Config;
 use Icinga\Application\Logger;
-use Icinga\Cli\Screen;
+use Icinga\Exception\IcingaException;
 use Icinga\Exception\NotReadableError;
 use Icinga\Util\Translator;
-use Icinga\Cli\Params;
-use Icinga\Application\Config;
-use Icinga\Application\ApplicationBootstrap as App;
-use Icinga\Exception\IcingaException;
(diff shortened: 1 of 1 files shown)


library/Icinga/Cli/Command.php
```diff
@@ -45,6 +48,9 @@ abstract class Command
 
     protected $defaultActionName = 'default';
 
+    /** @var string Module loading mode. Either {@link Command::LOAD_ENABLED} or false to disable module loading */
+    protected $loadModules = self::LOAD_ENABLED;
```
nilmerg: A use-case please, otherwise I'll demand this to be just `true` by default.

lippserd: We had a discussion about introducing something like `LOAD_DEPENDENCIES_ONLY` as well. So I started with the const `LOAD_ENABLED`. But I'm fine with only `true` for the moment.

nilmerg: Yes, though we ditched it because we have no proper mechanism to detect dependencies. And I suspect this will change in the foreseeable future..

PR #3858, opened 2019-07-11
Require time output for downtime schedule and end/expire times

### LINKED ISSUE #3478: Confusing UI for downtime events
The host downtime status thing is very confusing for me:

downtimes

If I mouseover on the word "Fixed" it will highlight the whole line "Fixed downtime by icingaadmin created on Jun 19 16:23 ǀ scheduled on Jun 25 expires on Jun 26". That's good so far. However the mouseover on the word "Fixed" doesn't give any hover but it gives only on the text  "Jun 19 16:23", "Jun 25" and "Jun 26". The UI doesn't give any indication that these dates would be somehow separate elements that can be hovered on and it took me a long time to figure out how to find out the exact time for "Jun 25" downtime.

## Possible Solution

Let's get rid of the text " created on Jun 19 16:23 ǀ " or maybe put in in the hover text and then display "Jun 25" and "Jun 26" in the full format, so the text would be like this:
`Fixed downtime by icingaadmin | scheduled on Jun 25 12:20 expires on Jun 26 19:00 `

The hover text could be:
`Downtime created on Jun 19 16:23`

## Your Environment

lippserd:
fixes #3478

### DIFF
--- a/modules/monitoring/application/views/scripts/show/components/downtime.phtml
+++ b/modules/monitoring/application/views/scripts/show/components/downtime.phtml
@@ -74,7 +74,7 @@ if (empty($object->comments) && ! $addLink) {
                     }
                     $state .= ' ' . sprintf(
                         $this->translate('expires %s', 'Last format parameter represents the downtime expire time'),
-                        $this->timeUntil($downtime->end, false, true)
+                        $this->timeUntil($downtime->is_flexible ? $downtime->scheduled_end : $downtime->end, false, true)
                     );
                 }
             }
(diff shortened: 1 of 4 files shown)


modules/monitoring/application/views/scripts/show/components/downtime.phtml
```diff
@@ -61,20 +61,12 @@ if (empty($object->comments) && ! $addLink) {
                         $this->timeUntil($downtime->is_flexible ? $downtime->scheduled_end : $downtime->end)
```
nilmerg: This still also needs the new parameter.

lippserd: Done.

PR #3860, opened 2019-07-12
Make overdue/unreachable checks better visible

### LINKED ISSUE #3212: Make overdue (or unreachable) checks better visible
Hi,

Could overdue or unreachable checks be shown with a lighter color or more 'greyish'? From other issues I understand that visualizing unreachable (where dependencies hit) is hard with the current IDO schema. (There are other issues for that already open, like #2232 )

But could at least every check, that was not checked in time be shown, no matter if overdue or unreachable be shown as outdated? I know there is the "last check result" and the dot besides check source but I think it would be very helpful to have this clearly visible from every overview, dashoard etc.. In other words I want to see it at one glance without having to open the detail view or the overdue dashboard.

What could be a bit tricky is the fact that not every check is executed at the exact time when it is scheduled. So maybe making it grey-ish after two times the regular check interval could be a good policy.

Cheers,
Thomas

nilmerg:
## Overdue Host
Bildschirmfoto von 2019-07-29 09-16-02

## Overdue Services
Bildschirmfoto von 2019-07-29 09-24-08

resolves #3212

### DIFF
--- a/modules/monitoring/application/controllers/HostController.php
+++ b/modules/monitoring/application/controllers/HostController.php
@@ -103,7 +103,8 @@ class HostController extends MonitoredObjectController
             'service_passive_checks_enabled',
             'current_check_attempt' => 'service_current_check_attempt',
             'max_check_attempts'    => 'service_max_check_attempts',
-            'service_check_command'
+            'service_check_command',
+            'service_next_update'
         ));
         $this->applyRestriction('monitoring/filter/objects', $query);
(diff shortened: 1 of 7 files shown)


nilmerg:
Included a check's latency into the calculation now. Also, the entire result is now doubled, not just `last_check + check_interval|retry_interval`.

Had very few late results with 1,5k checks and a check interval of 20s. (Load was over 100) None with the default check interval of 60s.

PR #3862, opened 2019-07-12
Provide ConfigFormEventsHook

### LINKED ISSUE #3768: Allow to hook into a configuration form's handling
Configuration forms currently do not allow to integrate any customized handling.

Developers of modules should be able to subscribe to multiple states a configuration form may produce.

* onSuccess (`onSuccess`)
* isValid (`onIsValid`)

This may be implemented by a new hook. (e.g. `FormEventHook` or `FormStateHook`)

Multiple hooks should be processed one-by-one while any of these can interrupt the sequence by throwing an exception.

lippserd:
fixes #3768

The `ConfigFormEventsHook` allows developers to hook into the handling of configuration forms. It provides three methods:

* `appliesTo()`
* `isValid()`
* `onSuccess()`

`appliesTo()` determines whether the hook should run for a given configuration form.
Developers should use `instanceof` checks in order to decide whether the hook should run or not.
If `appliesTo()` returns `false`, `isValid()` and `onSuccess()` won't get called for this hook.

`isValid()` is called after the configuration form has been validated successfully.
An exception thrown here indicates form errors and prevents the config from being stored.
The exception's error message is shown in the frontend automatically.
If there are multiple hooks indicating errors, every error will be displayed.

`onSuccess()` is called after the configuration has been stored successfully.
Form handling can't be interrupted here. Any exception will be caught, logged and notified.

Hook example:

```php
namespace Icinga\Module\Acme\ProvidedHook;

use Icinga\Application\Hook\ConfigFormEventsHook;
use Icinga\Forms\ConfigForm;
use Icinga\Forms\Security\RoleForm;

class ConfigFormEvents extends ConfigFormEventsHook
{
    public function appliesTo(ConfigForm $form)
    {
        return $form instanceof RoleForm;
    }

    public function onSuccess(ConfigForm $form)
    {
        $this->updateMyModuleConfig();
    }

    protected function updateMyModuleConfig()
    {
        // ...
    }
}
```


application/forms/ConfigForm.php
```diff
@@ -52,9 +53,19 @@ public function setIniConfig(Config $config)
...
 
-    /**
-     * {@inheritdoc}
-     */
+    public function isValid($formData)
+    {
+        $valid = parent::isValid($formData);
+
+        if ($valid && ConfigFormEventsHook::runIsValid($this) === false) {
+            $this->addErrors(ConfigFormEventsHook::getLastErrors());
```
nilmerg: Please use our form notifications here. They're used everywhere else and look better.

```suggestion
            foreach (ConfigFormEventsHook::getLastErrors() as $message) {
                $this->error($message);
            }
```

lippserd: Makes sense. Done.

PR #3867, opened 2019-07-16
Fix confusing downtime notification history

### LINKED ISSUE #2963: History shown for downtimes and notifications can be confusing
## Current Behavior
Sort order in the history view seems to be only based on the timestamp, so can be confusing if you get something like this:
grafik
User thought something is wrong because it displays notification during downtime.

## Expected Behavior
View should show entries more logical. In this case it would be "Downtime End" -> "Critical Event (Hard)" -> "Notification".

## Your Environment
* Icinga Web 2 version and modules (System - About): 2.4.1
* Version used (`icinga2 --version`): 2.6.3
* Operating System and version: RHEL7

lippserd:
fixes #2963 

@nilmerg This depends on the Markdown `<p>` in `<p>` rendering fix and should be reviewed once we have this.

### DIFF
--- a/modules/monitoring/application/views/scripts/partials/event-history.phtml
+++ b/modules/monitoring/application/views/scripts/partials/event-history.phtml
@@ -179,38 +179,47 @@ $rowAction = Url::fromPath('monitoring/event/show');
                 <div class="state-meta"><?= $this->formatTime($event->timestamp) ?></div>
             </td>
             <td>
-                <?php if ($this->isOverview): ?>
-                    <?= $this->qlink(
-                        $event->host_display_name,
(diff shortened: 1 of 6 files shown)


nilmerg:
We don't need to fix anything with the markdown rendering. We're rendering a bunch of words in block context, Parsedown will ever produce a paragraph here. The only solutions to solve this are:

a) Use `<section>` instead, `<div>` can't be used since we now allow everything in markdown including `<h1>`
b) Restore the differentiation between block level and phrasing content and only allow phrasing content in these views. (i.e. comments)

a) is c68abe7ff57656c329e1f4bec795788f1875a7b8. b) requires way more work.

PR #3868, opened 2019-07-16
Support subquery filters in filter chains

lippserd:
Before, combined membership filters were only possible in filter expressions, e.g. `hostgroup_name=(a&b)`.
Now, also `(hostgroup_name=a&hostgroup_name=b)` is supported (even w/o the parantheses) which makes it easier to build such filters with the filter editor.


lippserd:
ref/NC/615573

PR #3870, opened 2019-07-18
Collapsible plugin output

### LINKED ISSUE #3566: Collapsible plugin output
Make plugin return values collapsible im service view. 
Some returns are whole logs from the error with multiple sites of text. 

Maybe just preview the first 5 lines. To open the other lines, make a little "+" or "click here for more".

nilmerg:
The following details now collapse:

* Plugin output
* Plugin perfdata
* Custom variables
* Notes

Plugin output and perfdata are grouped by their check_command name. For example, expanding `disk`'s output on host `localhost` expands it for every other host which has this check.
All custom variables are grouped by their object's type. Once custom variables are expanded for a service every other service has its variables also expanded. The same applies for notes.

resolves #3566

### DIFF
--- a/modules/monitoring/application/views/helpers/Customvar.php
+++ b/modules/monitoring/application/views/helpers/Customvar.php
@@ -34,9 +34,11 @@ class Zend_View_Helper_Customvar extends Zend_View_Helper_Abstract
             return '[]';
         }
         $out = "<ul>\n";
+
         foreach ($array as $val) {
             $out .= '<li>' . $this->customvar($val) . "</li>\n";
         }
+
         return $out . "</ul>\n";
     }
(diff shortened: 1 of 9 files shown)


nilmerg:
@dnsmichi @Thomas-Gelf Custom variables also?

dnsmichi:
@nilmerg would be nice, yes. Some environments have a long list of CVs (or those which a JSON encoded and rendered as large blobs).

nilmerg:
Fixed bugs and applied the changes we discussed. Though, the flickering is still there as it can't be completely avoided.

lippserd:
I don't see flickering anymore 👍

PR #3878, opened 2019-07-26
Compat: Provide empty LESS mixin .code()

lippserd:
Some 3rd party modules (e.g. vsphere) use this mixin (though we never had it).
With lessphp version 0.5.0 using undefined mixins now throws exceptions.
Since we can't rely that users upgrade those modules first,
we provide the mixin for compatibility here.

### DIFF
--- a/public/css/icinga/compat.less
+++ b/public/css/icinga/compat.less
@@ -32,3 +32,9 @@ table.action {
 table.avp {
   .name-value-table()
 }
+
+.code() {
+  // Some 3rd party modules (e.g. vsphere) use this mixin (though we never had it)
+  // With lessphp version 0.5.0 using undefined mixins now throws exceptions
+  // Since we can't rely that users upgrade those modules first, we provide the mixin for compatibility here
+}


nilmerg:
Had a look at vsphere, but this module doesn't even have a `public` directory. Where is this mixin in use?

nilmerg:
And, we had it (b5468a122dbf63d24f9aa898e02c4d5b6539f734) and removed it a day later. (32006420c549f3dd02bfa6494cdab32c9d31590b) Though, I suspect that's not what's been used as mixin.

That brings me to my next question: Do we really want to provide every mixin that's undefined but in use by some module? (No matter who's at fault here, we because we removed it or the module dev who's used it without enough care?)

If so, please make sure to note every module down which is affected. Then fix them and remove this work-around with the next update. I don't want to carry such _stuff_ with me forever..

lippserd:
It's only about the `code` mixin. Sure, the modules [1] should be fixed so that we remove this with version 2.8.

[1]
* https://github.com/LordHepipud/icingaweb2-module-windows/commit/a5aee21f12c4f2d06c6294704c2546ad7a08319b
* https://github.com/Icinga/icingaweb2-module-vspheredb/blob/master/public/css/module.less#L137

PR #3880, opened 2019-07-29
Maintain sidebar state per tab/window

lippserd:
The sidebar state is now maintained per tab/window because users view
different things in different tabs/windows. E.g. users work with a
specific module constantly clicking through the menu while having
other tabs/windows with some important dashboards open.
When the sidebar is closed in one tab/widnow, it should not close
the sidebar in the other tabs/windows.

### DIFF
--- a/public/js/icinga/storage.js
+++ b/public/js/icinga/storage.js
@@ -171,6 +171,10 @@
          * @returns {void}
          */
         onChange: function(key, callback, context) {
+            if (this.backend !== window.localStorage) {
+                throw new Error('[Storage] Only the localStorage emits events');
+            }
+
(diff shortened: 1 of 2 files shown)


public/js/icinga/storage.js
```diff
@@ -6,6 +6,8 @@
 
     const KEY_TTL = 7776000000; // 90 days (90×24×60×60×1000)
 
+    var backend = window.localStorage;
```
nilmerg: Do this in the constructor. This way you'll change the backend for every single storage instance in a window. Not just for the sidebar.

lippserd: Makes sense!

lippserd:
Done.

PR #3887, opened 2019-08-01
Don't render wildcard expressions as sub-query filters

### LINKED ISSUE #3886: Wildcard filters in chains broken
Hi.

Updating 2.6 to 2.7 broke _monitoring/filter/objects_ using * (wildcard).

Example: I would like to match Host Group Names _"foo-set1"_ and _"foo-set2"_ so the filter is set as **"hostgroup_name=foo-*"**.

Thanks.

nilmerg:
fixes #3886


nilmerg:
Different fix required. Done by @lippserd.

Issue #3892, opened 2019-08-01
Theme solarized-dark on 2.7.0 has contrast issues / ambiguous controls


Diftraku:
## Describe the bug
After updating Icinga Web 2 to 2.7.0 via CentOS repos, I noticed couple issues with the choice of colours causing some issues with either text being invisible or control elements having ambiguous states due to their colours.

## To Reproduce
1. Update Icinga Web 2 to version 2.7.0
2. Browse to the login page (first issue)
3. Login and open any service with at least warning state, click acknowledge (first issue, textareas)
4. On the same page, further down on the form (second issue).
5. Navigate to Configuration -> Application (first issue, dropdowns)
6. On the same page, after the options (nitpick on the button colours)

## Expected behavior
Text should be clearly visible on the appropriate elements and the controls should not be ambiguous if they are toggled or not.

## Screenshots
First one is visible on the login page (the username field has `helloworld` written in it):
![image](https://user-images.githubusercontent.com/171226/62276983-e50fa500-b44d-11e9-814d-e1791e27be2e.png)
This is also present on textareas (the textarea has `Hello world!` written in it):
![image](https://user-images.githubusercontent.com/171226/62277147-3455d580-b44e-11e9-9cc4-197f6bc17058.png)
And on dropdowns:
![image](https://user-images.githubusercontent.com/171226/62277351-ab8b6980-b44e-11e9-96b9-51ee09d041b5.png)

The second issue relates to the toggles: which ones are enabled and which are disabled?
![image](https://user-images.githubusercontent.com/171226/62277190-4f284a00-b44e-11e9-8aad-c65755a4d94e.png)

As a last nitpick, the primary/secondary colours for the buttons could use some love:
![image](https://user-images.githubusercontent.com/171226/62277325-9a425d00-b44e-11e9-86bd-911ca00b0b8b.png)

## Your Environment
* Icinga Web 2 version and modules (System - About): 2.7.0
* Web browser used: Google Chrome v75.0.3770.142 (64-bit)
* Icinga 2 version used (`icinga2 --version`): r2.10.5-1
* PHP version used (`php --version`): 7.0.27
* Server operating system and version: CentOS Linux release 7.6.1810 (Core)

## Additional context
SELinux is not enabled on the server.

nilmerg:
Hi, thanks for the report!

We seem to have overlooked our own themes. :roll_eyes: 

@theFeu Please have a look at all the built-in themes and fix them if they show the same issue.

[...]

PR #3899, opened 2019-08-05
Bugfix/solarized theme contrast issues 3892

theFeu:
refs #3892

### DIFF
--- a/public/css/themes/solarized-dark.less
+++ b/public/css/themes/solarized-dark.less
@@ -211,6 +211,10 @@ textarea {
   text-decoration: underline;
 }
 
+table.avp tbody th {
+  background-color: transparent;
+}
+
 .icinga-module.module-monitoring {
   @timeline-notification-color: #1650CF;
   @timeline-hard-state-color: #A24600;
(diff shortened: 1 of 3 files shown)


public/css/themes/solarized-dark.less
```diff
@@ -266,3 +317,65 @@ textarea {
     color: @icinga-blue;
   }
 }
+
+/* Form styles */
+
+@input-background: lighten(@base02, 5);
+
+.icinga-form,
+.controls {
```
nilmerg: It's `.icinga-controls` and since you're only changing the style here `.icinga-form` is not required.

public/css/themes/solarized-dark.less
```diff
@@ -266,3 +317,65 @@ textarea {
...
+  & input[type="checkbox"]:focus + .toggle-switch .toggle-slider {
+    box-shadow: 0 0 0 2px @body-bg-color, 0 0 0 4px fade(@button-primary-color, 40);
+  }
+
+  & input[type="checkbox"]:checked + .toggle-switch .toggle-slider:before {
+    border: 1px solid @button-primary-color;
+  }
+}
+
+.button {
```
nilmerg: Does this collide with the mixin?

theFeu: It doesn't, it just overwrites the colours used

PR #3903, opened 2019-08-05
Support wildcard filters in chains

### LINKED ISSUE #3886: Wildcard filters in chains broken

lippserd:
fixes #3886

### DIFF
--- a/library/Icinga/Data/Db/DbQuery.php
+++ b/library/Icinga/Data/Db/DbQuery.php
@@ -298,13 +298,31 @@ class DbQuery extends SimpleQuery
         }
 
         if (is_array($expression)) {
+            $comp = [];
+            $pattern = [];
+            foreach ($expression as $value) {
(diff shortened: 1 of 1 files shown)


Thomas-Gelf:
Stumbled over this as it broke some configured restrictions over here. I can confirm that this helps with #3886. Rendered query looks slightly different than it used to. Example with `hostgroup_name=A*`:

### v2.6.3
```sql
WHERE (hgo.name1 LIKE 'A%')
```

### v2.7.0
```sql
WHERE (hgo.name1 IN ('A*'))
```

### master with this patch
```sql
WHERE ((hgo.name1 LIKE 'A%'))
```

PR #3904, opened 2019-08-06
Rework layout with flexbox

### LINKED ISSUE #3810: Avoid fixed positioning in main layout
## Problem
Fixed positioning requires a lot of js for layouting basic elements and therefore makes it unstable.

## Solution
Use flexbox for main layout

## Considerations
Different layout behavior for js and no-js, see WordPress Admin interface.

flourish86:
Flexbox layout is now implemented.

resolves #3810 

Some notes:
* It works fine in recent versions of Chrome, Safari and Firefox (I tested on macOS, we should test other OSes)
* It works surprisingly great in Edge
* It works in IE 11. Though there are some minor inconsistencies.
* Layout is completely broken in IE 10 and below. It might be possible to get it to work, but imho not with a reasonable amount of effort, since IE 10 basically supports flex-box but uses an older version with different syntax and required `-ms`-prefixes

### DIFF
--- a/application/layouts/scripts/body.phtml
+++ b/application/layouts/scripts/body.phtml
@@ -47,7 +47,7 @@ if ($this->layout()->inlineLayout) {
     </div>
     <?php endif ?>
     <div id="main" role="main">
-      <div id="col1" class="container<?= $moduleClass ?>"<?php if ($moduleName): ?> data-icinga-module="<?= $moduleName ?>" <?php endif ?> data-icinga-url="<?= $this->escape(Url::fromRequest()->without('renderLayout')->getAbsoluteUrl()); ?>"<?= $refresh; ?> style="display: block">
+      <div id="col1" class="container<?= $moduleClass ?>"<?php if ($moduleName): ?> data-icinga-module="<?= $moduleName ?>" <?php endif ?> data-icinga-url="<?= $this->escape(Url::fromRequest()->without('renderLayout')->getAbsoluteUrl()); ?>"<?= $refresh; ?>>
         <?= $this->render($inlineLayoutScript) ?>
       </div>
       <div id="col2" class="container"></div>
--- a/application/layouts/scripts/body.phtml
+++ b/application/layouts/scripts/body.phtml
@@ -46,21 +46,13 @@ if ($this->layout()->inlineLayout) {
       <?= $this->render('parts/navigation.phtml'); ?>
     </div>
     <?php endif ?>
-</div>
-<div id="main" role="main">
-  <div id="col1"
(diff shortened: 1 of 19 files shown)


theFeu:
So i clicked around in your branch a little and that's what I noticed:

----

It buggs me a little is that there is no longer a gap above the tabs like so:

image

image

----

Announcements and notifications are not centered properly:
image

image

application/layouts/scripts/body.phtml
```diff
@@ -23,36 +23,36 @@ if ($this->layout()->inlineLayout) {
...
+<div id="content-wrapper">
+    <?php if (! $this->layout()->isIframe): ?>
+    <div id="sidebar">
+        <div id="header">
+            <div id="header-logo-container">
+                <?= $this->qlink(
+                    '',
+                    Auth::getInstance()->isAuthenticated() ? 'dashboard' : '',
+                    null,
+                    array(
```
theFeu: We should us the new array syntax array() -> []

application/layouts/scripts/parts/navigation.phtml
```diff
@@ -19,17 +19,15 @@ if (! $this->auth()->isAuthenticated()) {
...
-                'menuRenderer' => (new Menu())->getRenderer()->setUseStandardItemRenderer()
-            )
-        ) ?>
-    </div>
+<div  id="menu" data-last-update="-1" data-base-target="_main" class="container"
+      data-icinga-url="<?= $this->href('layout/menu') ?>"  data-icinga-refresh="15">
+    <?= $this->partial(
+        'layout/menu.phtml',
+        'default',
+        array(
```
theFeu: array syntax again :)

application/layouts/scripts/body.phtml
```diff
@@ -66,3 +66,4 @@ if ($this->layout()->inlineLayout) {
   ?></ul>
   <div id="application-state-summary" class="container" data-icinga-url="<?= $this->url('application-state/summary') ?>" data-last-update="-1" data-icinga-refresh="60"></div>
 </div>
+
```
theFeu: Superfluous newline

public/js/icinga/ui.js
```diff
@@ -591,12 +591,12 @@
...
                     });
                 }
-            });
+            });*/
 
             this.fixControls($container);
         },
 
-        fixControls: function($container) {
+        fixControls2: function($container) {
```
theFeu: What's up with the fixControls2?
It's not being used anywhere -> could it just be hidden behind a comment if it is unused?

PR #3926, opened 2019-08-16
Add appropriate aria labels to tactical overview

theFeu:


theFeu:
> Why is this necessary? Does a screen reader really not read header tag contents?

The content of it is only "labelBig".

So for

20
hosts down

it will announce "20" instead of "20 hosts down"

theFeu:
> Why is this necessary? Does a screen reader really not read header tag contents?

in case you mean:
<h2 aria-label="<?= $this->translate('Host Summary') ?>"><?= $this->translate('Host Summary') ?></h2>

Most of the ones I tested actually don't.

PR #3934, opened 2019-08-28
Bugfix/duplicate ldap uri

mxhash:
Hi,

users use the URI in LDAP resources. The first commit is just to see the normalised hostname. The second just a try fix - even if it is not the best one. Just to hint issues on that.

Cheers,
Marius


nilmerg:
Closed due to inactivity.

Issue #3907, opened 2019-08-08
Indicate empty values in detail view


flourish86:
## Description
Empty values should be indicated as "empty". Just leaving it out, looks like something went wrong.

## To Reproduce
E.g.: Navigate to Detail view of an object with empty plugin output. The heading 'Plugin output' is displayed at the top of the content area, followed immediately by the next heading.

## Expected behavior
The user should be noticed that there's no plugin output.

## Screenshots

Screenshots taken in high-contrast theme.

![Artboard Copy](https://user-images.githubusercontent.com/6977434/62690931-ee5cbc80-b9cd-11e9-9819-b268cab257a5.jpg)

![Artboard Copy](https://user-images.githubusercontent.com/6977434/62690937-f157ad00-b9cd-11e9-9001-c5310aab06ee.jpg)

## Your Environment
Icinga Web 2.7.0

nilmerg:
That this is not an error is really an edge-case. Usually a check returns output. If no output is shown it's 99% of the time an issue with how the check has been executed. (Either by Icinga or a wrapper script)

Hence, I'm not bothered at all that it implies something is wrong the way it looks currently. Something **is** wrong after all. (In your example anyway, since the checks are pending.)

flourish86:
> Hence, I'm not bothered at all that it implies something is wrong the way it looks currently. Something is wrong after all. (In your example anyway, since the checks are pending.)

Yes, but having an error should not be communicated through UI being buggy. Even if it's an edge-case, we should have this solved. UI should be clear and transparent especially in edge-cases.

nilmerg:
The references PR is closed, so I'll close here as well. Icinga DB Web handles this fine.

PR #3938, opened 2019-09-02
Bugfix/indicate empty values in detail view 3907

flourish86:
Here's my draft for #3907 .

I opted for having the empty state in list view as well. Since it makes it clearer, that there's no output.

I see the argument of reduced information density in edge cases. But it also helps to make objects stand out better if there's an additional lack of plugin output. We should chose better over more.


nilmerg:
Sorry, but this completely ignores history entries and besides that requires some changes anyway. I don't see a point investing more time into this here since users lived with this straight from the start.

We should instead ensure this won't be the case in Icinga DB Web.

Which it is: Bildschirmfoto von 2020-04-09 13-43-10

Closing. Issue for Icinga DB Web: https://github.com/Icinga/icingadb-web/issues/48

PR #3943, opened 2019-09-04
Improve LDAP/ActiveDirectory authentication

lazyfrosch:
This PR bundles several improvements for LDAP and ActiveDirectory

## Changes

- [x] Implement Interface and Trait for external auth
      So one can implement it in other backends - so the backend can verify the externally set user
- [x] Implement external auth for any LDAP backend
      A user will only be logged in if the account exists in LDAP
- [x] Allow alternative attributes (multiple) in LDAP to lookup the user
      `userPrincipalName` is look upped in addition to `sAMAccountName` to find a user, but the user attribute is used then to identify the user in Icinga Web 2
- [x] Allow multiple domains for the (LDAP) backends
      So LDAP Forest, Kerberos Realm and possible E-Mails can be matched
      Also detect LDAP Naming Context **and** NETBIOS name by default
- [ ] Update authentication documentation, especially with examples
- [ ] Review Auth forms and update translation

## Target Audience

* Multi-Domain authentication
* Kerberos users (on webserver in front of Icinga Web 2)
* LDAP or AD where sAMAccountName and LDAP Domain is not used to build the Kerberos principal
  Example (which is valid in ActiveDirectory):

## Examples

Usernames vs. Kerberos principal must not match, maybe an e-mail address could also be used.

```
LDAP Domain:       ad.company.net
sAMAccountName:    ABCDEF
userPrincipalName: john.doe@company.com
mail:              john@company.biz
```

## Benefits

* External Auth can depend on a user backend, which may verify the user exists or is in a required group
* External Auth can now fail back to a login form to ask for other credentials
* Other names can be assigned to a user, e.g. e-mail, Kerberos principal or alternative user names

## Tests

TBD

ref/NC/628947

### DIFF
--- a/application/forms/Config/UserBackend/LdapBackendForm.php
+++ b/application/forms/Config/UserBackend/LdapBackendForm.php
@@ -262,6 +262,7 @@ class LdapBackendForm extends Form
                     . ' authenticating the user based on the username without the domain part.'
                     . ' If your LDAP backend holds usernames with a domain part or if it is not necessary in your setup'
                     . ' to authenticate users based on their domains, leave this field empty.'
+                    . ' (Can be a comma separated list)'
                 ),
                 'preserveDefault' => true,
                 'value'         => $this->getSuggestion('domain')
@@ -378,14 +379,16 @@ class LdapBackendForm extends Form
     {
         $cap = $this->getLdapCapabilities($formData);
 
+        $domain = $this->defaultNamingContextToFQDN($cap);
+
         if ($cap->isActiveDirectory()) {
             $netBiosName = $cap->getNetBiosName();
             if ($netBiosName !== null) {
-                return $netBiosName;
+                $domain .= ',' . $netBiosName;
             }
         }
 
-        return $this->defaultNamingContextToFQDN($cap);
+        return $domain;
     }
 
     /**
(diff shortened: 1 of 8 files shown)


lazyfrosch:
This is far from finished, but it makes sense to me to implement this in for broad usage.

Comments and thoughts are welcome, especially from @lippserd and @dgoetz

dgoetz:
Having configuration frontend and more flexibility for external auth would be great so :+1: 

For environment variables I have also seen `X_REMOTE_USER` sometimes, so not sure if we should add this to default list or have the list configurable?

I would say Kerberos from AD and FreeIPA, faked external auth via Apache rules and if we or another partner have some big customer with something special (like CAS) we should test this as platform. In the tests we should test access to Director API including self-service endpoints to ensure nothing breaks.

lazyfrosch:
The basic functionality is now implemented.

Still pending are (automated) tests, docs and examples.

lazyfrosch:
It seems like my obervations regaring userPrincipalName where not correct, the main Kerberos principal is still `samaccountname@domainfqdn`.

The `userPrincipalName` seems to be used as alternative, similar to an login via e-mail.

So the changes still make sense to pursue for more flexibility, will update the issue soon.

lazyfrosch:
My use-case no longer exists for this change.

PR #3956, opened 2019-09-19
Fix checks being marked as late when they are scheduled for later time

### LINKED ISSUE #3696: Checks that are only checked at specific timeslots are marked as late results
## Expected Behavior
A service that is only checked eg. between 16:00 and 18:00 should set its *Next check* variable to 16:00 and should not appear in the "late service check results"-dashlet.

## Current Behavior
The *Next check* is showing (in our case) one minute after the last check even though it is out of the checking period. (The example below was changed from 06:00-08:00 to 16:00-18:00, thats why the last check was at 07:59)
image

Also the check is listed in the "late service check results"-dashlet.
image

## Steps to Reproduce
1. Create a timeperiod
2. Assign it to a service check
3. Wait until the checking period is over

## Context
It is not that annoying because we barely never use the Overdue dashboard but it is a bit confusing when you check one of your services and it tells you that it should have been checked hours ago (even though it shouldn't).

## Your Environment
* Icinga Web 2 version and modules (System - About): 2.6.2, using Director -> current master
* Version used (`icinga2 --version`): r2.10.2-1
* Operating System and version: CentOS 7

theFeu:
fixes #3696

### DIFF
--- a/modules/monitoring/library/Monitoring/Backend/Ido/Query/HoststatusQuery.php
+++ b/modules/monitoring/library/Monitoring/Backend/Ido/Query/HoststatusQuery.php
@@ -101,17 +101,17 @@ class HoststatusQuery extends IdoQuery
             'host_next_notification'                => 'UNIX_TIMESTAMP(hs.next_notification)',
             'host_next_update'                      => 'CASE WHEN hs.has_been_checked = 0 OR hs.has_been_checked IS NULL
             THEN
-                NULL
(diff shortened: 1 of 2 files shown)


nilmerg:
Thanks for finding this! Though, since we're utilizing this also in other places I've adjusted the queries to consider `next_check` and to use the same algorithm as Icinga 2 does now. https://github.com/Icinga/icinga2/pull/7534

PR #3959, opened 2019-09-23
Asynchronous Content Updates

### LINKED ISSUE #3953: Asynchronous Content Updates
Forms (notably the filter editor) may want to respond to their submission just with an update
for themselves. This provides instant feedback to the client's interaction with the form. Though,
other containers may also require an update due to the interaction.

```
Valid Form Submission -> HTTP Redirect -> Form Update
                                       \
                                        \ -> Content Update
```

This requires a new HTTP response header which instructs our JS to also load a given url to a
particular container:
```http request
X-Icinga-Extra-Updates: <id>;<url>[, <id>;<url>]
```

However, since widgets depend upon another (e.g. pagination and list-content) the response to
such an extra update must be able to provide updates for multiple containers.

Responses therefore must be able to provide structured content:
```http request
X-Icinga-Multipart-Content: <sep>

for=<id>
<html>
<sep>
for=<id>
<html>
```

More is not required, yet. One could think of an additional request header which differentiates
an extra-update from an auto-refresh. Though I have no real use-case why we should need this.

nilmerg:
resolves #3953

### DIFF
--- a/public/js/icinga/loader.js
+++ b/public/js/icinga/loader.js
@@ -639,6 +639,26 @@
             delete this.requests[req.$target.attr('id')];
             this.icinga.ui.fadeNotificationsAway();
 
+            var extraUpdates = req.getResponseHeader('X-Icinga-Extra-Updates');
+            if (!! extraUpdates && req.getResponseHeader('X-Icinga-Redirect-Http') !== 'yes') {
+                var _this = this;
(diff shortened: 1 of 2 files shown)


nilmerg:
Must be rebased once `feature/shared-assets` is merged.

PR #3970, opened 2019-10-02
Refresh container if X-Icinga-Extra-Updates with length 1 is defined

ValeDaRold:
Enable the possibility to refresh the parent container of the given element passed through X-Icinga-Extra-Updates http-header.

### DIFF
--- a/public/js/icinga/loader.js
+++ b/public/js/icinga/loader.js
@@ -647,19 +647,23 @@
             if (!! extraUpdates && req.getResponseHeader('X-Icinga-Redirect-Http') !== 'yes') {
                 var _this = this;
                 $.each(extraUpdates.split(','), function (idx, el) {
-                    var parts = el.trim().split(';');
-                    if (parts.length !== 2) {
+                    var parts = el.trim().split(';'), $target, url;
+                    if (parts.length === 2) {
+                        $target = $('#' + parts[0]);
+                        if (! $target.length) {
+                            _this.icinga.logger.warn('Invalid target ID. Cannot load extra URL', el);
+                            return;
+                        }
+                        url = parts[1];
+                    } else if (parts.length === 1) {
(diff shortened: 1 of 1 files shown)


public/js/icinga/loader.js
```diff
@@ -647,19 +647,23 @@
             if (!! extraUpdates && req.getResponseHeader('X-Icinga-Redirect-Http') !== 'yes') {
                 var _this = this;
                 $.each(extraUpdates.split(','), function (idx, el) {
-                    var parts = el.trim().split(';');
-                    if (parts.length !== 2) {
+                    var parts = el.trim().split(';'), $target, url;
```
nilmerg: Please don't initialize variables on a single line. Use a separate line and `var` for each one.

public/js/icinga/loader.js
```diff
@@ -647,19 +647,23 @@
...
+                    if (parts.length === 2) {
+                        $target = $('#' + parts[0]);
+                        if (! $target.length) {
+                            _this.icinga.logger.warn('Invalid target ID. Cannot load extra URL', el);
+                            return;
+                        }
+                        url = parts[1];
```
nilmerg: Please leave an empty line after braces.

public/js/icinga/loader.js
```diff
@@ -647,19 +647,23 @@
...
+                    } else if (parts.length === 1) {
+                        $target = $(parts[0]).closest(".container").not(req.$target);
+                        url = $target.data('icingaUrl');
```
nilmerg: This will result in an error if the element is missing or a part of `req.$target`. Please check if you've got a result before accessing `$target`. And please log a separate message in case the result is `req.$target`, this way it's clear that this is prohibited and the developer doesn't get a wrong warning about a missing element.

PR #3971, opened 2019-10-02
Add EventDetailsExtensionHook and include it into EventController

ValeDaRold:
Add the possibility to insert custom HTML in EventController, right after the Details table. The Hooks must implement a function getHtmlForEvent that should return an HTML string, that will be directly rendered in Show.phtml.

### DIFF
--- a/modules/monitoring/application/controllers/EventController.php
+++ b/modules/monitoring/application/controllers/EventController.php
@@ -5,6 +5,8 @@ namespace Icinga\Module\Monitoring\Controllers;
 
 use DateTime;
 use DateTimeZone;
+use Icinga\Module\Monitoring\Hook\EventDetailsExtensionHook;
+use Icinga\Application\Hook;
 use InvalidArgumentException;
 use Icinga\Data\Queryable;
 use Icinga\Date\DateFormatter;
@@ -69,6 +71,24 @@ class EventController extends Controller
             array(array($this->view->escape($this->translate('Type')), $label)),
             $this->getDetails($type, $event)
         );
+
+        $this->view->extensionsHtml = array();
+        /** @var EventDetailsExtensionHook $hook */
+        foreach (Hook::all('Monitoring\\EventDetailsExtension') as $hook) {
+            try {
+                $html = $hook->getHtmlForEvent($object);
+            } catch (\Exception $e) {
+                $html = $this->view->escape($e->getMessage());
+            }
+
+            if ($html) {
+                $module = $this->view->escape($hook->getModule()->getName());
+                $this->view->extensionsHtml[] =
+                    '<div class="icinga-module module-' . $module . '" data-icinga-module="' . $module . '">'
+                    . $html
+                    . '</div>';
+            }
+        }
 
         $this->view->title = $this->translate('Event Overview');
         $this->getTabs()
--- a/modules/monitoring/application/views/scripts/event/show.phtml
+++ b/modules/monitoring/application/views/scripts/event/show.phtml
@@ -17,6 +17,12 @@ echo $object instanceof Service
 ?>
 </div>
 <div class="content">
+    <?php
+    foreach ($extensionsHtml as $extensionHtml) {
+        echo $extensionHtml;
+    }
+    ?>
+
     <h2><?= $this->escape($this->translate('Event Details')) ?></h2>
     <table class="event-details name-value-table" data-base-target="_next">
     <?php
--- a/modules/monitoring/library/Monitoring/Hook/EventDetailsExtensionHook.php
+++ b/modules/monitoring/library/Monitoring/Hook/EventDetailsExtensionHook.php
@@ -1,0 +1,79 @@
+<?php
+/* Icinga Web 2 | (c) 2017 Icinga Development Team | GPLv2+ */
+
+namespace Icinga\Module\Monitoring\Hook;
+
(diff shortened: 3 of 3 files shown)


modules/monitoring/application/controllers/EventController.php
```diff
@@ -5,6 +5,8 @@
 
 use DateTime;
 use DateTimeZone;
+use Icinga\Module\Monitoring\Hook\EventDetailsExtensionHook;
+use Icinga\Web\Hook;
```
nilmerg: This class is deprecated. Use `Icinga\Application\Hook` instead.

modules/monitoring/application/controllers/EventController.php
```diff
@@ -70,6 +72,14 @@ public function showAction()
             $this->getDetails($type, $event)
         );
 
+        $hookHtml = '';
+        /** @var EventDetailsExtensionHook $hook */
+        foreach (Hook::all('Monitoring\\EventDetailsExtension') as $hook) {
+            $hookHtml .= $hook->getHtmlForEvent($event);
+        }
+
+        $this->view->hookHtml = $hookHtml;
```
nilmerg: Please do this the same way as it's done here:
https://github.com/Icinga/icingaweb2/blob/201cfa2330b826f862aa0dff23ebd552fa2fc38e/modules/monitoring/library/Monitoring/Web/Controller/MonitoredObjectController.php#L85-L102

i.e. Error handling, use a module container (otherwise a module's css is not applied)

modules/monitoring/application/views/scripts/event/show.phtml
```diff
@@ -25,4 +25,5 @@ echo $object instanceof Service
     }
     ?>
     </table>
+    <?= $hookHtml ?>
```
nilmerg: The host/service detailview extensions are rendered [before any standard html](https://github.com/Icinga/icingaweb2/blob/master/modules/monitoring/application/views/scripts/partials/object/detail-content.phtml#L4). I'd like to see this here as well. Does your use-case require this to be at the very end?

modules/monitoring/library/Monitoring/Hook/EventDetailsExtensionHook.php
```diff
@@ -0,0 +1,21 @@
...
+/* Icinga Web 2 | (c) 2017 Icinga Development Team | GPLv2+ */
+
+namespace Icinga\Module\Monitoring\Hook;
+
+/**
+ * Base class for hooks extending the event view of monitored objects
+ *
+ * Extend this class if you want to extend the event view of monitored objects with custom HTML.
+ */
+abstract class EventDetailsExtensionHook
```
nilmerg: Please teach this to know about a view, it's module and a final public constructor. Just like the [DetailviewExtensionHook](https://github.com/Icinga/icingaweb2/blob/master/modules/monitoring/library/Monitoring/Hook/DetailviewExtensionHook.php).

nilmerg: Eh, forget this about the view. That's not necessary.

modules/monitoring/application/views/scripts/event/show.phtml
ValeDaRold: No, our use case does not strictly require to be printed at the end. I will move it right before the "Event Details" table.

modules/monitoring/application/controllers/EventController.php
```diff
@@ -70,6 +72,24 @@ public function showAction()
             $this->getDetails($type, $event)
         );
 
+        $this->view->extensionsHtml = array();
+        /** @var EventDetailsExtensionHook $hook */
+        foreach (Hook::all('Monitoring\\EventDetailsExtension') as $hook) {
+            try {
+                $html = $hook->setView($this->view)->getHtmlForObject($this->object);
```
nilmerg: Did you test this?

modules/monitoring/application/views/scripts/event/show.phtml
```diff
@@ -15,6 +15,12 @@ echo $object instanceof Service
     ? '<h2>' . $this->translate('Current Service State') . '</h2>' . $this->render('partials/object/service-header.phtml')
     : '<h2>' . $this->translate('Current Host State') . '</h2>' . $this->render('partials/object/host-header.phtml');
 ?>
+
+<?php
+foreach ($extensionsHtml as $extensionHtml) {
+    echo $extensionHtml;
+}
+?>
```
nilmerg: It's fine that you've moved this upwards but now it's part of the controls section. Didn't you mention to place it right before the "Event Details" table? Please do so :wink:

modules/monitoring/library/Monitoring/Hook/EventDetailsExtensionHook.php
nilmerg: Why is this resolved? I wrote passing the view is not necessary but the other two are still necessary. The module object may be useful for implementations which require configuration. The constructor should be implemented and final so that we can guarantee that any implementation can be successfully instantiated. (If an implementation needs initialization, it should override method `init()`)

nilmerg:
Please rebase with our master and squash all of your latest commits into the first one. Then I'll merge this. (Would have done this myself but I have no permission on your side anymore.)

PR #3993, opened 2019-11-06
Introduce option to hide specific customvars in web

nbuchwitz:
This PR introduces a new configuration option and filter for hiding specific customvars in the detail view of a host or service. It uses the same matching syntax as the existing blacklist feature.

Should be useful in environments which make heavy use of apply rules on customvars, which are often cryptic and bloat the customvar section in web.

### DIFF
--- a/modules/monitoring/application/forms/Config/SecurityConfigForm.php
+++ b/modules/monitoring/application/forms/Config/SecurityConfigForm.php
@@ -61,6 +61,21 @@ class SecurityConfigForm extends ConfigForm
                 )
             )
         );
+
+        $this->addElement(
+            'text',
+            'hidden_customvars',
+            array(
+                'allowEmpty'    => true,
+                'attribs'       => array('placeholder' => $this->getDefaultProtectedCustomvars()),
+                'label'         => $this->translate('Hidden Custom Variables'),
+                'description'   => $this->translate(
+                    'Comma separated case insensitive list of hidden custom variables.'
+                    . ' Use * as a placeholder for zero or more wildcard characters.'
+                    . ' Existence of those custom variables will not be shown, but remain usable for modules.'
+                )
+            )
+        );
     }
 
     /**
--- a/modules/monitoring/library/Monitoring/Object/MonitoredObject.php
+++ b/modules/monitoring/library/Monitoring/Object/MonitoredObject.php
@@ -419,7 +419,9 @@ abstract class MonitoredObject implements Filterable
     public function fetchCustomvars()
     {
         $blacklist = array();
+        $hidden = array();
         $blacklistPattern = '';
+        $hiddenPattern = '';
 
         if (($blacklistConfig = Config::module('monitoring')->get('security', 'protected_customvars', '')) !== '') {
             foreach (explode(',', $blacklistConfig) as $customvar) {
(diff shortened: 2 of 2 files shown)


nbuchwitz:
CC @Mikesch-mp

nilmerg:
Sorry @nbuchwitz, but I'll revert this from master. There is already such a feature, but even I overlooked it since it's rather hidden and probably rarely used/asked.

The monitoring module provides a restriction `monitoring/blacklist/properties` which, when configured with a value like `host.vars.lorem,service.vars.ips*`, does the exact same as this PR.

nbuchwitz:
If I'm not wrong the blacklist feature removes the custom vars from the scope of the modules. So a blacklisted var can't used for restrictions or modules like my map module

nilmerg:
Object restrictions are not affected by this at all. But sure, anyone using `MonitoredObject::fetchCustomvars()` ([like you](https://github.com/nbuchwitz/icingaweb2-module-map/search?q=customvars&unscoped_q=customvars) and [we](https://github.com/Icinga/icingaweb2/search?l=HTML%2BPHP&q=-%3Ecustomvars) do) is affected. But this PR doesn't make a difference here. It's also overriding [`$this->customvars`](https://github.com/Icinga/icingaweb2/pull/3993/files#diff-fd6e585a47c65543d60ce1b2ea225e1fR464) just like the [blacklist feature](https://github.com/Icinga/icingaweb2/blob/2829837191199afb4bb796f7202c17d5feab5f0e/modules/monitoring/library/Monitoring/Object/MonitoredObject.php#L511).

Anyhow as a module accessing the IDO (or in the future the Icinga DB) it needs to implement its own authorization checks. That's already the case for object restrictions and by using `MonitoredObject::$hostVariables` or `MonitoredObject::$serviceVariables` it's possible to bypass the monitoring module's authorization checks to do this also for the restriction `monitoring/blacklist/properties`.

nbuchwitz:
Sure, but the blacklist is re-adding the customvars with ``***`` as value, which tends to bloat the detail view. My PR just removes the customvars from the view, which is often the ihmo better solution as the user usually don't wanna see endless lists of ``***`` values?

nbuchwitz:
Maybe @Mikesch-mp can give some real world examples as the idea came up when we talked at the OSMC.

nilmerg:
I'm not talking about protected customvars, that's what is adding the asterisks. Take a look into a role of yours, there's a restriction `monitoring/blacklist/properties` which is what I'm talking about.

Mikesch-mp:
Simple example, think about a multi customer frontend with dashboards provided and not editable by customer which includes cube module. Blacklisted variables will block such dashboards using blacklisted variables in the cube module. Its not about to fully secure block access to such variables, just some people/customers dont need to see them everywhere.

nilmerg:
Then this PR does **not** accomplish this. This will **not** only hide variables from the UI, but also from any code accessing them, just like the blacklist does.

nilmerg:
To clarify this once more. I'm not against a way to only hide customvars from the UI. Not at all. I'm only telling you that this PR does not achieve this and is therefore not acceptable as it does the same as the blacklist feature.

Though, instead of opening a new PR which achieves the desired effect, please create an issue first with a proposal how this should be done. Because I'm sure this will also be necessary in Icinga DB Web and there we may have more options at hand.

Issue #3286, opened 2018-01-18
Allow to forbid password changes on specific user accounts


AhmedBenmoussa:
Hello everyone, 

I installed **_icinga2_** as our monitoring tool in my company. My configuration consists of some users with different permissions. 
I want to give all the users the ability to change their passwords. How do I add this permission without giving them access to all the configuration menu. Cauz in the permission list there is only (_config/*_). I also read in the [Documentation ](https://www.icinga.com/docs/icingaweb2/latest/doc/06-Security/) that there is also _(config/modules)_. 
Is there any other sub categories ? if yes, what are the children of the _config_ category ?

I'm using 
- **icinga2** _v2.4.1_
- **icingaweb2** _v2.1.0_

Thank you in advance. 
With all my respects, Ahmed.

nilmerg:
Hi Ahmed,

I highly recommend you to consider an upgrade of Icinga Web 2. The latest version is v2.5 and includes the feature you're asking for without the need to change anything. (There users have the possibility to change their password in the preferences.)

Best regards,
Johannes

dnsmichi:
The original question was to add a specific permission to allow/deny password changes for users.

https://monitoring-portal.org/t/separating-monitored-hosts-by-account-type/636/5?u=dnsmichi

Al2Klimov:
https://github.com/Icinga/icingaweb2/issues/3286#issuecomment-358676734

dnsmichi:
@Al2Klimov this is a feature request to add such a fine granular permission. I've talked to @lippserd and it does not yet exist. So why did you close this issue?

Al2Klimov:
No, it's definitively not:

> I want to give **all the users** the ability to change their passwords.

– https://github.com/Icinga/icingaweb2/issues/3286#issue-289662003

This is a feature request to let **all** users change their passwords – and this is already possible:

> The latest version is v2.5 and includes the feature you're asking for **without the need to change anything**.

– https://github.com/Icinga/icingaweb2/issues/3286#issuecomment-358676734

dnsmichi:
This issue is coming from the linked forum post which applies roles to users. The question was if one could give users the permission to change their password, or to explicitly forbid it.

I'm aware of the feature to globally enable/disable that. 

@AhmedBenmoussa 
`all the users` in a global fashion, or `all the users assigned to specific role`?

AhmedBenmoussa:
Hello everyone, 

[...]

PR #3997, opened 2019-11-14
Added Kiosk Mode in doc

Wintermute2k6:
Added Kiosk Mode description, based on customer wish ref/NC/19457 & ref/NC/19599

### DIFF
--- a/doc/20-Advanced-Topics.md
+++ b/doc/20-Advanced-Topics.md
@@ -517,23 +517,16 @@ Have a look [here](20-Advanced-Topics.md#web-setup-manual-from-source-config) fo
 ## Kiosk Mode Configuration <a id="kiosk-mode"></a>
 
 Be aware that when you create a kiosk user every person who has access to the kiosk is able to perform tasks on it.
-For a kiosk mode you have to enable/create a dedicated user for it in the GUI. First add a new user named `kiosk`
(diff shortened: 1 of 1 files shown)


Wintermute2k6:
false reference should be ref/IP/19457 & ref/IP/19599

nilmerg:
Isn't this more of a how-to which should be published on [community.icinga.com/c/howto](https://community.icinga.com/c/howto)?

dnsmichi:
@nilmerg I would prefer this being in the official docs, this question gets asked more than anything else.

Wintermute2k6:
It was requested by our partners and it doesn't provide a negative effect and also frees up more time which can be spend with actual projects.

nilmerg:
Just had a first proper look at it. Since this is related to #3286 I'd add a paragraph about it here as well. Though, this is also talking about a default autologin account, which wouldn't need a database user or even usergroup. Or is this on purpose mentioned?

Wintermute2k6:
I think this would be very helpful ! Question would be how configureable this kiosk user would be just filters or would there be the possibility to separately adjust him in the roles ?

Thanks in advance !

nilmerg:
An autologin user can be configured just the same as every other user. A role just needs to define that it applies to the user's name. (i.e. `kiosk`) Then no database is required.

PR #4039, opened 2019-12-22
Add error message to webrouter.php for custom setups

Fonata:
This pull requests adds a message for broken HTTP configurations to save administrators time. In a typical configuration, the new message will be added to the error log of the HTTP server, for example Apache or NGINX.

I lost a few hours trying to figure out the Apache2 configuration that Icinga Web 2 expects.

### DIFF
--- a/library/Icinga/Application/webrouter.php
+++ b/library/Icinga/Application/webrouter.php
@@ -15,7 +15,7 @@ if (isset($_SERVER['REQUEST_URI'])) {
     return false;
 }
 
-// Workaround, PHPs internal Webserver seems to mess up SCRIPT_FILENAME
+// Workaround, PHPs built-in web server seems to mess up SCRIPT_FILENAME
 // as it prefixes it's absolute path with DOCUMENT_ROOT
 if (preg_match('/^PHP .* Development Server/', $_SERVER['SERVER_SOFTWARE'])) {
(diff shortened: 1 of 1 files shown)


nilmerg:
How is this caused? Please provide an example configuration which leads to this error.

I'm also not sure whether it's a good idea to trigger an error here. May it be intentional?

Fonata:
My faulty assumption was that it would not matter what the name of the code directory is. I cloned this repository somewhere and configured Apache's `DocumentRoot` directive to point to the dir. From the HTTP URL perspective, there was no `icingaweb2` path component, which lead to a content-less white page.

A correct way to configure Apache would have been to set the `DocumentRoot` to the directory *containing* the `icingaweb2` dir.

nilmerg:
Thanks for the feedback. But honestly, I'm not sure what the implications of this change are. Returning `false` and producing no output usually causes the server to fall back to a different method of routing the request. We've got several cases in this exact script where this is done so maybe this is no different.

Your assumption was correct. It really doesn't matter what the directory is. As long as the webserver configuration does not depend on it. Which it obviously does in your case. Anyhow, I think you're problem is solved for now.

Due to this and the to me unknown implications I don't see any urgency of making it more obvious that it's a server configuration error. (Especially with this rather less-useful error message :wink: )

PR #4040, opened 2019-12-31
Fixed %s database name in DE translation

wopfel:
At the end of the Icinga Web 2 setup procedure, there's a nice welcome greeting.

Since there's no database name in the [original message], there's no text that could be replaced by the %s.

This is untested. Please let me know if there are other lines/files that have to be changed in order to make this work.

Kind regards,
Bernd

[original message] https://github.com/Icinga/icingaweb2/blob/81bb5272b59491ad021bcf0a5fefc582ce0112c4/modules/setup/library/Setup/Steps/DatabaseStep.php#L252


wopfel:
Example having the %s highlighted:

Example

nilmerg:
Requires a newly compiled *.mo file to have an effect. But I'll refresh the German translation anyway so I'll just merge this right away. Thank you.

wopfel:
@nilmerg Thanks for the explanation and the merge!

PR #4054, opened 2020-01-14
JS/Modal: Use Icinga.Utils.addUrlFlag()

lippserd:
Icinga.utils.addUrlParams() is not capable of handling our URL filter
syntax but Icinga.Utils.addUrlFlag() is (kind of :D).

requires #4053


nilmerg:
Obsoleted by #4056

PR #4065, opened 2020-02-03
CLI: Setup fake auth

lippserd:
Since version 2.7.x CLI actions load all enabled modules automatically.
This includes launching configuration.php and run.php. Though code
in those files should be restricted to a supported set of functions,
module devs may write any code here. If a module requires authentication
in those files, CLI actions fail because there is no auth possible.
With this patch, we setup a fake user named "cli" w/o any permission when
running CLI actions.

### DIFF
--- a/library/Icinga/Application/Cli.php
+++ b/library/Icinga/Application/Cli.php
@@ -5,6 +5,7 @@ namespace Icinga\Application;
 
 use Icinga\Application\Platform;
 use Icinga\Application\ApplicationBootstrap;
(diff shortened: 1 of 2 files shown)


nilmerg:
> I'm not a fan of the reflection usage. But the alternative would have been to provide a new public function in the Authentication class with the sole purpose to bypass the normal authentication procedure.

If the alternative would require a real sophisticated implementation, I'd agree. But the only thing required here is a setter. A setter that should have been in place from the very beginning.

Issue #4066, opened 2020-02-04
Colour blind theme unreadable monitoring health


theFeu:
When everything is OK under system > monitoring health, there is white text on light green ground in the colour blind module.

PR #4069, opened 2020-02-05
colorblind.less: Add contrast to text with light background

Chivvv:
https://github.com/Icinga/icingaweb2/issues/4066

Changed text colour to black as it was difficult to read.
Hope you have a nice day! (First contribution BTW <3)


theFeu:
There are some more issues with the colorblind theme and modules which need to be addressed.

But this fixes the visual bug in web2 and lgtm!

PR #4081, opened 2020-02-26
Remove 'not checked' from donuts

### LINKED ISSUE #3280: Tactical overview: donuts may misrepresent the environment
## Expected Behavior

The donuts (and their legends?) show the same total amount of monitored objects as the hosts lists.

## Current Behavior

Both the hosts donut and its legend show one more host.

## Possible Solution

Count every monitored object only once.

1. Create a host which is not being checked at all – and becomes pending forever.
2. Look at the donuts.

## Context

#3274

## Your Environment

Only the latest bleeding edge stuff.

baddonuts

theFeu:
fixes #3280

### DIFF
--- a/modules/monitoring/application/views/scripts/tactical/components/hostservicechecks.phtml
+++ b/modules/monitoring/application/views/scripts/tactical/components/hostservicechecks.phtml
@@ -53,7 +53,7 @@
                         <div class="box entry"><?= $this->qlink(
                             sprintf(
                                 $this->translatePlural('%d Disabled', '%d Disabled', $this->statusSummary->hosts_not_checked),
-                                $this->statusSummary->hosts_not_checked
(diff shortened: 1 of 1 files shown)


nilmerg:
I just noticed what `not_checked` actually means. These are hosts/services which won't be checked (checks disabled) and are pending. This *is* a difference from just being pending. So I don't feel that much comfortable removing this entirely from the donuts.

However, it is true that the way it is currently shown it misrepresents the environment. I'd suggest to subtract the amount of `not_checked` totals from their *checked* totals instead. Then the amount of hosts/services matches again.

theFeu:
Smaller change this time :)

More like it, @nilmerg ?

nilmerg:
Obsoleted by #4126

Issue #4098, opened 2020-03-05
Display last state change on service pane


steaksauce-:
## Is your feature request related to a problem? Please describe.
It would be useful to display the last state change in the service pane in a date format, so that information can quickly be gathered WITHOUT going to the history tab.

```
Last State Change | 02-21-2020 15:21:50 | (PING - VOICE)
Last Check | 03-05-2020 13:13:49

```

## Describe the solution you'd like
A modification to the code can easily be implemented to display this. I have an example of code that works with the current version, but it should be noted:
- I do not work with PHP
- This code was written by a self-taught "developer" who left a few years ago (I'm honestly shocked the code works in the newest version).
- This code can be used by someone who knows the language and can clean it up.

The code exists at the bottom of `/usr/share/icingaweb2/modules/monitoring/application/views/scripts/show/components/output.phtml`

```
<?php
if ($object->type == "service" )
{
?>
<h2><?= $this->translate('Timestamps') ?></h2>
<table>
<tr>
<?php
if (strpos(strtolower($object->service),"ping") !== false  )
{
?>
<td width=150><b><u>Last State Change</u></b></td>
<td><b><u>
<?= date('m-d-Y H:i:s',($object->last_state_change)) ?>
</u></b></td>
<?php
}
else
{
?>
<td width=150>Last State Change</td>
<td>
<?= date('m-d-Y H:i:s',($object->last_state_change)) ?>
</td>
<?php
}
?>
<td>
<?php print "(" . $object->service . ")"; ?>
</td>
</tr>
<tr>
<td>Last Check</td>
<td>
<?= date('m-d-Y H:i:s',($object->last_check)) ?>
</td>
</tr>
</table>
<br>
<?php
}
```

## Describe alternatives you've considered
As an organization, we have considered using the history tab, but with recent weather causing THOUSANDS of service outages in monitoring, our NOC has determined that the extra clicks take too long.

We can keep the custom code in place, but it would prevent us from safely upgrading Icingaweb 2 -- which is why we set up an entirely new Icinga2 instance earlier this year (not just this code, but the old developer had a lot of manual bug fixes in place).

## Additional context

![image](https://user-images.githubusercontent.com/8814619/76017835-de19ab00-5ee4-11ea-8d29-96cef672bbbc.png)

nilmerg:
Hi,

the detail view of a service (and a host's one too) already show both of these informations.

[...]

PR #4103, opened 2020-03-10
Add date to status output.

steaksauce-:
Adds a data underneath the status output that users can quickly glance at and copy without hovering over the status

Addresses #4098


nilmerg:
Icinga DB Web includes this now. We have no plans to add this to monitoring's detail areas as well.

PR #4108, opened 2020-03-17
Reload Javascript

nmartinii:


library/Icinga/Authentication/Auth.php
```diff
@@ -135,7 +135,7 @@ public function setAuthenticated(User $user, $persist = true)
         } else {
             $preferences = new Preferences();
         }
-        // TODO(el): Quick-fix for #10957. Only reload CSS if the theme changed.
+        // TODO(el): Quick-fix for #10957. Only reload CSS if the theme changed. check
```
lippserd: I think this change slipped in. Please fix.

library/Icinga/Web/Controller/ActionController.php
```diff
@@ -68,6 +68,8 @@ class ActionController extends Zend_Controller_Action
 
     protected $reloadCss = false;
 
+    protected $reloadJs = false;
```
lippserd: Add PHPDoc.

library/Icinga/Web/Controller/ActionController.php
```diff
@@ -240,6 +242,12 @@ protected function reloadCss()
         return $this;
     }
 
+    protected function reloadJs()
```
lippserd: Add PHPDoc.

library/Icinga/Web/Controller/ActionController.php
```diff
@@ -240,6 +242,12 @@ protected function reloadCss()
...
+    {
+        $this->reloadJs = true;
+        return $this;
```
lippserd: Add newline before return.

library/Icinga/Web/Response.php
```diff
@@ -198,6 +205,30 @@ public function getRequest()
...
+     * Set whether to instruct client side script code to reload JS
+     *
+     * @param   bool    $reloadJs
+     *
+     * @return  $this
+     */
+    public function setReloadJs($reloadJs)
+    {
+        $this->reloadJs = $reloadJs;
+        return $this;
```
lippserd: Add newline before return.

library/Icinga/Web/Response.php
```diff
@@ -198,6 +205,30 @@ public function getRequest()
...
+     */
+    public function isReloadJs()
+    {
+        return $this->reloadJs;
+    }
+
+    /**
+     * Set whether to instruct client side script code to reload JS
+     *
+     * @param   bool    $reloadJs
```
lippserd: Please remove the extra spaces from PHPDoc. I know we in indented everything in the past, but it is enough to add alignment around parameter names (if there is more than one), i.e.

```php
/**
 * Example doc
 *
 * @param int    $short    Something short
 * @param string $fullname Full name
 *
 * @return $this
 */
function foobar($short, $fullname)
{
    return $this;
}
```

public/js/icinga.js
```diff
@@ -226,6 +226,7 @@
                     window.icinga = new window.Icinga(oldConfig);
                     window.icinga.initialize();
                     window.icinga.ui.reloadCss();
+                    window.icinga.ui.reloadJs();
```
lippserd: Why?

public/js/icinga.js
```diff
@@ -235,6 +236,7 @@
                     window.Icinga = oldIcinga;
                     window.icinga = new Icinga(oldConfig);
                     window.icinga.ui.reloadCss();
+                    window.icinga.ui.reloadJs();
```
lippserd: Again, why?

public/js/icinga/loader.js
```diff
@@ -615,6 +615,11 @@
                 this.icinga.ui.reloadCss();
             }
 
+            var jsreload = req.getResponseHeader('X-Icinga-Reload-Js');
+            if (jsreload) {
```
lippserd: You don't need the extra variable.

public/js/icinga/ui.js
```diff
@@ -128,6 +128,12 @@
             });
         },
 
+        reloadJs: function () {
+            var icinga = this.icinga;
```
lippserd: You don't need the extra variable.

library/Icinga/Web/Response.php
nmartinii: Should I fix this too in any other PHPDoc in this file?

library/Icinga/Web/Response.php
nmartinii: In Response.php there are other cases like this. Should i add newlines too?

nilmerg:
Superseded by #4350

PR #4112, opened 2020-03-24
Remember me

### LINKED ISSUE #2495: Implement a "remember me" feature
This issue has been migrated from Redmine: https://dev.icinga.com/issues/12213

**Created by Manul on 2016-07-23 12:57:04 +00:00**

I'd really love not to have to log in on every visit at least from some hosts I regularly use to access Icinga Web 2.

I understand that security is a concern with such a feature and that it would have to be implemented carefully. I had a look at how owncloud implements it since I guess their security requirements are roughly on the same level and the following is an outline of a solution I think could work, based mainly on their implementation:

(DISCLAIMER: I'm not a security expert and I have very limited knowledge of PHP so I might have partly or even completely misunderstood their approach. Please consider all flaws in the following outline mine and all merits owncloud's)

    Configuration options (mandatory)
    ---------------------------------

    - must be possible to dis-/enable site-wide by admin
    - must be possible to dis-/enable per session by user

    Configuration options (optional)
    --------------------------------

    - should be possible to dis-/enable per user by
      a) admin
      b) the user
    - max duration should be configurable by admin
    - per session duration should be selectable on login page

    On presenting login page
    ------------------------

    - check if feature is enabled
    - if yes, present check box/option for duration
    (- possibly disable/hide check box after entry of username if disabled for that user)

    On successful login (with activated feature)
    --------------------------------------------

    (- check if permitted for user)
    - generate random token
    - store uid,token,lifetime in DB
    - store uid,token(,lifetime) in Cookie with appropriate lifetime

    On new session/visit
    --------------------

    - if feature enabled:
      - check if Cookie present
      - check if Cookie still valid
      - check if uid from Cookie exists
      - check if feature enabled for uid
      - check token,lifetime against DB
      - if all tests okay, log user in
      - delete successfully used token and generate new one as above

    If user's credentials change
    ---------------------------

    - delete all tokens for user from DB

    Additional UI
    -------------

    - present option for admin to delete all tokens for specific user/all users
    - present option for user to delete all tokens

I'm looking forward to your comments and insights on this.

lippserd:
resolves #2495


application/forms/Authentication/LoginForm.php
```diff
@@ -95,9 +104,66 @@ public function onSuccess()
...
+
+                // Extract the public key from $res to $pubKey
+                $pubKey = openssl_pkey_get_details($res);
+                $pubKey = $pubKey["key"];
+
+                $name = $user->getUsername();
+                $pswd = $password;
+
+                // Encrypt the username to $name_encrypted using the public key
+                openssl_public_encrypt($name, $name_encrypted, $pubKey);
```
lippserd: Use `$encryptedUsername` as variable name.

application/forms/Authentication/LoginForm.php
```diff
@@ -95,9 +104,66 @@ public function onSuccess()
...
+
+                // Encrypt the username to $pswd_encrypted using the public key
+                openssl_public_encrypt($pswd, $pswd_encrypted, $pubKey);
```
lippserd: Use `$encryptedPassword` as variable name.

application/forms/Authentication/LoginForm.php
```diff
@@ -95,9 +104,66 @@ public function onSuccess()
...
+                $res = openssl_pkey_new($config);
+
+                // Extract the private key from $res to $privKey
+                openssl_pkey_export($res, $privKey);
+
+                // Extract the public key from $res to $pubKey
+                $pubKey = openssl_pkey_get_details($res);
+                $pubKey = $pubKey["key"];
+
+                $name = $user->getUsername();
```
lippserd: Use `$username` as variable name here.

application/forms/Authentication/LoginForm.php
```diff
@@ -95,9 +104,66 @@ public function onSuccess()
...
+                // Encrypt the username to $pswd_encrypted using the public key
+                openssl_public_encrypt($pswd, $pswd_encrypted, $pubKey);
+
+                // change crypt in binary code
+                $bname = base64_encode($name_encrypted);
+                $bpswd = base64_encode($pswd_encrypted);
+
+                // make an Array and serialize it to use it as string
+                $data = [$bname,$bpswd,$pubKey];
+                $data = serialize($data);
```
lippserd: We should not use `serialize()` here because that opens an attack vector if it is somehow possible to inject or manipulate cookie values. `serialize()` is not the problem but `unserialize()` is. Please find out why and create an alternative with `implode()` and `explode()`.

application/forms/Authentication/LoginForm.php
```diff
@@ -95,9 +104,66 @@ public function onSuccess()
...
+
+                //TODO setCookies here and send
+                $rememberme_cookie = new RememberMeCookie();
+                $rememberme_cookie->setValue($data);
+                $this->getResponse()->setCookie($rememberme_cookie);
```
lippserd: You may skip the variable creation and just call:

```php
$this->getResponse()->setCookie(
    (new RememberMeCookie())->setValue($data)
);
```

application/forms/Authentication/LoginForm.php
```diff
@@ -150,4 +169,42 @@ public function onRequest()
...
     }
+
+    /**
+     * This function encrypt username and password,
+     * change these into binary code
+     * @param $username
+     * @param $password
+     * @return array
+     */
+    public function encryptRememberMeData($username, $password)
```
lippserd: It's good that you made a function for the encryption but we have to go a step further because we also have to save the created keys in oder to store them into the database. And of course, we have to test the encryption/decryption.

I opt for creating the class `Icinga\Crypt\RSA` like this:

```php

namespace Icinga\Crypt;

class RSA
{
    public static function keygen();

    public function getPublicKey();

    public function getPrivateKey();

    public function decrypt(...$data);

    public function decryptFromBase64(...$data);

    public function encrypt(...$data);

    public function encryptToBase64(...$data);

    public function loadKey(...$key);
}
```

`RSA::keygen()` generates a new private key and public and returns it as array with the private as first element and the public key as second. The default key size should be 4096 bits.`RSA::loadKey()` should auto-detected the key type. I don't want to specify whether I am loading a private or public key but `loadKey()` should handle this. Please find out how we could do this. Every other method should be self-explanatory.

Usage would be as follows:

```php

// Encryption
$rsa = new RSA();
$rsa->loadKey(...RSA::keygen());

$data = $rsa->encryptToBase64($username, $password);
$data[] = base64_encode($rsa->getPublicKey());
$data = implode('|', $data);

// Decryption
$data = load_cookie();
$data = explode('|', $data);
$publicKey = base64_decode(array_pop($data));

$rsa = new RSA();
$rsa->loadKey($publicKey, load_private_key_from_database($publicKey));

list($username, $password) = $rsa->decryptFromBase64(...$data);

File: application/forms/Authentication/LoginForm.php [hunk 7]
```diff
@@ -98,6 +111,22 @@ public function onSuccess()
         $authenticated = $authChain->authenticate($user, $password);
         if ($authenticated) {
             $auth->setAuthenticated($user);
+            if ($this->getElement('rememberme')->isChecked()) {
+                $rsa = new RSA();
+                $rsa->loadKey(...RSA::keygen());
+                $data = $rsa->encryptToBase64($user->getUsername(), $password);
+                $data[] = base64_encode($rsa->getPublicKey());
+
+                $values['username']     = $user->getUsername();
```
lippserd: Please don't align/indent the equal signs here. A single space after the parameter name is enough.

application/forms/Authentication/LoginForm.php
```diff
@@ -58,6 +63,14 @@ public function createElements(array $formData)
                 'class'         => isset($formData['username']) ? 'autofocus' : ''
             )
         );
+        $this->addElement(
+            'checkbox',
+            'rememberme',
+            [
+                'required'      => false,
```
lippserd: Please indent only up to the longest key + 1 space, i.e.

```php
$array = [
    'long-option' => 'long'
    'short'       => 'short'
];
```

PhpStorm should do that automatically for you. On mac, after selecting the code, press option + command + L.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,203 @@
+<?php
+/* Icinga Web 2 | (c) 2020 Icinga Development Team | GPLv2+ */
+
+namespace Icinga\Crypt;
+
+use InvalidArgumentException;
+use UnexpectedValueException;
+
+/**
+ * Class RSA
```
lippserd: The first line of a class' or method's PHPDoc is the first thing what's read when looking at API docs and should summarise with as less words as possible what the class or method is doing.

I'd prefer something like the following here:

> Data encryption and decryption using RSA keys also supporting key generation

Then you may remove the following extra description:

> This class allows to encrypt and decrypt data.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,203 @@
...
+ * ** Decryption **
+ *
+ * $rsa = new RSA();
+ * $rsa->loadKey($publicKey, load_private_key_from_database($publicKey));
+ *
+ * Decrypt and decodes the given data which is encoded in MIME base64
+ * list($myVar1, $myVar2) = $rsa->decryptFromBase64(...$data);
+ * </code>
+ *
+ * @package Icinga\Crypt
```
lippserd: Please remove the `@package` tag. We don't use that. Also don't forget to remove the superfluous empty lines after removing the tag.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,203 @@
...
+class RSA
+{
+    /** @var string */
+    private $pubKey;
+
+    /** @var string */
+    private $privKey;
+
+    /**
+     * Generates a new private and public key pair
```
lippserd: We use the simple present tense with the method as subject in the I-form for our API docs, i.e.

> Generate ...

Should read: (I, the method) generate a new private and public key pair when I'm called.

By the way, this is nearly the same as for our commit messages where the subject is the commiter, e.g.

> Fix PHPDoc tense of RSA::keygen()

Should read: (I, the commiter) fix ...

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,203 @@
...
+                'digest_alg' => 'sha512',
+                'private_key_bits' => 4096,
+                'private_key_type' => OPENSSL_KEYTYPE_RSA,
+            ]
+        );
+        // Extract the private key from $res to $privKey
+        openssl_pkey_export($res, $privKey);
+        // Extract the public key from $res to $pubKey
+        $pubKey = openssl_pkey_get_details($res);
+        $pubKey = $pubKey["key"];
```
lippserd: Please use single quotes.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,203 @@
...
+    public static function keygen()
+    {
+        $res = openssl_pkey_new(
+            [
+                'digest_alg' => 'sha512',
+                'private_key_bits' => 4096,
+                'private_key_type' => OPENSSL_KEYTYPE_RSA,
+            ]
+        );
+        // Extract the private key from $res to $privKey
```
lippserd: Please remove the comments. You're just documenting what the code is doing here.
Comments are only meaningful if they document the why, not the what (and only if the why is not that obvious).
We don't need comments here.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,203 @@
...
+    private $privKey;
+
+    /**
+     * Generates a new private and public key pair
+     *
+     * @return array
+     */
+    public static function keygen()
+    {
+        $res = openssl_pkey_new(
```
lippserd: Add the array brackets on the same lines as the function opening and closing parenthesis, i.e.

```php
foobar([
    ...
]);
```

This is ok here since the array is the one and only parameter.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,203 @@
...
+    /**
+     * Get the public key
+     *
+     * @return string
+     */
+    public function getPublicKey()
+    {
+        if (empty($this->pubKey)) {
+            throw new UnexpectedValueException('No public key set');
+        }
```
lippserd: Add a newline before the `return`.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,203 @@
...
+    /**
+     * Get the private key
+     *
+     * @return string
+     */
+    public function getPrivateKey()
+    {
+        if (empty($this->privKey)) {
+            throw new UnexpectedValueException('No private key set');
+        }
```
lippserd: Add a newline before the `return`.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,203 @@
...
+    /**
+     * Decrypt the given data
+     *
+     * @param mixed ...$data
+     *
+     * @return array
+     */
+    public function decrypt(...$data)
+    {
+        if (empty(array_filter($data))) {
```
lippserd: This counts for all encryption/decryption functions:
Please don't filter the input data. If the function is called without parameters, we just return the empty array.
With `array_filter` you deny values which should be accepted by our functions.
Please add test cases for the following:

* The encrypt/decrypt methods return the empty array if no parameters are given
* Parameters that are considered empty by `empty()` or considered `false` in `array_filter` must work

Please also check the decrypted value of an encrypted bool `false`.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,203 @@
...
+     */
+    public function loadKey(...$key)
+    {
+        if (count($key) > 2) {
+            throw new InvalidArgumentException(
+                'loadKey() do not expects more then 2 parameters, '.count($key).' given'
+            );
+        }
+        foreach ($key as $k) {
+            if (strpos($k, "PUBLIC")) {
```
lippserd: Please use single quotes.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,203 @@
...
+     * Save the given values in the Variables (private and public key)
+     *
+     * @param mixed ...$key
+     *
+     * @return $this
+     */
+    public function loadKey(...$key)
+    {
+        if (count($key) > 2) {
+            throw new InvalidArgumentException(
```
lippserd: You should use sprintf here, i.e.

```php
throw new InvalidArgumentException(sprintf(
    '%s expects at most 2 keys, %d given',
    __METHOD__,
    count($key)
));

File: library/Icinga/Crypt/RSA.php [hunk 20]
```diff
@@ -0,0 +1,203 @@
...
+            $encodedValues[] = base64_encode($valueToEncode);
+        }
+
+        return $encodedValues;
+    }
+
+    /**
+     * Save the given values in the Variables (private and public key)
+     *
+     * @param mixed ...$key
```
lippserd: Please change `mixed` to `string`.

library/Icinga/Web/RememberMeCookie.php
```diff
@@ -0,0 +1,26 @@
+<?php
+/* Icinga Web 2 | (c) 2020 Icinga Development Team | GPLv2+ */
+
+namespace Icinga\Web;
+
+/**
+ * Class RememberMeCookie
```
lippserd: See the other comment regarding PHPDoc for classes. Fix the first line and remove the package tag.

library/Icinga/Rememberme/Common/Database.php
```diff
@@ -0,0 +1,29 @@
...
+use ipl\Sql\Config as SqlConfig;
+use ipl\Sql\Connection;
+use PDO;
+
+trait Database
+{
+    protected function getDb()
+    {
+        $config = new SqlConfig(ResourceFactory::getResourceConfig(
+            IcingaConfig::module('customer')->get('backend', 'resource')
```
lippserd: Let's just hardcode the resource name to `remember-me` for the moment.
When we've implemented everything, we merge the schema into Icinga Web's schema and adapt the trait accordingly.

application/forms/Authentication/LoginForm.php
```diff
@@ -98,6 +111,21 @@ public function onSuccess()
         $authenticated = $authChain->authenticate($user, $password);
         if ($authenticated) {
             $auth->setAuthenticated($user);
+            if ($this->getElement('rememberme')->isChecked()) {
+                $rsa = new RSA();
+                $rsa->loadKey(...RSA::keygen());
+                $data = $rsa->encryptToBase64($user->getUsername(), $password);
+                $data[] = base64_encode($rsa->getPublicKey());
+
+                $values['username'] = $user->getUsername();
```
lippserd, line 120: One last change to this block here and it's good:

You don't have to cache the values but may passt it directly to insert, i.e.

```php
$this->getDb()->insert('rememberme', [
    ...
]);

Comment (lippserd) [reply application/forms/Authentication/LoginForm.php:120]: And, you have to pass the timestamp for `ctime` and `mtime`.

File: application/controllers/AuthenticationController.php [hunk 24]
```diff
@@ -64,6 +102,14 @@ public function loginAction()
     public function logoutAction()
     {
         $auth = $this->Auth();
+        if (isset($_COOKIE['remember-me'])) {
```
lippserd, line 105: The "forget remember-me cookie code" must be moved down to the place where the code is executed if it's not an external user who is logging out.

Now you check the remember-me cookie even if the user is not authenticated or is an external user. Which is superfluous in both cases.

application/controllers/AuthenticationController.php
```diff
@@ -64,6 +102,14 @@ public function loginAction()
...
+            unset($_COOKIE['remember-me']);
+            $this->getResponse()->setCookie(
```
lippserd, line 107: I think it makes sense to introduce the `forgetMe()` function to the `Cookie` class. Upon calling `forgetMe()` the instance clones itself, sets the value to `null`, sets the expire time to `0` and returns the clone. Then you call it like this:

```php
$this->getResponse()->setCookie(
    (new RememberMeCookie())->forgetMe()
);
```

test/php/library/Icinga/Crypt/RSATest.php
```diff
@@ -0,0 +1,91 @@
...
+        $rsa = (new RSA())->loadKey($publicKey, $privateKey);
+        $this->assertSame($privateKey, $rsa->getPrivateKey());
+        $this->assertSame($publicKey, $rsa->getPublicKey());
+
+        $rsa = (new RSA())->loadKey($privateKey, $publicKey);
+        $this->assertSame($privateKey, $rsa->getPrivateKey());
+        $this->assertSame($publicKey, $rsa->getPublicKey());
+    }
+
+    function testEncryptReturnEmptyArrayIfNoParameterGivenAndReturnEncryptedValueIfParameterIsGiven()
```
lippserd, line 50: Your tests are a bit odd because of all the `empty` and `array_filter` calls and bool comparisons with `assertSame`.
Testing for bools is done via `assertTrue` and `assertFalse`. If you want to test for the empty array use:

```php
$this->assertSame([], $actual);
```

Tests should try cover exactly one use case, so you should have 4 tests to begin with that only test if the empty array is returned when no arguments are given to the encryption and decryption functions, i.e.:

```php
public function testEncryptReturnsEmptyArrayIfNoArgumentsGiven()
{
    $this->assertSame([], (new RSA())->encrypt();
}
```

The same goes for the other functions/tests.

Now I think it does not make sense to test encryption and decryption isolated but have one test case for that, i.e.:

```php
public function testEncryptionAndDecryption()
{
    $rsa = (new RSA())->loadKey(...RSA::keygen());

    $data = ['foo', 'bar'];

    $this->assertSame($data, $rsa->decrypt(...$rsa->encrypt(...$data)));
}

public function testEncryptionToBase64AndDecryptionFromBase64()
{
    ....
}
```

I just saw that all functions are missing the `public` keyword. I started with the tests so it is kind of my fault 😆 
But please add it.

application/controllers/AuthenticationController.php
```diff
@@ -3,18 +3,25 @@
...
+use Icinga\Web\RememberMeCookie;
 use Icinga\Web\Url;
+use ipl\Sql\Select;
 
 /**
  * Application wide controller for authentication
  */
 class AuthenticationController extends Controller
 {
+    use database;
```
lippserd, line 24: Fix typo.

application/controllers/AuthenticationController.php
```diff
@@ -35,6 +42,36 @@ public function loginAction()
...
+        if (isset($_COOKIE['remember-me'])) {
+            $data = explode('|', $_COOKIE['remember-me']);
+            $publicKey = base64_decode(array_pop($data));
+
+            $select = (new Select())
+                ->from('rememberme')
+                ->columns('*')
+                ->where(['public_key = ?' => $publicKey]);
+
+            $DBData = $this->getDb()->select($select)->fetch();
```
lippserd, line 54: Just use `$dbData` if you really want to use that parameter name. I did not go over the "login via remember me cookie" code completely but do that in the next review round ;)

library/Icinga/Web/Cookie.php
```diff
@@ -262,4 +269,13 @@ public function setValue($value)
         $this->value = $value;
         return $this;
     }
+
+    public function forgetMe()
+    {
+        $this->object = $this;
```
lippserd, line 275: Cloning works different. You are just assigning the pointer to the instance to a new variable. With `$this->object` you still refer to the same instance in `$this`. Please see https://www.php.net/manual/en/language.oop5.references.php and https://www.php.net/manual/en/language.references.php. In general this is "pass by value" and "pass by reference".  For cloning please see https://www.php.net/manual/en/language.oop5.cloning.php.

library/Icinga/Web/Cookie.php
```diff
@@ -262,4 +269,13 @@ public function setValue($value)
         $this->value = $value;
         return $this;
     }
+
+    public function forgetMe()
```
lippserd, line 273: Please add PHPDoc with the following information:

* What this method does
* How it's done
* This function alone does not clear the cookie. What else is needed?

For describing that this method returns a new instance of the class use `@return static`.

library/Icinga/Web/RememberMeCookie.php
```diff
@@ -0,0 +1,19 @@
+<?php
+/* Icinga Web 2 | (c) 2020 Icinga Development Team | GPLv2+ */
+
+namespace Icinga\Web;
+
+/**
+ * This class sets the given value of the cookie for 30 days
```
lippserd, line 7: Please change to something like

> Cookie for the "remember me" feature upon login with a default expiration time of 30 days

test/php/library/Icinga/Crypt/RSATest.php
```diff
@@ -0,0 +1,84 @@
...
+        (new RSA())->getPublicKey();
+    }
+
+    /**
+     * @expectedException UnexpectedValueException
+     */
+    public function testGetPrivateKeyThrowsExceptionIfNoPrivateKeySet()
+    {
+        (new RSA())->getPrivateKey();
+    }
```
lippserd, line 35: There's a newline after the function missing.

test/php/library/Icinga/Crypt/RSATest.php
```diff
@@ -0,0 +1,84 @@
...
+        $rsa = (new RSA())->loadKey($publicKey, $privateKey);
+        $this->assertSame($privateKey, $rsa->getPrivateKey());
+        $this->assertSame($publicKey, $rsa->getPublicKey());
+
+        $rsa = (new RSA())->loadKey($privateKey, $publicKey);
+        $this->assertSame($privateKey, $rsa->getPrivateKey());
+        $this->assertSame($publicKey, $rsa->getPublicKey());
+    }
+
+    public function testEncryptReturnsEmptyArrayIfNoArgumentsGiven()
```
lippserd, line 49: Please also test the other 3 functions for the empty array return (in new test methods).

test/php/library/Icinga/Crypt/RSATest.php
```diff
@@ -0,0 +1,84 @@
...
+    public function testEncryptionToBase64AndDecryptionFromBase64()
+    {
+        $rsa = (new RSA())->loadKey(...RSA::keygen());
+
+        $data = ['foo', 'bar'];
+
+        $this->assertSame($data, $rsa->decryptFromBase64(...$rsa->encryptToBase64(...$data)));
+    }
+    
+    public function testDecryptReturnsExactSameValuesAsEncrypted()
```
lippserd, line 73: Please change to `testEncryptionAndDecryptionWithJSON`.

test/php/library/Icinga/Crypt/RSATest.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        $rsa = (new RSA())->loadKey(...RSA::keygen());
```
lippserd, line 75: Add newline after this line for readability.

test/php/library/Icinga/Crypt/RSATest.php
```diff
@@ -0,0 +1,84 @@
...
+        $data = ['int' => 1, 'float' => 1.1, 'yes' => true, 'no' => false, 'null' => null, 'empty-string' => ''];
+        $encodedData = json_encode($data);
+        $encrypted = $rsa->encrypt($encodedData);
+        $decrypted = $rsa->decrypt(...$encrypted);
+        $decodedData = json_decode($decrypted[0],true);
+        $this->assertSame($decodedData, $data);
```
lippserd, line 81: Add newline before this line for readability.

test/php/library/Icinga/Crypt/RSATest.php
```diff
@@ -0,0 +1,84 @@
...
+
```
lippserd, line 82: Remove the newline.

test/php/library/Icinga/Crypt/RSATest.php
```diff
@@ -0,0 +1,84 @@
...
+        $data = ['foo', 'bar'];
+
+        $this->assertSame($data, $rsa->decryptFromBase64(...$rsa->encryptToBase64(...$data)));
+    }
+    
+    public function testDecryptReturnsExactSameValuesAsEncrypted()
+    {
+        $rsa = (new RSA())->loadKey(...RSA::keygen());
+        $data = ['int' => 1, 'float' => 1.1, 'yes' => true, 'no' => false, 'null' => null, 'empty-string' => ''];
+        $encodedData = json_encode($data);
```
lippserd, line 77: Just use `$encoded` as variable name.

test/php/library/Icinga/Crypt/RSATest.php
```diff
@@ -0,0 +1,84 @@
...
+        $encrypted = $rsa->encrypt($encodedData);
+        $decrypted = $rsa->decrypt(...$encrypted);
+        $decodedData = json_decode($decrypted[0],true);
```
lippserd, line 80: Just use `$decoded` as variable name and there is a space missing after the comma.

application/controllers/AuthenticationController.php
```diff
@@ -35,11 +38,19 @@ public function loginAction()
...
+                );
+                (new RememberMe())->remove($this->Auth()->getUser()->getUsername());
+            }
+        }
         if ($this->Auth()->isAuthenticated()) {
             // Call provided AuthenticationHook(s) when login action is called
             // but icinga web user is already authenticated
             AuthenticationHook::triggerLogin($this->Auth()->getUser());
-            $this->redirectNow($form->getRedirectUrl());
+            $this->redirectNow($this->params->get('redirect'));
```
lippserd, line 53: The default redirect to `dashboard` is missing now if there is no redirect parameter given. Please have look how to get the default redirect URL from the `LoginForm` and how it is passed to `UrlParams::get()`.

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,188 @@
...
+    /**
+     * @var object
+     */
+    protected $rsa;
+
+    /**
+     * Check if cookie is set
+     *
+     * @return bool
+     *
```
lippserd, line 42: Superfluous newline.

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,188 @@
...
+     */
+    public static function hasCookie()
+    {
+        return isset($_COOKIE[static::COOKIE]);
+    }
+
+    /**
+     * Get cookie values
+     *
+     * @return void|static
```
lippserd, line 52: It does not return `void` but only `static`.

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,188 @@
...
+        $data = explode('|', $_COOKIE[static::COOKIE]);
+        $publicKey = base64_decode(array_pop($data));
+
+        $select = (new Select())
+            ->from('rememberme')
+            ->columns('*')
+            ->where(['public_key = ?' => $publicKey]);
+
+        $rememberMe = new static();
+        $dbData = $rememberMe->getDb()->select($select)->fetch();
```
lippserd, line 65: Why are you looping over the row and create a new variable here? You may use the result set directly, i.e.

```php
$rs = $rememberMe->getDb()->select($select)->fetch();

$rememberMe->rsa = (new RSA())->loadKey($rs->private_key, $publicKey);
```

Note that we define the following in our database trait

```
PDO::ATTR_DEFAULT_FETCH_MODE => PDO::FETCH_OBJ
```

That means, that every fetched row is an instance of `stdClass` with the columns as public properties.

`PDO::FETCH_ASSOC` on the other hand would create us an associative array for every and `PDO::FETCH_BOTH` allows both property and array access.

There is also one thing missing here. Imagine that the database entry is missing. Regardless of how it came to that.

Please have a look what `fetch()` returns in this case and what we should do if that happens.

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+     * Set the values for 'remember-me' cookie
+     *
+     * @return Cookie
+     */
+    public function getCookie()
+    {
+        $value = $this->rsa->encryptToBase64($this->username);
+        $value[] = $this->encryptedPassword[0];
```
lippserd, line 108: Watch out here. `$encryptedPassword` must never be an array but a string. And array access also works for strings.

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,188 @@
...
+            ->setHttpOnly(true)
+            ->setValue(implode('|', $value));
+    }
+
+    /**
+     * Unset the values for 'remember-me' cookie
+     *
+     * @return Cookie
+     */
+    public static function unsetCookie()
```
lippserd, line 122: Just call this method `forget`.

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,188 @@
...
+    {
+        unset($_COOKIE[static::COOKIE]);
+
+        return (new Cookie(static::COOKIE))
```
lippserd, line 126: Please make use of `Cookie::forgetMe()` here.

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,188 @@
...
+     *
+     * Save new keys in database
+     */
+    public function persist()
+    {
+        $this->getDb()->insert('rememberme', [
+            'username' => $this->username,
+            'private_key' => $this->rsa->getPrivateKey(),
+            'public_key' => $this->rsa->getPublicKey(),
+            'ctime' => date('Y-m-d H:i:s'),
```
lippserd, line 169: Let the database handle the timestamp with `new Expression('NOW()')`. Note that `Expression` is `ipl/Sql/Expression`. The same goes for `mtime`.

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,188 @@
...
+     * @throws \Icinga\Exception\AuthenticationException
+     */
+
+    public function authenticate()
+    {
+        list($password) = $this->rsa->decryptFromBase64($this->encryptedPassword);
+        $auth = Auth::getInstance();
+        $authChain = $auth->getAuthChain();
+        $authChain->setSkipExternalBackends(true);
+        $user = new User($this->username[0]);
```
lippserd, line 146: Again watch out for the array access here. $username must be a string not an array.

application/controllers/AuthenticationController.php
```diff
@@ -35,11 +41,32 @@ public function loginAction()
             $this->redirectNow(Url::fromPath('setup'));
         }
         $form = new LoginForm();
+        if (RememberMe::hasCookie()) {
+            $authenticated = false;
```
lippserd, line 45: Add a newline after the variable definition please.

application/controllers/AuthenticationController.php
```diff
@@ -35,11 +41,32 @@ public function loginAction()
...
+            try {
+                $rememberMe = RememberMe::fromCookie();
+                $authenticated = $rememberMe->authenticate();
+                if ($authenticated) {
```
lippserd, line 49: The authenticated check can be moved outside the if. You already have a `! $authenticated` check there. Also the line can be shortened to `$authenticated = RememberMe::fromCookie()->authenticate()`.

application/controllers/AuthenticationController.php
```diff
@@ -35,11 +41,32 @@ public function loginAction()
...
+                    $this->getResponse()->setCookie(
+                        $rememberMe->renewCookie()
+                    );
+                }
+            } catch (RuntimeException $e) {
+                Logger::error("Can't authenticate user via remember me cookie: %s", $e->getMessage());
+            } catch (AuthenticationException $e) {
+                Logger::error($e);
+            }
+            if (! $authenticated) {
```
lippserd, line 59: Add a newline before the if please.

application/controllers/AuthenticationController.php
```diff
@@ -35,11 +41,32 @@ public function loginAction()
             $this->redirectNow(Url::fromPath('setup'));
         }
         $form = new LoginForm();
+        if (RememberMe::hasCookie()) {
```
lippserd, line 44: Add a newline before the if please.

application/controllers/AuthenticationController.php
```diff
@@ -35,11 +41,32 @@ public function loginAction()
...
+                Logger::error("Can't authenticate user via remember me cookie: %s", $e->getMessage());
+            } catch (AuthenticationException $e) {
+                Logger::error($e);
+            }
+            if (! $authenticated) {
+                $this->getResponse()->setCookie(
+                    RememberMe::unsetCookie()
+                );
+            }
+        }
```
lippserd, line 64: Add a newline after the closing bracket please.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,274 @@
...
+
+namespace Icinga\Crypt;
+
+use InvalidArgumentException;
+use UnexpectedValueException;
+
+/**
+ * Data encryption and decryption using RSA keys also supporting key generation
+ *
+ * # Example Usage
```
lippserd, line 12: Example usage does not reflect our latest changes. The methods are no longer variadic.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,274 @@
...
+
+        return $this->privKey;
+    }
+
+    /**
+     * Decrypt the given data using the set private key
+     *
+     * See {@link loadKey()} for providing the private key.
+     *
+     * @param string|array $data
```
lippserd, line 118: You can be more precise with the array here, i.e. `string[]`.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,274 @@
...
+     *
+     * @return string|array
```
lippserd, line 120: Same as above, please use `string[]`.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,274 @@
...
+     *
+     * @throws UnexpectedValueException If the private key is not set
+     */
+    public function decrypt($data)
+    {
+        $privateKey = $this->getPrivateKey();
+        if (is_array($data)) {
```
lippserd, line 127: Add a newline before the if please.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,274 @@
...
+            $decrypted = [];
+
+            foreach ($data as $value) {
+                openssl_private_decrypt($value, $decrypted[], $privateKey);
+            }
+
+            return $decrypted;
+        }
```
lippserd, line 135: Add a newline after the closing bracket please.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,274 @@
...
+
+        return $decrypted;
+    }
+
+    /**
+     * Decode from Base64 and decrypt the given data using the set private key
+     *
+     * See {@link loadKey()} for providing the private key.
+     *
+     * @param string|array $data
```
lippserd, line 146: `string[]` instead of array.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,274 @@
...
+     *
+     * @return string|array
```
lippserd, line 148: `string[]` instead of array.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,274 @@
...
+    /**
+     * Encrypt the given data using the set public key
+     *
+     * See {@link loadKey()} for providing the public key.
+     *
+     * Please note that this method expects string type arguments. Thus, all arguments are
+     * stringified upon encryption which may lead to unexpected results when decrypting.
+     * Use {@link json_encode()} if you have to encrypt other scalar types than string.
+     *
+     * @param string|array $data
```
lippserd, line 176: `string[]` instead of array.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,274 @@
...
+     *
+     * @return string|array
```
lippserd, line 178: `string[]` instead of array.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,274 @@
...
+    public function encrypt($data)
+    {
+        if (is_array($data)) {
+            $encrypted = [];
+
+            foreach ($data as $value) {
+                openssl_public_encrypt($value, $encrypted[], $this->getPublicKey());
+            }
+            return $encrypted;
+        }
```
lippserd, line 191: Add a newline after the closing bracket please.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,274 @@
...
+     */
+    public function encrypt($data)
+    {
+        if (is_array($data)) {
+            $encrypted = [];
+
+            foreach ($data as $value) {
+                openssl_public_encrypt($value, $encrypted[], $this->getPublicKey());
+            }
+            return $encrypted;
```
lippserd, line 190: Add a newline before the return please.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,274 @@
...
+
+        return $encrypted;
+    }
+
+    /**
+     * Encrypt the given data using the set public key and encode to Base64
+     *
+     * See {@link loadKey()} for providing the public key.
+     *
+     * @param string|array $data
```
lippserd, line 202: `string[]` instead of array.

library/Icinga/Crypt/RSA.php
```diff
@@ -0,0 +1,274 @@
...
+     *
+     * @return string|array
```
lippserd, line 204: `string[]` instead of array.

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /**
+     * Set the values for 'remember-me' cookie
+     *
+     * @return Cookie
+     */
+    public function getCookie()
+    {
+        $value[] = $this->rsa->encryptToBase64($this->username);
```
lippserd, line 104: Do this instead:

```php
$value = [
    encrytped username,
    encrypted password,
    public key
];

File: library/Icinga/Web/RememberMe.php [hunk 68]
```diff
@@ -0,0 +1,206 @@
...
+
+    /**
+     * Database insert for the given private and public key
+     *
+     * Remove old entry for given user if exists
+     * Save new keys in database
+     */
+    public function persist()
+    {
+        $this->remove();
```
lippserd, line 162: Add a newline after this call please.

test/php/library/Icinga/Crypt/RSATest.php
```diff
@@ -0,0 +1,91 @@
...
+    public function testEncryptionAndDecryptionWithJSON()
+    {
+        $rsa = (new RSA())->loadKey(...RSA::keygen());
+
+        $data = ['int' => 1, 'float' => 1.1, 'yes' => true, 'no' => false, 'null' => null, 'empty-string' => ''];
+
+        $encoded = json_encode($data);
+        $encrypted = $rsa->encrypt($encoded);
+        $decrypted = $rsa->decrypt($encrypted);
+        $decoded = json_decode($decrypted,true);
```
lippserd, line 87: There's a space missing after the comma.

application/controllers/AuthenticationController.php
```diff
@@ -35,11 +41,32 @@ public function loginAction()
...
         }
         $form = new LoginForm();
+        if (RememberMe::hasCookie()) {
+            $authenticated = false;
+            try {
+                $rememberMe = RememberMe::fromCookie();
+                $authenticated = $rememberMe->authenticate();
+                if ($authenticated) {
+                    $this->getResponse()->setCookie(
+                        $rememberMe->renewCookie()
```
lippserd, line 51: This should be `$rememberMe->renew()->getCookie()` where `renew()` just returns a new `RememberMe` instance. This way we have full flexibility. Maybe we have to fiddle with the newly created key pair and this way we're prepared.

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /**
+     * Renew the cookie
+     *
+     * Delete old database entry and insert new data for the given user
+     *
+     * @return Cookie
+     */
+    public function renewCookie()
```
lippserd, line 194: This should be `renew()` as stated above. Maybe we should also not call `remove()` and `persist()` automatically here because that is really to much side effect for the function. And since `persist()` calls `remove()` anyway, renewing a remember cookie is as simple and expressive as

```php
$rememberMe
    ->renew()
    ->persist()
    ->getCookie();
```

What do you think?

lippserd:
There are still some things missing we have not talked about yet. At the moment, the feature relies on a separate database. But we should move that to Web's own database. So, please move the schema instructions to `etc/schema/mysql.sql`. Don't forget to create an upgrade file as well and add PostgreSQL support. Web does not require a database at the moment. But this will definitely change in the near future. So, for now we assume that this is the case. Please move the database trait to `Icinga\Common` so that it returns a connection to the Web database. You get the configured resource of the Web database via `Config::app()->get('global', 'config_resource')`.

etc/schema/mysql.schema.sql
```diff
@@ -40,3 +40,16 @@ CREATE TABLE `icingaweb_user_preference`(
   `mtime`    timestamp NULL DEFAULT NULL ON UPDATE CURRENT_TIMESTAMP,
   PRIMARY KEY (`username`,`section`,`name`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+CREATE TABLE rememberme (
```
nilmerg, line 44: All the other tables also use a table prefix. Please add `icingaweb_` here as well.

etc/schema/pgsql.schema.sql
```diff
@@ -100,3 +100,25 @@ CREATE UNIQUE INDEX idx_icingaweb_user_preference
     lower((section)::text),
     lower((name)::text)
 );
+
+CREATE TABLE "rememberme" (
```
nilmerg, line 104: Here please add the table prefix as well.

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,232 @@
...
+use RuntimeException;
+
+class RememberMe
+{
+    use Database;
+
+    /**
+     * Constant cookie
+     */
+    const COOKIE = 'remember-me';
```
nilmerg, line 22: Cookies also use a prefix. Please add `icingaweb2-` as prefix here.

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,232 @@
...
+            ->columns('*')
+            ->where(['public_key = ?' => $publicKey]);
+
+        $rememberMe = new static();
+        $rs = $rememberMe->getDb()->select($select)->fetch();
+
+        if (! $rs) {
+            throw new RuntimeException('No database entry found for the given key');
+        }
+        $rememberMe->rsa = (new RSA())->loadKey($rs->private_key, $publicKey);
```
nilmerg, line 72: Please add an empty line after closing braces.

library/Icinga/Web/RememberMe.php
```diff
@@ -19,7 +19,7 @@ class RememberMe
     /**
      * Constant cookie
      */
-    const COOKIE = 'remember-me';
+    const COOKIE = 'icingaweb-remember-me';
```
nilmerg, line 22: It's `icingaweb2` ;)

library/Icinga/Web/RememberMe.php
```diff
@@ -21,6 +21,11 @@ class RememberMe
      */
     const COOKIE = 'icingaweb-remember-me';
 
+    /**
+     * Constant table for database
+     */
+    const TABLE = 'icingaweb_rememberme';
```
nilmerg, line 27: Nice one! Good idea :+1:

nilmerg:
I don't quite like the placement of the remember me switch on the login page. That the switch is centered is not that ugly, but that the label of the element is not and that far on the left is weird.

Bildschirmfoto von 2020-04-17 17-59-17

public/css/icinga/login.less
```diff
@@ -103,6 +103,24 @@
     }
   }
 
+  form[id^='form_login-']  > div:nth-child(3) {
```
lippserd, line 106: Hey @sukhwinder33445,

Since we still support IE, I don't think it is a good idea to use [CSS Grid](https://caniuse.com/#feat=css-grid).

It is still possible to fix that with "traditional" CSS though:

* The `control-group` of the remember me checkbox must have `text-align: left`
* The `label` for the remember me checkbox must have `float: left`
* The `control-label-group` must have `display: inline-block` and `width: auto`

Then you only have to add some margins.

Because we're adding a special rule/override here, it makes sense to put those rules right before the `#login` bracket closes:

```php
...
# login {
...
  // Remember me checkbox alignment
  .control-group:nth-child(3) {
    ...

    > .control-label-group {
      ...
    }

    > label {
      ...
    }
  }
}
```

sukhwinder33445:
> I just saw the blog post on this upcoming feature and I'm wondering why you are using RSA or asymmetric encryption in general. It looks like you only use the public key as a handle for the corresponding private key. The same could be achieved by just using symmetric encryption and use of the already existing `id` column in the database as the key handle.
> 
> Also, from a quick look at the code, I don't see why the username is stored both in the database table and in cookie.

Hey Julianbrost,

Thank you for reading our blog post and commenting here.

You're absolutely right, we could do this with symmetric encryption. Though we didn't think of that at all to be honest.
Our main concern was/is maximum security and we wanted to make it nearly impossible to guess the shared secret.

We use the username to remove the db entry when the user logs out.

julianbrost:
> You're absolutely right, we could do this with symmetric encryption. Though we didn't think of that at all to be honest.

As it is currently implemented, I think it is possible to encrypt an arbitrary username and password with the public key from the cookie and send it back to the server. This could allow two things:
- If there happens any additional input validation within the login form (I don't know if it does), this can be bypassed with the cookie.
- You can encrypt a different username in the cookie than what is stored in the database. While the current code looks fine in that respect, you would have to remember this for every future change and I think it would be easy to accidentally introduce bugs there by assuming both values are the same.

Also `openssl_public_encrypt` uses PKCS#1 v1.5 padding by default which is vulnerable for Bleichenbacher's attack. I'm not sure if this is actually a problem here, however it is generally recommended to use OAEP padding instead, see for example the man page for `RSA_public_encrypt` from OpenSSL.

> Our main concern was/is maximum security and we wanted to make it nearly impossible to guess the shared secret.

By shared secret you mean the user's password?

> We use the username to remove the db entry when the user logs out.

I see. The code looks like each user can only have one remember me session at a time. Is that intended?

sukhwinder33445:
flowchart

library/Icinga/Web/Menu.php
```diff
@@ -130,5 +130,16 @@ public function init()
...
+        if (Auth::getInstance()->hasPermission('manage/rememberme')) {
+            $this->getItem('system')->addChild($this->createItem('Logged_in_users', [
+                'icon'        => 'host',
+                'description' => t('List of users who stay logged in'),
+                'label'       => t('Logged in users'),
+                'url'         => 'manageuserdevices',
+                'permission'  => 'manage/rememberme',
+                'priority'    => 890
+            ]));
+        }
```
nilmerg, line 143: * Please add the menu entry just like the other entries are added (directly in the array definition)
* Drop the explicit permission check, you're already passing the permission as config option which suffices
* Use the label `User Sessions` instead
* Remember to use the adjust permission name

public/css/icinga/login.less
```diff
@@ -103,6 +103,11 @@
     }
   }
 
+  input:focus,
```
nilmerg, line 106: Please remove this. For the general input types this is already the case. Only the submit button requires this.

public/css/icinga/login.less
```diff
@@ -134,6 +139,21 @@
...
+
+  .control-group:nth-child(3) > div{
+    display: inline-block;
+    width: auto;
+  }
+
+  .control-group:nth-child(3) > label {
+    float: left;
+    margin-right: 1em;
+  }
```
nilmerg, line 156: Please combine these sections with the main section and use the proper classes:

```
  .control-group:nth-child(3) {
    text-align: left;
    margin-top: 1.25em;

    > .control-label-group {
      display: inline-block;
      width: auto;
    }

    > .toggle-switch {
      float: left;
      margin-right: 1em;
    }
  }
```

nilmerg (changes requested):
Completed the first batch. I didn't take a look at most of the newly introduced files. I think I'll look at them once you've completed the encryption and user-agent changes.

library/Icinga/Authentication/Auth.php
```diff
@@ -442,4 +447,19 @@ public function setupUser(User $user)
         $admissionLoader = new AdmissionLoader();
         $admissionLoader->applyRoles($user);
     }
+
+    /**
+     * Read the 'remember-me' cookie and authenticate
+     *
+     * @return bool     True if user authentication is succeed, false if not
+     */
+    public function authenticateFromRememberMeCookie()
```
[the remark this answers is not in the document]
sukhwinder33445: This method in not in use. I will remove this.

library/Icinga/Authentication/Auth.php
```diff
@@ -365,8 +370,8 @@ public function setupUser(User $user)
 
         if ($config->get('global', 'config_backend', 'db') !== 'none') {
             $preferencesConfig = new ConfigObject([
-                'store'     => $config->get('global', 'config_backend', 'db'),
-                'resource'  => $config->get('global', 'config_resource')
+                'store' => $config->get('global', 'config_backend', 'db'),
+                'resource' => $config->get('global', 'config_resource')
```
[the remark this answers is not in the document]
sukhwinder33445: This is a result of rebase, but I think it is not wrong, because the master branch has exactly the same indentation.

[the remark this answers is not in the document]
nilmerg: Nope, [it does not](https://github.com/Icinga/icingaweb2/blob/master/library/Icinga/Authentication/Auth.php#L368-L369)

application/forms/Security/RoleForm.php
```diff
@@ -48,6 +48,105 @@ public function init()
...
+            ],
+            $helper->filterName('config/*') => [
+                'name'          => 'config/*',
+                'description'   => $this->translate('Allow config access')
+            ],
+            $helper->filterName('manage/rememberme') => [
+                'name'          => 'manage/rememberme',
+                'description'   => $this->translate('Allow this user to view and delete other users 
+                rememberme cookies')
+            ]
```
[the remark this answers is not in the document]
nilmerg: > That's the only change that is yours.

With that I've meant that all the other changes you'v introduced are redundant. Please remove them.

nilmerg (changes requested):
We've talked about the user logout issue? This is still the case and now I know how reproduce it: Login using your normal browser window and tick "stay logged in". Then open a private window of the same browser and login without ticking "stay logged in". Logout from the private window. Close the normal browser window and restart it. The login screen appears. (stay logged in doesn't work anymore)

Without looking at the code, I believe this is because the user-agent is the same. Yes, it's technically the same browser so this is not a surprise, but I guess that's also the case if I use a browser from a different machine but with the same OS and browser. That should work. Maybe use a different way to identify a particular session. (the `tag` or `iv`?)

Functional-wise it's working fine. The only exception (albeit a big one) is that every UI interaction leads to a `404 Not Found` for me. That's because many (if not all) routes don't use the correct notation for CamelCase controller names. This results in Zend looking for e.g. `MydevicesController` which works for you, because your OS is MacOS which's file system is case-insensitive. Mine is not and so the controller is not found. If you use CamelCase in a controller name or action, make sure to put dashes in routes where an uppercase char would be after otherwise. (e.g. `mydevices` would then be `my-devices`)

I've updated also some previous comments:
* https://github.com/Icinga/icingaweb2/pull/4112#discussion_r632555869
* https://github.com/Icinga/icingaweb2/pull/4112#discussion_r631101091

library/Icinga/Web/Menu.php
```diff
@@ -54,6 +54,14 @@ public function init()
                     'label'       => t('Announcements'),
                     'url'         => 'announcements',
                     'priority'    => 710
+                ],
+                    'sessions'    => [
```
nilmerg, line 58: That's still wrong :laughing: 

That's correct:

```suggestion
                ],
                'sessions' => [
```

application/controllers/AuthenticationController.php
```diff
@@ -67,16 +94,22 @@ public function logoutAction()
...
             $this->redirectToLogin();
         }
+        $user = $auth->getUser();
         // Get info whether the user is externally authenticated before removing authorization which destroys the
         // session and the user object
-        $isExternalUser = $auth->getUser()->isExternalUser();
+        $isExternalUser = $user->isExternalUser();
         // Call provided AuthenticationHook(s) when logout action is called
-        AuthenticationHook::triggerLogout($auth->getUser());
+        AuthenticationHook::triggerLogout($user);
```
nilmerg, line 102: The changes you've made here aren't necessary since you don't need `$user` below anymore. Please undo them.

application/forms/Security/RoleForm.php
```diff
@@ -48,6 +48,100 @@ public function init()
...
+                    'name'          => $restriction->name,
+                    'label'         => preg_replace(
+                        '~^(\w+)(\/.*)~',
+                        '<em>$1</em>$2',
+                        $view->escape($restriction->name)
+                    ),
+                    'description'   => $restriction->description
+                ];
+            }
+        }
```
nilmerg, line 144: That's still here :thinking:

Please remove everything.

application/controllers/AuthenticationController.php
```diff
@@ -35,11 +39,34 @@ public function loginAction()
...
+                $this->getResponse()->setCookie(RememberMe::forget());
+            }
+        }
+
         if ($this->Auth()->isAuthenticated()) {
             // Call provided AuthenticationHook(s) when login action is called
             // but icinga web user is already authenticated
             AuthenticationHook::triggerLogin($this->Auth()->getUser());
-            $this->redirectNow($form->getRedirectUrl());
+            $this->redirectNow($this->params->get('redirect', $form->getRedirectUrl()));
```
nilmerg, line 69: Why is this change necessary? That's being used by the form anyway, so why are you fetching the param directly again?

nilmerg (changes requested):

application/controllers/ManageUserDevicesController.php
```diff
@@ -0,0 +1,69 @@
...
+                'manage-user-devices',
+                array(
+                    'title'             => $this->translate('List of users who stay logged in'),
+                    'label'             => $this->translate('Users'),
+                    'url'               => 'manage-user-devices',
+                    'data-base-target'  => '_self'
+                )
+            )->activate('manage-user-devices');
+
+        $this->assertPermission('application/sessions');
```
nilmerg, line 32: Check the permission in method `init`. This way it's done for every action. Currently only the index action requires the permission.

application/controllers/ManageUserDevicesController.php
```diff
@@ -0,0 +1,69 @@
...
+            ->add(
+                'manage-devices',
+                array(
+                    'title' => $this->translate('List of devices'),
+                    'label' => $this->translate('Devices'),
+                    'url'   => 'manage-user-devices/devices'
+                )
+            )->activate('manage-devices');
+
+        $name = $this->params->get('name');
```
nilmerg, line 52: This parameter should be required. Use `getRequired` instead.

library/Icinga/Web/RememberMeUserList.php
```diff
@@ -0,0 +1,104 @@
...
+
+        if (empty($this->getUsers())) {
+            $tbody->add(Html::tag('td', t('No user found')));
+        } else {
+            foreach ($this->getUsers() as $user) {
+                $element = Html::tag('tr');
+                $link = new Link(
+                    $user->username,
+                    Url::fromPath($this->getUrl())->addParams(['name' => $user->username]),
+                    ['title' => "Remove stay logged in for {$user->username}"]
```
nilmerg, line 93: That title is wrong. The link guides me to a list of sessions. I can of course remove them there, but that's not the result I get when I click that link.

library/Icinga/Web/RememberMeUserDevicesList.php
```diff
@@ -0,0 +1,144 @@
...
+use ipl\Web\Widget\Link;
+
+class RememberMeUserDevicesList extends BaseHtmlElement
+{
+    protected $tag = 'table';
+
+    protected $defaultAttributes = [
+        'class'             => 'common-table',
+        'data-base-target'  => '_self',
+        'title'             => 'click to remove this cookie from the database'
```
nilmerg, line 18: Please remove the title here. It's the link of each row that's responsible to remove a session, not the entire table. (which I can't click on anyway)

library/Icinga/Web/RememberMeUserList.php
nilmerg: It's also not localized. If you mark this localizable, remember to use sprintf.

library/Icinga/Web/RememberMeUserDevicesList.php
```diff
@@ -0,0 +1,144 @@
...
+        return $this;
+    }
+
+    protected function assemble()
+    {
+        $thead = Html::tag('thead');
+        $theadRow = Html::tag('tr')
+            ->add(Html::tag(
+                'th',
+                t("List of devices and browsers {$this->getUsername()} is currently logged in:")
```
nilmerg, line 102: Use sprintf to inject the username here. The `{ $expr }` syntax is invalid if the string is localizable.

library/Icinga/Web/RememberMeUserDevicesList.php
```diff
@@ -0,0 +1,144 @@
...
+            $tbody->add(Html::tag('td', t('No device found')));
+        } else {
+            foreach ($this->getDevicesList() as $device) {
+                $agent = new UserAgent($device);
+                $element = Html::tag('tr')
+                    ->add(Html::tag('td', $agent->getOs()))
+                    ->add(Html::tag('td', $agent->getBrowser()))
+                    ->add(Html::tag('td', $device->random_iv));
+
+                $link =(new Link(
```
nilmerg, line 125: There's a space missing after the equal sign.

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    /**
+     * Authenticate via the remember me cookie
+     *
+     * @return bool
+     *
+     * @throws \Icinga\Exception\AuthenticationException
+     */
+
```
nilmerg, line 185: remove that empty line

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,318 @@
...
+    public function authenticate()
+    {
+        $password = $this->aesCrypt->decryptFromBase64($this->encryptedPassword);
+        $auth = Auth::getInstance();
+        $authChain = $auth->getAuthChain();
+        $authChain->setSkipExternalBackends(true);
+        $user = new User($this->username);
+        if (! $user->hasDomain()) {
+            $user->setDomain(Config::app()->get('authentication', 'default_domain'));
+        }
```
nilmerg, line 195: add an empty line here

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,318 @@
...
+     * Remove remember me information from the database
+     *
+     * @param string $iv
+     *
+     * @return $this
+     */
+    public function remove($iv)
+    {
+        $this->getDb()->delete(static::TABLE, [
+            'random_iv = ?'         => bin2hex($iv)
```
nilmerg, line 240: If there's only a single key-value pair, just use a single space before `=>`

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,318 @@
...
+     *
+     * Any previous stored information is automatically removed.
+     *
+     * @param null $iv
+     *
+     * @return $this
+     */
+    public function persist($iv = null)
+    {
+        $this->remove($iv);
```
nilmerg, line 215: what happens if `$iv` is null? If you provide `null` as default, you have to guarantee the method call succeeds. Currently it doesn't, or runs a `DELETE` for no reason.

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,318 @@
...
+
+        return $authenticated;
+    }
+
+    /**
+     * Persist the remember me information into the database
+     *
+     * Any previous stored information is automatically removed.
+     *
+     * @param null $iv
```
nilmerg, line 209: It's a string, not (only) null

application/controllers/AuthenticationController.php
sukhwinder33445: Because `$form->getRedirectUrl()` redirects to the default page (dashboard).

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -0,0 +1,222 @@
...
+    /** @var string The authentication tag which is passed by reference when using AEAD cipher mode */
+    private $tag;
+
+    /** @var string The cipher method */
+    private $method = 'aes-128-gcm';
+
+    public function __construct($random_bytes_len = 128)
+    {
+        $len = openssl_cipher_iv_length($this->method);
+        $this->iv = openssl_random_pseudo_bytes($len);
```
nilmerg, line 59: below you're using `random_bytes`. Please do so here as well.

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -0,0 +1,222 @@
...
+        $this->key = random_bytes($random_bytes_len);
+    }
+
+    /**
+     * Set the key
+     *
+     * @return $this
+     *
```
nilmerg, line 67: remove this superfluous empty line in the comment

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -0,0 +1,222 @@
...
+     * Get the key
+     *
+     * @return string
+     *
+     * @throws UnexpectedValueException If the key is not set
+     */
+    public function getKey()
+    {
+        if (empty($this->key)) {
+            throw new UnexpectedValueException('No key set');
```
nilmerg, line 86: That's more of a `LogicException` or `RuntimeException`

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -0,0 +1,222 @@
...
+     * Get the IV
+     *
+     * @return string
+     *
+     * @throws UnexpectedValueException If the IV is not set
+     */
+    public function getIV()
+    {
+        if (empty($this->iv)) {
+            throw new UnexpectedValueException('No iv set');
```
nilmerg, line 114: That's also more of a `LogicException` or `RuntimeException`

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -0,0 +1,222 @@
...
+     * @return string
+     *
+     * @throws UnexpectedValueException If the Tag is not set
+     *
+     * @throws InvalidArgumentException If the Tag is length is not 16
+     */
+    public function getTag()
+    {
+        if (empty($this->tag) || strlen($this->tag) !== 16) {
+            throw new UnexpectedValueException('No tag set');
```
nilmerg, line 144: That empty check should result again in either a `LogicException` or `RuntimeException`.
The length check should already be performed in the setter, here it's too late.

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        if (strlen($this->tag) !== 16) {
+            throw new InvalidArgumentException('Invalid tag value');
+        }
```
nilmerg, line 149: That's the same as above, isn't it?

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -0,0 +1,222 @@
...
+
+    /**
+     * Decrypt the given data using the key, iv and tag
+     *
+     * @param string $data
+     *
+     * @return string
+     *
+     * @throws UnexpectedValueException If decryption fails
+     *
```
nilmerg, line 162: empty line, remove

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -0,0 +1,222 @@
...
+
+    /**
+     * Encrypt the given data using the key, iv and tag
+     *
+     * @param string $data
+     *
+     * @return string encrypted data
+     *
+     * @throws UnexpectedValueException If decryption fails
+     *
```
nilmerg, line 196: ömpty line, r€möv3 :wink:

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -0,0 +1,222 @@
...
+     */
+    public function encrypt($data)
+    {
+        $encrypt = openssl_encrypt($data, $this->method, $this->getkey(), 0, $this->getIV(), $this->tag);
+
+        if (is_bool($encrypt) && $encrypt === false) {
+            throw new UnexpectedValueException('Encryption failed');
```
nilmerg, line 203: That is definitely a `RuntimeException`

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -0,0 +1,222 @@
...
+     * @return string
+     *
+     * @throws UnexpectedValueException If decryption fails
+     *
+     */
+    public function decrypt($data)
+    {
+        $decrypt = openssl_decrypt($data, $this->method, $this->getKey(), 0, $this->getIV(), $this->getTag());
+        if (is_bool($decrypt) && $decrypt === false) {
+            throw new UnexpectedValueException('Decryption failed');
```
nilmerg, line 168: That is definitely a `RuntimeException`

nilmerg (changes requested):
It's 2021 when it's merged, so please update all file headers of all files **you** introduced to use this year instead of 2020.

application/controllers/AuthenticationController.php
nilmerg: Not for me. Regardless of whether I tick "stay logged in" on or not.

nilmerg (changes requested):
I just did a test while my config backend is of type `ini` and I had no `config_resource` configured.

In this case, the login is not accessible. It shows an error instead. (The one in the comment) The same happens upon logout.

Once you adjusted the `Database` trait, use the new method `hasDb()` so that no call to `getDb()` is made if it returns `false`.

I don't mean that you should call `hasDb()` prior to every call to `getDb()`, not necessarily. It's far better to just ignore the entire remember me feature code, if there's no database.

So what I additionally would like to see:
* The checkbox for staying logged in should be disabled (It should still be visible, but a user shouldn't be able to tick it)
* The menu entry and preferences tab should work fine, but should simply show an empty list
  * A hint for the reason should be there. But we should keep this simple: Show a notification with `Icinga\Web\Notification::warning()` upon access of each route showing a list

You can test if it works by adjusting `/etc/icingaweb2/config.ini` in the `web` docker container. Use `config_backend="ini"` and remove the `config_resource` option.

library/Icinga/Web/RememberMe.php
sukhwinder33445: If $iv is null, nothing will be deleted. Should I rather make an if($iv) condition?

application/controllers/AuthenticationController.php
sukhwinder33445: You can reproduce it as follows:

1. Log in with "stay logged" checked.
2. Go to a specific url, copy it and close the browser completely.
3. Open the browser and enter the copied URL.

If you are redirected only with $form->getRedirectUrl(), you will end up on dashbords.

(I think it's because if cookie is present, you won't be redirected to login page at all. So the hidden redirect element is not filled and you end up on default url ).

To change this behaviour, I have added this change.

nilmerg: I see. Thanks for the explanation.

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,332 @@
...
+        if (! (new static())->hasDb()) {
+            return [];
+        }
+
+        $select = (new Select())
+            ->from(static::TABLE)
+            ->columns('username')
+            ->groupBy('username');
+
+        return (new static())->getDb()->select($select)->fetchAll();
```
nilmerg, line 299: Again, a single `new static()` should suffice

library/Icinga/Web/RememberMe.php
```diff
@@ -0,0 +1,332 @@
...
+        if (! (new static())->hasDb()) {
+            return [];
+        }
+
+        $select = (new Select())
+            ->from(static::TABLE)
+            ->columns(['http_user_agent', 'random_iv'])
+            ->where(['username = ?' => $username]);
+
+        return (new static())->getDb()->select($select)->fetchAll();
```
nilmerg, line 320: single new static, you get the idea? :wink:

library/Icinga/Web/UserAgent.php
```diff
@@ -0,0 +1,85 @@
+<?php
+/* Icinga Web 2 | (c) 2021 Icinga GmbH | GPLv2+ */
+namespace Icinga\Web;
```
nilmerg, line 3: empty line before

application/forms/Authentication/LoginForm.php
```diff
@@ -60,6 +64,17 @@ public function createElements(array $formData)
...
         );
+        $this->addElement(
+            'checkbox',
+            'rememberme',
+            [
+                'label' => $this->translate('Stay logged in'),
+            ]
+        );
+        if (! $this->hasDb()) {
+            $this->getElement('rememberme')->setAttrib('disabled', true);
```
nilmerg, line 75: Maybe show also a title in this case: `You can't stay logged in without a database configuration backend`

nilmerg (changes requested):
It's near completion! :+1:

application/controllers/ManageUserDevicesController.php
```diff
@@ -0,0 +1,84 @@
...
+
+        $usersList = (new RememberMeUserList())
+            ->setUsers(RememberMe::getAllUser())
+            ->setUrl('manage-user-devices/devices');
+
+        $this->addContent($usersList);
+
+        if (! $this->hasDb()) {
+            Notification::warning(
+                sprintf($this->translate('No db instance found for users'))
```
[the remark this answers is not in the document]
nilmerg: and remove the sprintf

nilmerg:
I also just noticed that, when deleting a session (either in the same browser or in a different one), the user gets an error upon logout.

nilmerg (approved):
Can't quite believe it, but... LGTM!

Issue #3682, opened 2019-02-07
MySQL query fails in History section because of collation issue


northbear:
Hi, guys 

Icingaweb2 provides MySQL backtrace error messages in sections "History"->"Event Overview" and "History"->"Timeline". In all other sections it looks all works fine. 
Error message is about illegal mix of collations (*latin1_general_ci* and *utf8mb4_0900_ai_ci*). It looks strange because default character set and collation used in icingaweb2 mysql database are defined as **utf8mb4** and **utf8mb4_unicode_ci** respectively. Default charset of PHP is also UTF8. 

So I have no idea collation latin1_general_ci managed to be hardcoded in SQL queries. Another question is where it takes *utf8mb4_0900_ai_ci* collation if it doesn't match defaults used?

```
SQLSTATE[HY000]: General error: 1267 Illegal mix of collations (latin1_general_ci,EXPLICIT) and (utf8mb4_0900_ai_ci,COERCIBLE) for operation 'UNION', query was: SELECT eh.id, eh.host_name, eh.host_display_name, eh.service_description, eh.service_display_name, eh.object_type, eh.timestamp, eh.state, eh.output, eh.type FROM (SELECT sth.id, sth.timestamp, sth.output, sth.type, sth.state, sth.object_type, sth.host_name, sth.service_description, sth.host_display_name, sth.service_display_name FROM (SELECT hh.statehistory_id AS id, ('host') AS object_type, h.display_name COLLATE latin1_general_ci AS host_display_name, ho.name1 AS host_name, NULL AS service_description, NULL AS service_display_name, NULL AS service_host_name, (CASE WHEN hh.state_type = 1 THEN 'hard_state' ELSE 'soft_state' END) AS type, UNIX_TIMESTAMP(hh.state_time) AS timestamp, hh.object_id, hh.state, (CASE WHEN hh.state_type = 1 THEN hh.output ELSE '[ ' || hh.current_check_attempt || '/' || hh.max_check_attempts || ' ] ' || hh.output END) AS output FROM icinga_statehistory AS hh
INNER JOIN icinga_objects AS ho ON ho.object_id = hh.object_id AND ho.is_active = 1 AND ho.objecttype_id = 1
[...]

Issue #3689, opened 2019-02-20
File Upload fails on PHP 7.2 because of count(null)


Thomas-Gelf:

gjongenelen:
Why is this commit not released yet? Facing this error on  Icingaweb 2.6.3 at php7.3

nilmerg:
@gjongenelen It's been overlooked while backporting 7.x related fixes for v2.6.3. :roll_eyes: 

@Thomas-Gelf Please create a PR or set at least a milestone next time.

PR #4120, opened 2020-04-03
Set default IDO charset to latin1

mtorromeo:
latin1 seems to be the only supported charset for MySQL but the current upstream default charset is utf8mb4.

It would be preferable to switch the charset used by icinga to utf8mb4 and the corresponding collations but since there doesn't seem to be any intention of doing this, latin1 should be set explicitly.

This is even more important since the IdoMysqlConnection does not have an attribute to set the charset manually.

This does not seem to be the best place to set the charset but it does not seem that it is used by anything other than the MySQL adapter.

This fixes issues such as #3682 that have been closed prematurely and have nothing to do with the charset of the tables.

### DIFF
--- a/modules/monitoring/library/Monitoring/Backend/MonitoringBackend.php
+++ b/modules/monitoring/library/Monitoring/Backend/MonitoringBackend.php
@@ -222,7 +222,11 @@ class MonitoringBackend implements Selectable, Queryable, ConnectionInterface
     public function getResource()
     {
         if ($this->resource === null) {
-            $this->resource = ResourceFactory::create($this->config->get('resource'));
+            $config = ResourceFactory::getResourceConfig($this->config->get('resource'));
+            if ($this->is('ido') && $config->type === 'db' && $config->db === 'mysql' && $config->charset === null) {
+                $config->charset = 'latin1';
+            }
+            $this->resource = ResourceFactory::createResource($config);
             if ($this->is('ido') && $this->resource->getDbType() !== 'oracle') {
                 // TODO(el): The resource should set the table prefix
                 $this->resource->setTablePrefix('icinga_');


nilmerg:
Hi,

please note that by setting a default charset in `DbConnection` this applies to **any** database connection. This is not only bad because it's just required for the IDO but also only for MySQL.
Also note that the `latin1` charset is only required for the connection Icinga Web 2 is using. The IDO schema defines this charset by itself in various places already and those override the database's default charset. That's why this works with the `IdoMysqlConnection` not providing a configuration for this even if the database has a default charset of `utf8mb4`. For more details please have a look at https://github.com/Icinga/icinga2/issues/5412. (Which I already mentioned in #3689 )

We also can't provide a default anywhere because the charset needs to be defined as part of the resource configuration which has no notion about what the IDO might be. There it's just a MySQL resource, like any other.

The only thing we may do to help users not knowing that `latin1` is the suggested connection charset is by mentioning this in the documentation. Where it's indeed missing. So if you're open for that I happily await your proposal. :)

mtorromeo:
I was already suspecting that it was not the right place to put this configuration into but the issue is not just a documentation one.

If icingaweb2 is the only application that requires latin1, it should call `SET NAMES latin1` when initializing its own connection.

I realize that there are many places where latin1 is explicitly applied as a collation in the queries but this does not solve anything, the whole connection needs to be set to latin1 and sometimes it cannot be configured server-wide in /etc/my.cnf.

I don't know where to touch icingaweb2 so that it calls `SET NAMES` only for its own connection or I would have proposed that change instead.

mtorromeo:
Also:
> That's why this works with the IdoMysqlConnection not providing a configuration for this even if the database has a default charset of utf8mb4

but the reality is that it's not working and while this might not be the right change and knowing that it might have side-effects, it does solve that particular error and I can assure you that there are no mixed colletions/charsets in the schema.

nilmerg:
> I don't know where to touch icingaweb2 so that it calls SET NAMES only for its own connection or I would have proposed that change instead.

That's a resource configuration option. Just configure `latin1` as charset in your IDO resource configuration.

Then, if you've imported the IDO with the default collation and charset of the rdbms you don't have to change anything else for special characters to work. Just did this myself again and umlauts or even japanese symbols work fine on MySQL 8.

With these defaults:

```
mysql> SELECT default_character_set_name FROM information_schema.SCHEMATA 
    -> WHERE schema_name = "icinga2";
+----------------------------+
| DEFAULT_CHARACTER_SET_NAME |
+----------------------------+
| utf8mb4                    |
+----------------------------+
1 row in set (0.00 sec)

mysql> SELECT CCSA.character_set_name FROM information_schema.`TABLES` T,
    ->        information_schema.`COLLATION_CHARACTER_SET_APPLICABILITY` CCSA
    -> WHERE CCSA.collation_name = T.table_collation
    ->   AND T.table_schema = "icinga2"
    ->   AND T.table_name = "icinga_comments";
+--------------------+
| CHARACTER_SET_NAME |
+--------------------+
| utf8mb4            |
+--------------------+
1 row in set (0.00 sec)

mysql> SELECT character_set_name FROM information_schema.`COLUMNS` 
    -> WHERE table_schema = "icinga2"
    ->   AND table_name = "icinga_comments"
    ->   AND column_name = "comment_data";
+--------------------+
| CHARACTER_SET_NAME |
+--------------------+
| latin1             |
+--------------------+
1 row in set (0.00 sec)
```

mtorromeo:
Those 3 queries all return `latin1` on my server but still if I don't set the connection charset to `latin1` I get the error `Illegal mix of collations (latin1_general_ci,EXPLICIT) and (utf8mb4_0900_ai_ci,COERCIBLE) for operation 'UNION'` in the history page.

mtorromeo:
> That's a resource configuration option. Just configure latin1 as charset in your IDO resource configuration

This is not clear to me. Where should I configure the charset? I tried in /etc/icinga2/features-available/ido-mysql.conf but I get `Error: Attribute 'charset' does not exist.`

mtorromeo:
Ok, sorry about the noise. I get what you meant now.
Changing the ido resource configuration through the web interface, indeed fixed the issue I had. Thanks a lot!

Still, the question remains: if this works only with latin1, shouldn't this charset just be imposed by the application, or, at the very least, be the default in case the configuration is empty?

nilmerg:
> shouldn't this charset just be imposed by the application, or, at the very least, be the default in case the configuration is empty?

Yes. There's also a place for this I forgot recently. The class `MonitoringBackend` is the central accessor of the IDO resource. The method [`getResource()`](https://github.com/Icinga/icingaweb2/blob/master/modules/monitoring/library/Monitoring/Backend/MonitoringBackend.php#L222) already implements a specific configuration. It can do the same for the charset.

If you're still up for it, you may change that. Though, please the way that it only applies the charset if there's nothing else configured.

nilmerg:
...and it's a MySQL backend :wink:

mtorromeo:
It doesn't seem like I can change any connection settings in [getResource()](https://github.com/Icinga/icingaweb2/blob/master/modules/monitoring/library/Monitoring/Backend/MonitoringBackend.php#L222) because the connection is immediately established in the [DbConnection::__construct()](https://github.com/Icinga/icingaweb2/blob/95328553ca1336701044282c3b065a20f36af039/library/Icinga/Data/Db/DbConnection.php#L78) and by the time I get the object in getResource it's already too late.

nilmerg:
Correct. What about making class `DbConnection` establish the connection lazily? (i.e. not until it is really required) For this you'd just have to call `connect()` only where property `$dbAdapter` is accessed. Within `connect()` you then check if `$dbAdapter` is unset and return early if not.

lippserd:
> Correct. What about making class `DbConnection` establish the connection lazily? (i.e. not until it is really required) For this you'd just have to call `connect()` only where property `$dbAdapter` is accessed. Within `connect()` you then check if `$dbAdapter` is unset and return early if not.

Wait a moment 😆

Now, we just create the resource which is specified in `backends.ini`.

But you could change that to the following flow:

* Read the resource name from `backends.ini`
* Get the resource configuration via `ResourceFactory::getResourceConfig()`
* Change the config
* Call `ResourceFactory::createResource()` with the changed configuration

Should work this way, right?

nilmerg:
:man_facepalming: I knew I'd miss my daily :coffee: sooner or later

mtorromeo:
I updated the PR based on the feedback. It works for me.

pilot513:
Hi.
I have Icinga2 installed on OEL 8.1. In this case, the backend runs on MySQL 8.
mysql-server.x86_64   8.0.17-3.module+el8.0.0+5253+1dce7bb2   @ol8_appstream

Icinga web2:  2.7.3-1.el8.icinga    @icinga-stable-release

Do I understand correctly that in this version of Icinga2web there are still no these corrections?
I applied your changes to the specified file myself.
How can I clear errors so that they do not appear?

pilot513:
Like this:
`SQLSTATE[HY000]: General error: 1267 Illegal mix of collations (latin1_general_ci,EXPLICIT) and (utf8mb4_0900_ai_ci,COERCIBLE) for operation 'UNION', query was: SELECT eh.id, eh.object_type, eh.host_name, eh.host_display_name, eh.service_description, eh.service_display_name, eh.timestamp, eh.state, eh.output, eh.type FROM (SELECT n.id, n.timestamp, n.output, n.type, n.state, n.object_type, n.host_name, n.service_description, n.host_display_name, n.service_display_name FROM (SELECT hn.notification_id AS id, UNIX_TIMESTAMP(hn.start_time) AS timestamp, ...`

nilmerg:
@pilot513 You just have to set the charset to `latin1` yourself. Do this in your resource configuration in Icinga Web 2 for the IDO connection.

pilot513:
Is it correct (below) now ?
изображение

nilmerg:
Yes.

pilot513:
Thanks for help me!

pilot513:
**Sorry, but error continue**

изображение
изображение

**What should I do to get rid of these errors?**

nilmerg:
Then please make sure the resource you adjusted is really the one that's used for accessing the IDO. (The database is probably `icinga` or `icinga2`)

For further questions on the matter please head over to our forums: https://community.icinga.com

Issue #3770, opened 2019-05-02
Database configuration backend


nilmerg:
At the moment we store most of our configuration on disk in INI files. While this allows easy editing it requires duplicating configuration in a clustered environment.

Subtasks:
* [ ] Dashboards (#1704)
* [ ] Navigation (#2209)

PR #4135, opened 2020-04-21
Enforce database as configuration backend

sukhwinder33445:
refs #3770

### DIFF
--- a/application/forms/Config/General/ApplicationConfigForm.php
+++ b/application/forms/Config/General/ApplicationConfigForm.php
@@ -72,8 +72,8 @@ class ApplicationConfigForm extends Form
 
         // we do not need this form for setup because we set the database there as default.
         // this form is only displayed in configuration -> application if preferences backend type of ini is recognized
-        if ($this->getRequest()->getModuleName() != 'setup'
-            && $formData['global_config_backend'] === 'ini'
+        if (isset($formData['global_config_backend'])
+            && $formData['global_config_backend'] === 'db'
         ) {
             $this->addElement(
                 'select',
@@ -95,7 +95,6 @@ class ApplicationConfigForm extends Form
                 [
                     'required'  => true,
                     'value'     => 'db',
-                    'disabled'  => true
                 ]
             );
         }
--- a/application/forms/Config/General/ApplicationConfigForm.php
+++ b/application/forms/Config/General/ApplicationConfigForm.php
@@ -70,8 +70,7 @@ class ApplicationConfigForm extends Form
             )
         );
 
-        if (
-            $this->getRequest()->getModuleName() != 'setup'
+        if ($this->getRequest()->getModuleName() != 'setup'
             && $formData['global_config_backend'] === 'ini'
         ) {
             $this->addElement(
@@ -107,9 +106,9 @@ class ApplicationConfigForm extends Form
                 array(
                     'required'      => true,
                     'multiOptions'  => array_merge(
-                                            ['' => sprintf(' - %s - ', $this->translate('Please choose'))],
-                                            array_combine($backends, $backends)
-                     ),
+                        ['' => sprintf(' - %s - ', $this->translate('Please choose'))],
+                        array_combine($backends, $backends)
+                    ),
                     'disable'       => [''],
                      'value'         => '',
                     'label'         => $this->translate('Configuration Database')
--- a/application/forms/Config/General/ApplicationConfigForm.php
+++ b/application/forms/Config/General/ApplicationConfigForm.php
@@ -78,9 +78,7 @@ class ApplicationConfigForm extends Form
                 'autosubmit'    => true,
                 'label'         => $this->translate('User Preference Storage Type'),
                 'multiOptions'  => array(
-                    'ini'   => $this->translate('File System (INI Files)'),
-                    'db'    => $this->translate('Database'),
-                    'none'  => $this->translate('Don\'t Store Preferences')
+                    'db'    => $this->translate('Database')
                 )
             )
         );
--- a/application/forms/Config/General/ApplicationConfigForm.php
+++ b/application/forms/Config/General/ApplicationConfigForm.php
@@ -112,7 +112,7 @@ class ApplicationConfigForm extends Form
                      ),
                     'disable'       => [''],
                      'value'         => '',
-                    'label'         => $this->translate('Database Connection')
+                    'label'         => $this->translate('Configuration Database')
                 )
             );
         }
--- a/application/forms/Config/General/ApplicationConfigForm.php
+++ b/application/forms/Config/General/ApplicationConfigForm.php
@@ -73,7 +73,7 @@ class ApplicationConfigForm extends Form
         // we do not need this form for setup because we set the database there as default.
         // this form is only displayed in configuration -> application if preferences backend type of ini is recognized
         if (isset($formData['global_config_backend'])
-            && $formData['global_config_backend'] === 'db'
+            && $formData['global_config_backend'] === 'ini'
         ) {
             $this->addElement(
                 'select',
@@ -95,10 +95,11 @@ class ApplicationConfigForm extends Form
                 [
                     'required'  => true,
                     'value'     => 'db',
+                    'disabled'  => true
                 ]
             );
         }
-        if (isset($formData['global_config_backend']) && $formData['global_config_backend'] === 'db') {
+        if (! isset($formData['global_config_backend']) || $formData['global_config_backend'] === 'db') {
             $backends = array();
             foreach (ResourceFactory::getResourceConfigs()->toArray() as $name => $resource) {
                 if ($resource['type'] === 'db') {
--- a/application/forms/Config/General/ApplicationConfigForm.php
+++ b/application/forms/Config/General/ApplicationConfigForm.php
@@ -72,9 +72,7 @@ class ApplicationConfigForm extends Form
 
         // we do not need this form for setup because we set the database there as default.
         // this form is only displayed in configuration -> application if preferences backend type of ini is recognized
-        if (isset($formData['global_config_backend'])
-            && $formData['global_config_backend'] === 'ini'
-        ) {
+        if (isset($formData['global_config_backend']) && $formData['global_config_backend'] === 'ini') {
             $this->addElement(
                 'select',
                 'global_config_backend',
@@ -99,6 +97,7 @@ class ApplicationConfigForm extends Form
                 ]
             );
         }
+
         if (! isset($formData['global_config_backend']) || $formData['global_config_backend'] === 'db') {
             $backends = array();
             foreach (ResourceFactory::getResourceConfigs()->toArray() as $name => $resource) {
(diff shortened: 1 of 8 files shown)


nilmerg:
Just did a basic functionality test. Things I discovered:

* Class `GeneralConfigForm` correctly shows the warning, though doesn't let me then change the backend type to database. We should probably still show the selection to change this if INI is detected as backend. And only hide it completely if it's already a database backend.
* Class `GeneralConfigForm` does not let me choose which resource to use anymore. It's fine that it doesn't let me choose the backend type, but we still have to let the user choose which database resource to use as configuration backend.
* The log message should include that it's the **preferences** backend of type INI that is deprecated. (Current message: `Ini backend type is deprecated and will be removed with version 2.10`)
* I can continue in the wizard although both `pdo_mysql` and `pdo_pgsql` are missing. The wizard must let me continue only if at least one of these is available.
* The wizard also does not demand from me to setup the configuration backend resource. It's fine that I don't have the choice to select any backend type, but then the wizard must ask me where my database is located so that it can set it up.

nilmerg:
Tested again. Noticed the following:

* Class `GeneralConfigForm` also allows me to choose not to save preferences at all. Please remove that option from the list as it's also obsolete now. The only possible change should be from INI to DB.
* Class `GeneralConfigForm` shows `Database Connection` for the resource dropdown as label. Please show `Configuration Database` instead.

Otherwise it looks okay. Except the resource setup in the wizard, but you're already on it. :)

nilmerg:
Choosing an external authentication backend in the wizard shows me this:
```
Trying to access array offset on value of type null

#0 /icingaweb2/modules/setup/library/Setup/WebWizard.php(235): Icinga\Application\ApplicationBootstrap->Icinga\Application\{closure}(Integer, String, String, Integer, Array)
#1 /icingaweb2/library/Icinga/Web/Wizard.php(271): Icinga\Module\Setup\WebWizard->setupPage(Object(Icinga\Module\Setup\Forms\AuthenticationPage), Object(Icinga\Web\Request))
#2 /icingaweb2/modules/setup/application/controllers/IndexController.php(46): Icinga\Web\Wizard->handleRequest()
#3 /icingaweb2/library/vendor/Zend/Controller/Action.php(507): Icinga\Module\Setup\Controllers\IndexController->indexAction()
#4 /icingaweb2/library/Icinga/Web/Controller/Dispatcher.php(76): Zend_Controller_Action->dispatch(String)
#5 /icingaweb2/library/vendor/Zend/Controller/Front.php(937): Icinga\Web\Controller\Dispatcher->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
#6 /icingaweb2/library/Icinga/Application/Web.php(300): Zend_Controller_Front->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
#7 /icingaweb2/library/Icinga/Application/webrouter.php(99): Icinga\Application\Web->dispatch()
#8 /icingaweb2/public/index.php(4): require_once(String)
#9 {main}
```

modules/setup/library/Setup/WebWizard.php
```diff
@@ -649,9 +636,10 @@ public function getRequirements($skipModules = false)
             )
         )));
 
-        $mysqlSet = new RequirementSet(true);
+        $isOptional = Platform::extensionLoaded('pdo_pgsql');
```
nilmerg, line 639: Don't use `Platform` here. Create a new `RequirementSet` instead and set it as mandatory.

i.e.
```
$set ----- $dbSet
         /   OR  \
        /         \
$mysqlSet          $pgsqlSet
```

application/forms/Config/General/ApplicationConfigForm.php
```diff
@@ -70,21 +70,32 @@ public function createElements(array $formData)
...
-            array(
-                'required'      => true,
-                'autosubmit'    => true,
-                'label'         => $this->translate('User Preference Storage Type'),
-                'multiOptions'  => array(
-                    'ini'   => $this->translate('File System (INI Files)'),
-                    'db'    => $this->translate('Database'),
-                    'none'  => $this->translate('Don\'t Store Preferences')
+        if ($this->getRequest()->getModuleName() != 'setup'
+            && $formData['global_config_backend'] === 'ini'
```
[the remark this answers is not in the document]
sukhwinder33445: this form should not be displayed in the setup and should only be displayed in the configuration -> application if the ini file is recognized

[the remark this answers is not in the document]
nilmerg: Okay, fine. But the wizard doesn't let you choose the type anymore at all, so it can't be of type INI and therefore the second condition should suffice, shouldn't it?

modules/setup/library/Setup/WebWizard.php
```diff
@@ -259,6 +245,12 @@ public function setupPage(Form $page, Request $request)
                     unset($pageData['setup_config_db_resource']);
                     unset($pageData['setup_config_db_creation']);
                 }
+            } elseif ($authData != null) {
+                // If you validate the database and then come back to change the authentication type
+                // but do not change it, you will get an database configuration related error message on the next page.
+                // To avoid this error, the 'setup_config_db_resource' page must be unset.
```
nilmerg, line 251: I can't reproduce this. Not on the master and not with your change here undone. Please make sure you are really sure that this is necessary.

application/forms/Config/General/ApplicationConfigForm.php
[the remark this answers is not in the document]
sukhwinder33445: yes, you are right, but we have to add a check if that variable exists.

application/forms/Config/General/ApplicationConfigForm.php
```diff
@@ -70,21 +70,32 @@ public function createElements(array $formData)
...
-            )
-        );
-
+            );
+        } else {
+            $this->addElement(
+                'hidden',
+                'global_config_backend',
+                array(
+                    'required'  => true,
```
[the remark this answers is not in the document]
sukhwinder33445: I have added this, but it does not work as expected, and do not let me change the configuration database and config backend type.

modules/setup/library/Setup/WebWizard.php
sukhwinder33445: This is necessary if user choose external as Auth type,  validate database and then come back to change Auth type, but do not change.
I saw that user must validate DB again if he come back (if database was selected as Auth), so I set the if condition for all. We can change it to just external if needed

application/forms/Config/General/ApplicationConfigForm.php
```diff
@@ -70,22 +70,36 @@ public function createElements(array $formData)
...
-                    'none'  => $this->translate('Don\'t Store Preferences')
-                )
-            )
-        );
-
-        if (isset($formData['global_config_backend']) && $formData['global_config_backend'] === 'db') {
+        // we do not need this form for setup because we set the database there as default.
+        // this form is only displayed in configuration -> application if preferences backend type of ini is recognized
+        if (isset($formData['global_config_backend'])
+            && $formData['global_config_backend'] === 'ini'
```
nilmerg, line 76: It's not too long for both conditions on the same line. Please remove that line break.

application/forms/Config/General/ApplicationConfigForm.php
```diff
@@ -70,22 +70,36 @@ public function createElements(array $formData)
...
+                'hidden',
+                'global_config_backend',
+                [
+                    'required'  => true,
+                    'value'     => 'db',
+                    'disabled'  => true
+                ]
+            );
+        }
+        if (! isset($formData['global_config_backend']) || $formData['global_config_backend'] === 'db') {
```
nilmerg, line 102: Here's again an empty line missing ;)

application/forms/Config/General/ApplicationConfigForm.php
```diff
@@ -98,8 +112,13 @@ public function createElements(array $formData)
...
                 array(
                     'required'      => true,
-                    'multiOptions'  => $backends,
-                    'label'         => $this->translate('Database Connection')
+                    'multiOptions'  => array_merge(
+                        ['' => sprintf(' - %s - ', $this->translate('Please choose'))],
+                        $backends
+                    ),
+                    'disable'       => [''],
+                     'value'        => '',
```
nilmerg, line 120: There's still this superfluous space right before `'value'`

lippserd:
Rebase and squash please.

modules/setup/library/Setup/WebWizard.php
```diff
@@ -711,7 +709,23 @@ public function getRequirements($skipModules = false)
                 'setup.requirement.class'
             )
         )));
-        $set->merge($pgsqlSet);
+        $dbSet->merge($pgsqlSet);
+        $set->merge($dbSet);
+
+        $dbRequire = new RequirementSet(true);
+        $dbRequire->add(new SetRequirement(array(
+            'optional'      => $dbSet->getState(),
```
nilmerg, line 717: Why is the optional state dependent on the set's state? This requirement should always be mandatory (`'optional' => false`)

modules/setup/library/Setup/WebWizard.php
```diff
@@ -711,7 +709,23 @@ public function getRequirements($skipModules = false)
...
+            'condition'     => $dbSet->getState(),
```
nilmerg, line 718: Please pass the actual set here to the requirement. It's a *Set*Requirement after all, not a *Boolean*Requirement.

modules/setup/library/Setup/WebWizard.php
```diff
@@ -711,7 +709,23 @@ public function getRequirements($skipModules = false)
                 'setup.requirement.class'
             )
         )));
-        $set->merge($pgsqlSet);
+        $dbSet->merge($pgsqlSet);
+        $set->merge($dbSet);
+
+        $dbRequire = new RequirementSet(true);
```
nilmerg, line 715: You don't need a requirement set here. Just add the `SetRequirement` directly to `$set`.

sukhwinder33445: This is not possible because merge() requires a parameter of type RequirementSet

nilmerg: That's why I used `add`, I'm sure there's a corresponding method. :wink:

sukhwinder33445: I got it. :)

modules/setup/library/Setup/WebWizard.php
```diff
@@ -711,7 +709,23 @@ public function getRequirements($skipModules = false)
...
+        $set->merge($dbSet);
+
+        $dbRequire = (new SetRequirement(array(
+            'optional'      => false,
+            'condition'     => $dbSet,
+            'title'         =>'Database',
+            'alias'         => 'PDO-MySQL OR PDO-PostgreSQL',
+            'description'   => mt(
+                'setup',
+                'To store users or preferences in a database at least one module PDO-MySQL OR PDO-PostgreSQL for PHP
```
nilmerg, line 722: Please don't mention the reason why a database module is required. On one hand it's not only the user preferences but also users themselves, on the other hand both reasons are mentioned directly above. So I think we can reduce this to mentioning that a database is mandatory and hence these modules are required.

nilmerg (changes requested):

nilmerg (approved):

PR #4146, opened 2020-05-06
Allow to set restrictions for super user in the UI

### LINKED ISSUE #4068: It should be possible to set restrictions for users with administrative access in the UI
## Describe the bug

While it is perfectly possible (and desirable) to define restrictions for users having `permissions = *`, in the UI this is currently requires the following steps:

* switch admin access off
* specify your restriction
* switch admin access on
* store your role

Yonas-net:
resolves #4068

### DIFF
--- a/application/forms/Security/RoleForm.php
+++ b/application/forms/Security/RoleForm.php
@@ -207,9 +207,10 @@ class RoleForm extends RepositoryForm
                         'checkbox',
                         $name,
                         [
-                            'onclick'       => 'return false;',
-                            'onkeydown'     => 'return false;',
-                            'value'         => $hasFullPerm,
+                            'autosubmit'    => isset($spec['isFullPerm']),
+                            'disabled'      => isset($formData[self::WILDCARD_NAME]) &&
+                                                $formData[self::WILDCARD_NAME]?: null,
+                            'value'         => $formData[self::WILDCARD_NAME],
                             'label'         => preg_replace(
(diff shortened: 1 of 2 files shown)


Yonas-net:
I tested it like you told me to, so after enabling administrative access I set the other permissions to read only, if I save it that way the previous configurations are lost. I tested it again without to set the other permissions to read-only and this is what comes out (e.g 'permissions' =>``'*,businessprocess/showall,businessprocess/create,businessprocess/modify``).

is this how it would work?

Yonas-net:
Since all other permissions are set to zero or read-only after I have activated administrative access, I don't know how I should save the previous configuration with activated administrative access, because the getValues only receives ``permission=*``.

nilmerg:
`getValues()` only returns values for elements wich are not ignored (option `ignore => true`). This option is set to true when the administrative permission is enabled. Remove this and you'll get all other checkbox states as well.

nilmerg:
Points 1, 3 and 5 are now true. For point 2 however only the first part (read-only elements) is true, the checkboxes are not enabled though. I couldn't verify point 4 because when storing a role with administrative access and some other permissions set, the checkbox for the administrative permission is not enabled when re-opening the role.

nilmerg:
Points 1 to 3 are now true, 4 and 5 however are not.

nilmerg:
Superseded by #4301

PR #4161, opened 2020-05-21
Add theme with different soft state colors

### LINKED ISSUE #4160: A theme with different colors for SOFT and HARD states
Most of the time I would like to ignore services that are in SOFT Critical state. But currently on a first glance they look the same as the ones in HARD Critical state because of the same color.

I would like to be able to create a color theme where a different color is used for SOFT and HARD Critical state and to create a new theme based on the default one that uses it. I would leave the current theme as default. 

## Describe alternatives you've considered
None.

## Additional context
Now:
Screenshot 2020-05-21 at 18 00 04
My (bad) mockup of how a different colors could look like:

gdubicki:
fixes #4160

### DIFF
--- a/public/css/icinga/mixins.less
+++ b/public/css/icinga/mixins.less
@@ -185,10 +185,22 @@
       background-color: @color-warning-handled;
     }
   }
+  &.state-warning-soft {
+    background-color: @color-warning-soft;
+    &.handled {
+      background-color: @color-warning-soft-handled;
+    }
+  }
   &.state-critical {
     background-color: @color-critical;
     &.handled {
       background-color: @color-critical-handled;
+    }
+  }
+  &.state-critical-soft {
+    background-color: @color-critical-soft;
+    &.handled {
+      background-color: @color-critical-soft-handled;
     }
   }
   &.state-down {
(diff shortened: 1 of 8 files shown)


public/css/icinga/mixins.less
```diff
@@ -185,12 +185,24 @@
       background-color: @color-warning-handled;
     }
   }
+  &.state-warning-soft {
```
nilmerg, line 188: Shouldn't this be `&.state-warning.soft`?

nilmerg:
Are you going to continue with this?

gdubicki:
Yes, I will @nilmerg . Sorry, I was busy with something else. But adding this feature is getting back on my radar so I should do it within the next week or two.

gdubicki:
When I think about it more, @nilmerg , I think that what we really need is a much bigger feature than this small PR. We would like to globally not treating soft critical state as like a hard state.

nilmerg:
> We would like to globally not treating soft critical state as like a hard state.

A critical state should always be a hard state?

nilmerg:
You seem to not want to continue with this, at least not with solving your request with a theme. If you really want to be able to change the state handling, you should head over to [Icinga 2](https://github.com/Icinga/icinga2).

PR #4170, opened 2020-06-08
Make auto refresh interval configurable

### LINKED ISSUE #2819: Auto refresh interval should be configurable
Icingaweb 2.4.1

As a user, I can enable or disable auto refresh (e.g. when viewing a service, the tab reloads every x seconds). But unfortunately I can not configure how often the tab/page is automatically reloaded.

Browsing the code, I found several hard-coded values for auto refresh settings, separated by object type (e.g. service, host..).

I would like to suggest that a user can configure the auto refresh interval for those objects within his profile settings. In addition, there should be a minimal interval, e.g. 5s, in order to prevent hundreds of users from "DoSing" the local webserver.

Al2Klimov:
fixes #2819

### DIFF
--- a/application/forms/PreferenceForm.php
+++ b/application/forms/PreferenceForm.php
@@ -44,6 +44,7 @@ class PreferenceForm extends Form
     public function init()
     {
         $this->setName('form_config_preferences');
+        $this->setSubmitLabel($this->translate('Save to the Preferences'));
     }
 
     /**
@@ -90,12 +91,6 @@ class PreferenceForm extends Form
      */
     public function onSuccess()
     {
-        if (!($this->getElement('btn_submit_preferences')->isChecked()
-            || $this->getElement('btn_submit_session')->isChecked()
-        )) {
-            return false;
-        }
-
         $this->preferences = new Preferences($this->store ? $this->store->load() : array());
 
         $oldTheme = $this->preferences->getValue('icingaweb', 'theme');
@@ -124,7 +119,7 @@ class PreferenceForm extends Form
         }
 
         try {
-            if ($this->store && $this->getElement('btn_submit_preferences')->isChecked()) {
+            if ($this->store && $this->getElement('btn_submit')->isChecked()) {
                 $this->save();
                 Notification::success($this->translate('Preferences successfully saved'));
             } else {
@@ -330,7 +325,7 @@ class PreferenceForm extends Form
         if ($this->store) {
             $this->addElement(
                 'submit',
-                'btn_submit_preferences',
+                'btn_submit',
                 array(
                     'ignore'        => true,
                     'label'         => $this->translate('Save to the Preferences'),
@@ -363,7 +358,7 @@ class PreferenceForm extends Form
         );
 
         $this->addDisplayGroup(
-            array('btn_submit_preferences', 'btn_submit_session', 'preferences-progress'),
+            array('btn_submit', 'btn_submit_session', 'preferences-progress'),
             'submit_buttons',
             array(
                 'decorators' => array(
@@ -372,6 +367,20 @@ class PreferenceForm extends Form
                 )
             )
         );
+    }
+
+    public function addSubmitButton()
+    {
+        return $this;
+    }
+
+    public function isSubmitted()
+    {
+        if (parent::isSubmitted()) {
+            return true;
+        }
+
+        return $this->getElement('btn_submit_session')->isChecked();
     }
 
     /**
--- a/application/forms/PreferenceForm.php
+++ b/application/forms/PreferenceForm.php
@@ -90,6 +90,12 @@ class PreferenceForm extends Form
      */
     public function onSuccess()
     {
+        if (!($this->getElement('btn_submit_preferences')->isChecked()
+            || $this->getElement('btn_submit_session')->isChecked()
+        )) {
+            return false;
+        }
+
         $this->preferences = new Preferences($this->store ? $this->store->load() : array());
 
         $oldTheme = $this->preferences->getValue('icingaweb', 'theme');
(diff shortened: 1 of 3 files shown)


application/forms/PreferenceForm.php
```diff
@@ -283,6 +294,36 @@ public function createElements(array $formData)
...
+                // Using Form#translatePlural() not for $speed==1 and $speed!=1,
+                // but for different $speed-dependent plural forms, e.g. in Russian
+                $speeds[$speed] = sprintf($this->translatePlural('%dx slower', '%dx slower', $speed), $speed);
+                $speeds[rtrim(sprintf('%F', 1.0 / $speed), '0')] = sprintf(
+                    $this->translatePlural('%dx faster', '%dx faster', $speed),
+                    $speed
+                );
+            }
+
+            krsort($speeds);
```
nilmerg, line 310: Please use the options as discussed in the issue:

```suggestion
            $speeds = [
                0.5 => $this->translate('Fast', 'refresh_speed'),
                ''  => $this->translate('Default', 'refresh_speed'),
                2.0 => $this->translate('Moderate', 'refresh_speed'),
                4.0 => $this->translate('Slow', 'refresh_speed')
            ];
```

application/forms/PreferenceForm.php
```diff
@@ -283,6 +294,36 @@ public function createElements(array $formData)
...
+
+            $this->addElement(
+                'select',
+                'auto_refresh_speed',
+                array(
```
nilmerg, line 315: Please use `[` and `]`.

application/forms/PreferenceForm.php
```diff
@@ -283,6 +294,36 @@ public function createElements(array $formData)
...
+                    'required'      => false,
+                    'label'         => $this->translate('Auto refresh speed'),
+                    'description'   => $this->translate(
+                        'This option allows you to speed up or to slow down the global page content auto refresh'
+                    ),
+                    'multiOptions'  => $speeds,
+                    'value'         => 1
```
nilmerg, line 322: Use the empty string here as default value. This results in the preference setting being removed from the configuration once the user chooses `Default`.

library/Icinga/Web/Controller/ActionController.php
```diff
@@ -519,7 +519,19 @@ protected function postDispatchXhr()
...
         if ($this->autorefreshInterval !== null) {
-            $resp->setAutoRefreshInterval($this->autorefreshInterval);
+            $user = $this->getRequest()->getUser();
+
+            if ($user === null) {
+                $speed = 1;
+            } else {
+                $speed = (float) $user->getPreferences()->getValue('icingaweb', 'auto_refresh_speed', 1.0);
+
+                if ($speed <= 0) {
```
nilmerg, line 529: Don't validate the factor, but the result after applying it. We want a minimum speed but defined in absolute seconds: 5

Though, only if it's the result after the user's speed has been applied. If a controller sets an interval below 5 seconds, it **must** be respected.

library/Icinga/Web/Controller/ActionController.php
```diff
@@ -519,7 +519,19 @@ protected function postDispatchXhr()
...
+                    $speed = 1;
+                }
+            }
+
+            $resp->setAutoRefreshInterval($this->autorefreshInterval * $speed);
```
nilmerg, line 534: Round the result to nearest integer. Our JS does not tick faster than 1s anyway.

library/Icinga/Web/Response.php
```diff
@@ -302,7 +302,8 @@ protected function prepare()
                 $this->setHeader('X-Icinga-Reload-Css', 'now', true);
             }
             if (($autoRefreshInterval = $this->getAutoRefreshInterval()) !== null) {
-                $this->setHeader('X-Icinga-Refresh', $autoRefreshInterval, true);
+                // Using sprintf('%F',$autoRefreshInterval) to set e.g. '1.25' and not '1,25'
+                $this->setHeader('X-Icinga-Refresh', sprintf('%F', $autoRefreshInterval), true);
```
nilmerg, line 306: This should be an integer, so this change isn't necessary.

nilmerg (changes requested):
Also, please allow callers of `ActionController::setAutorefreshInterval()` to disable the application of a user's refresh speed customizations. I don't want to see a new property for that though, adjust the speed directly in `setAutorefreshInterval()` then to avoid that.

Once that's done, disable user customizations for the refresh speeds defined by the  `ApplicationStateController`.

application/forms/PreferenceForm.php
```diff
@@ -90,6 +90,12 @@ public function save()
      */
     public function onSuccess()
     {
+        if (!($this->store && ($this->getElement('btn_submit_preferences')->isChecked()
+            || $this->getElement('btn_submit_session')->isChecked()
+        ))) {
+            return false;
+        }
+
```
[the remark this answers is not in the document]
Al2Klimov: For property handling the newly introduced autosubmit.

Al2Klimov:
Wouldn’t the user wonder (and complain) why the customization doesn’t work for action XY?

application/forms/PreferenceForm.php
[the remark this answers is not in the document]
nilmerg: `onSuccess()` shouldn't run in case of an auto submit.

nilmerg:
Of course this is not the default and it should only be used rarely. The application-state requests however run in the background, that's why the user doesn't ever notice any change and shouldn't be able to adjust it's interval.

library/Icinga/Web/Controller/ActionController.php
Al2Klimov: What for? And why 5?

nilmerg: > What for?

Because it's unreasonable to refresh a page at a 1 second interval that's the result of a factor defined in some configuration file. We don't want to allow users to DDOS themselves.

> And why 5?

Because. Well, because there has to be a lower limit, and 5 seconds was the suggestion of a user in the issue and I think it's a sufficiently low/high limit.

nilmerg (approved):

PR #4114, opened 2020-03-26
Enhanced dashboards

Yonas-net:
resolves #3809

PR #4180, opened 2020-06-17
Automatically load module ipl

nilmerg:
IPL is a requirement for the following features:

#4135 
#4112 
#4114


nilmerg:
Obsolete since #4271

PR #4185, opened 2020-06-18
Bugfix/keep closing x for panes available at all times 3461

flourish86:
refs #3461 

Works in Safari so far. Has to be tested in other browsers yet.


nilmerg:
By looking at the changes, I don't see why that would help or why it's better than #3787. Closing.

PR #4192, opened 2020-06-23
Setup Wizard: enable IPL if enabling Icinga DB

Al2Klimov:


nilmerg:
As mentioned, this issue only applies to installations running v1.0-rc1 of Icinga DB Web. Since this version is not supported at all in the wizard but the current master is, I'l close this here.

PR #4198, opened 2020-07-02
CLI: write errors to stderr

Al2Klimov:


library/Icinga/Cli/Screen.php
```diff
@@ -94,7 +94,7 @@ public function colorize($text, $fgColor = null, $bgColor = null)
     public static function instance()
     {
         if (self::$instance === null) {
-            if (function_exists('posix_isatty') && posix_isatty(STDOUT)) {
+            if (function_exists('posix_isatty') && posix_isatty(STDERR)) {
```
nilmerg, line 97: This breaks colored output if stderr is being redirected but stdout is not. Please make `Screen` configurable so that the log output that uses colors defines that it wants to use stderr. stdout must be the default.

nilmerg (changes requested):

Al2Klimov:
Bildschirmfoto 2020-11-24 um 17 09 29

nilmerg (approved):

PR #4217, opened 2020-08-06
Url::fromPath(): var_export() input value to complain about

Al2Klimov:
... not to hide its type.

### DIFF
--- a/library/Icinga/Web/Url.php
+++ b/library/Icinga/Web/Url.php
@@ -166,8 +166,8 @@ class Url
 
         if (! is_string($url)) {
             throw new ProgrammingError(
-                'url "%s" is not a string',
-                $url
+                'url %s is not a string',
+                var_export($url, true)
             );
         }
 


Al2Klimov:
```
➜  icingaweb2 git:(bugfix/url-programmingerror) php -a
Interactive shell

php > echo var_export(null, true);
NULL
php > var_dump(var_export(null, true));
string(4) "NULL"
php > var_dump(var_export(1, true));
string(1) "1"
php > var_dump(var_export("", true));
string(2) "''"
php > var_dump(var_export([], true));
string(9) "array (
)"
php > var_dump(var_export((object)[], true));
string(17) "(object) array(
)"
```

Al2Klimov:
Commit based on v2.8.0 for eventual backporting via merge.

nilmerg:
Please use the master as base, this won't be backported as it's not a critical bug or anything.

nilmerg (approved):

Issue #4211, opened 2020-07-16
Host/Service actions with only a label are not shown


Arukadox:
## Describe the bug
Hi, recently I've updated my RHEL 7.5 installation to 7.8 along with icinga2 from 2.9 to 2.11 and icingaweb2 to 2.8.1-1.
Most of the update went like a charm but I've noticed that right now I'm getting error under "Problem handling" field on each service in icinga web.
**Problem handling
Icinga\Exception\ProgrammingError in /usr/share/php/Icinga/Web/Url.php:168 with message: url "" is not a string**
I've tried to search for solution for many hours and ended up digging in php-fpm installation etc. but without any luck.
So any solution or workaround would be greatly appreciated.

## Screenshots
![image](https://user-images.githubusercontent.com/36514789/87725400-89087280-c7bd-11ea-8a2a-7b1fb9d20f5b.png)

## Your Environment

```
[root@icinga]# icinga2 --version
icinga2 - The Icinga 2 network monitoring daemon (version: 2.11.4-1)

System information:
  Platform: Red Hat Enterprise Linux Server
  Platform version: 7.8 (Maipo)
  Kernel: Linux
  Kernel version: 3.10.0-1127.13.1.el7.x86_64
  Architecture: x86_64

Build information:
  Compiler: GNU 4.8.5
  Build host: runner-ltrjqz9n-project-322-concurrent-0

[root@icinga]# icingacli --version
Icinga Web 2  2.8.1
Git Commit    233bd29e4104125b4e5ef631e8c16dde33dadd9a
PHP Version   7.3.11
MODULE       VERSION
audit        1.0.1
director     1.7.0
fileshipper  1.1.0
incubator    0.5.0
ipl          v0.3.0
monitoring   2.8.1
pnp          1.1.0
reactbundle  0.7.0
vspheredb    1.1.0

[root@icinga]# php -v
PHP 7.3.11 (cli) (built: Oct 31 2019 08:30:29) ( NTS )
```

Kind Regards
Michal

nilmerg:
Hi,

That's probably a module. Disable the modules one by one and check if it disappears after disabling a particular one. This may already allow you to track the issue down, if not, head back here.

Arukadox:
Hello I've tried to disable modules one by one and even all of them leaving only monitoring module active and this error is still there.

Al2Klimov:
Hello @Arukadox and thank you for reporting!

Please share the error message in your current setup + #4217.

Best,
AK

[...]

PR #4218, opened 2020-08-06
RecursiveNavigationRenderer#__toString(): show whole stacktrace on error

Al2Klimov:


Al2Klimov:
https://github.com/Icinga/icingaweb2/issues/4211#issuecomment-669816909

Al2Klimov:
Commit based on v2.8.0 for eventual backporting via merge.

nilmerg:
Please use master as base. I also don't think showing the stacktrace is necessary so please only log the exception as error.

Al2Klimov:
We're already showing just the exception. Why do you think this isn't neccessary?

nilmerg:
Because it clutters the view. The single line `describe` produces suffices to show the user something is wrong. The stacktrace doesn't provide any more advantage. If necessary one can then take a look in the log and the full stacktrace is available.

nilmerg:
Oh, and please stop the assignee ping pong. Mentioning me is enough if I'm not participating already.

Al2Klimov:
> The stacktrace doesn't provide any more advantage.

Fine...

PR #4236, opened 2020-09-29
Add notification beads for disabled active checks

rjhenry:
We mistakenly disabled active checks in our Icinga environment and did
not notice for some time. This commit hooks into the badge mechanic used
when notifications are turned off, and also provides beads when active
host checks and active service checks are disabled.

### DIFF
--- a/modules/monitoring/library/Monitoring/Web/Navigation/Renderer/BackendAvailabilityNavigationItemRenderer.php
+++ b/modules/monitoring/library/Monitoring/Web/Navigation/Renderer/BackendAvailabilityNavigationItemRenderer.php
@@ -28,7 +28,12 @@ class BackendAvailabilityNavigationItemRenderer extends BadgeNavigationItemRende
             ->select()
             ->from(
                 'programstatus',
-                array('is_currently_running', 'notifications_enabled')
+                array(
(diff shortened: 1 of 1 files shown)


Al2Klimov (dismissed):

modules/monitoring/library/Monitoring/Web/Navigation/Renderer/BackendAvailabilityNavigationItemRenderer.php
```diff
@@ -59,6 +64,16 @@ public function getCount()
             }
             $count = 0;
             $titles = array();
+            if (! (bool) $programStatus->active_host_checks_enabled) {
+                $count = 1;
```
nilmerg, line 68: I know, you just did what the previous implementation also did. But don't you think increasing `$count` with each case is a reasonable change as well? Multiple titles are already shown, though the count is always `1` :thinking:

rjhenry: Good call - I've changed this now.

nilmerg (approved):

PR #4243, opened 2020-10-22
Bufgix/safari service grid not scrollable

flourish86:


nilmerg:
The css doesn't touch anything servicegrid related, this must mean that not only the grid suffered from this bug but also other horizontally overflowing containers. Did you have this in fact with other containers? Did you test this with other overflowing ones?

flourish86:
> The css doesn't touch anything servicegrid related, this must mean that not only the grid suffered from this bug but also other horizontally overflowing containers. Did you have this in fact with other containers? Did you test this with other overflowing ones?

No, didn't find any other issues regarding the overflow.

flourish86:
This was solved otherwise, so I'll close.

PR #4265, opened 2020-11-04
Shared navigation url bug

sukhwinder33445:
**The bug**

The self created navigation URL was displayed in plain text as you can see on the screenshot.

Screenshot 2020-11-05 at 11 36 31

**Why this happend?**

This happened because the url was not escaped correctly for html special characters and if there was a special character in the url, it was displayed as plain text.

We have fixed this bug and also made changes in other places where such behavior might occur.

### DIFF
--- a/library/Icinga/Web/Url.php
+++ b/library/Icinga/Web/Url.php
@@ -866,6 +866,6 @@ class Url
      */
     public function __toString()
     {
-        return $this->getAbsoluteUrl('&amp;');
+        return htmlspecialchars($this->getAbsoluteUrl(), ENT_COMPAT | ENT_SUBSTITUTE | ENT_HTML5, 'UTF-8', true);
     }
 }


nilmerg:
Please include what this will fix in the description.

lippserd:
ref/IP/28425

nilmerg (changes requested):
Just looked at it. I'd rather change `Icinga\Web\Url::__toString()` so that it returns a fully encoded url, not just encoded ampersands. This way any future usage does not fall into the same pit and it covers all of your changes as well already.

It's also safe to adjust Url's string casting, purely due to the fact that ampersands are already encoded there since forever. I've already used `getAbsoluteUrl()` directly due to that many times when I didn't wanted an encoded url.

Though, I'm not a fan of having this encoding in the `Icinga\Web\Url` class as it isn't its responsibility. But again, due to the ampersand being encoded this was always the case. Just not that broad so I think this is the easier fix.

nilmerg (approved):
LGTM!

PR #4272, opened 2020-11-11
Add library support

### LINKED ISSUE #4271: Library Support
Our latest collection of projects (ipl-web, ipl-sql, ...) is bundled in a module (https://github.com/Icinga/icingaweb2-module-ipl). The same applies to third party libraries which are bundled in https://github.com/Icinga/icingaweb2-module-reactbundle.

This integration of libraries into Icinga Web 2 is far from ideal. Their CSS is isolated and can't be used outside of the bundle. Hence we implemented support for assets in #3961. The problem with this however is, that static assets are not supported. Also, these are in fact libraries which are more often than not an integral part of another module or even Icinga Web 2 itself. The moment Icinga Web 2 depends on a module however, lead to this issue.

We need to support true libraries. Completely detached from our module integration.

This means, that a library...
* is placed in a path identified by an environment variable `ICINGAWEB_LIBDIR` (i.e. `ICINGAWEB_LIBDIR[=/usr/share/php-Icinga]`)
* is loaded by the bootstrapper and later on available via a singleton (e.g. `Library::has($name)` and `Library::getVersion($name)`)
* must be a composerized project and has to provide a `composer.json` and `vendor/autoload.php`

Icinga Web 2 then...
* bundles a library's CSS the same way as themes in terms of scope, they'll need to be loaded first however, in order to override their rules partially if desired
* bundles a library's JS in a way that mimics *requireJS* (which allows libraries to provide JS rather independent from a specific framework)
* delivers static assets from libraries from a special route (e.g. `assets/ipl/<asset-path>`)

## Nice to have

Modules should be able to define their dependencies to libraries with version constraints. The same must be possible for Icinga Web 2 itself. These dependencies must then be validated and reported if not fulfilled.

nilmerg:
## What is a library anyway?

* A library is a composer `project` which is independent from other libraries and thus provides all requirements by itself
* A composer `library` is **not** a library in this case as it doesn't ship its own dependencies and is not supported

## What is expected from a library?

* Its name must be part of `composer.json` (Hence the name of the directory on disk is not significant)
* Its version must be part of `composer.json` or alternatively to be found in `VERSION`
* It must provide its own autoloader in `vendor/autoload.php` if it wants to provide PHP
* CSS, JS and static assets need to be placed in `asset/css`, `asset/js` and `asset/static` respectively

## How is it integrated in Icinga Web 2?

* Libraries are searched in paths provided by the environment variable `ICINGAWEB_LIBDIR`
  * Multiple paths need to be separated by colon
  * If it doesn't exist, the default path is `/usr/share/php-icinga`
* It is registered during bootstrapping, the PHP autoloader is implicitly loaded
* CSS is **not** scoped in any way. However, it is loaded first so other CSS can more easily override it
* JS is also not scoped, though there's limited support for *requireJS*'s `define()` (Not supported is [Sugar](https://requirejs.org/docs/whyamd.html#sugar))
  * A byproduct of that is that this is now also supported for modules
* Static assets are available at `lib/<lib-name>/<asset-path>`
  * CSS requiring static assets has to use a LESS variable for the base URL which then Icinga Web 2 or a module need to define
* Modules requiring libraries can't be enabled if not all library dependencies are met
  * A byproduct of that is that this also applies to module dependencies
* Errors caused by missing classes or functions of libraries/modules now also include a hint if a module has unmet dependencies

## Any more byproducts?

There's now a new field for module meta data: `Requires`

### Syntax

`Requires: <lib-name>[ (<ver>)][, <lib-name>[ (<ver>)]] ...`

or
```
Requires:
  Libraries: <name>[ (<ver>)][, <name>[ (<ver>)]] ...
  Modules: <name>[ (<ver>)][, <name>[ (<ver>)]] ...
```

### Example

```
Module: icingadb
Version: 1.0.0-rc1
Requires:
  Libraries: ipl (>=0.5.0)
  Modules: monitoring (>=2.8.0)
Description: Icinga DB Web
  UI for Icinga DB – Provides a graphical interface to your Icinga monitoring
```

### About page

Screenshot from 2020-11-13 12-27-28

### Module page

Screenshot from 2020-11-16 09-20-28

resolves #4271

### DIFF
--- a/application/clicommands/ModuleCommand.php
+++ b/application/clicommands/ModuleCommand.php
@@ -90,17 +90,21 @@ class ModuleCommand extends Command
     /**
      * Enable a given module
      *
-     * Usage: icingacli module enable <module-name>
+     * Usage: icingacli module enable <module-name> [--force]
      */
     public function enableAction()
     {
         if (! $module = $this->params->shift()) {
             $module = $this->params->shift('module');
         }
+
+        $force = $this->params->shift('force', false);
+
         if (! $module || $this->hasRemainingParams()) {
             return $this->showUsage();
         }
-        $this->modules->enableModule($module);
+
+        $this->modules->enableModule($module, $force);
     }
 
     /**
--- a/application/controllers/AboutController.php
+++ b/application/controllers/AboutController.php
@@ -12,6 +12,7 @@ class AboutController extends Controller
     public function indexAction()
     {
         $this->view->version = Version::get();
+        $this->view->libraries = Icinga::app()->getLibraries();
         $this->view->modules = Icinga::app()->getModuleManager()->getLoadedModules();
         $this->view->title = $this->translate('About');
         $this->view->tabs = $this->getTabs()->add(
--- a/application/controllers/ConfigController.php
+++ b/application/controllers/ConfigController.php
@@ -168,7 +168,7 @@ class ConfigController extends Controller
             $form->handleRequest();
         } catch (Exception $e) {
             $this->view->exceptionMessage = $e->getMessage();
-            $this->view->moduleName = $form->getValue('name');
+            $this->view->moduleName = $form->getValue('identifier');
             $this->view->action = 'enable';
             $this->render('module-configuration-error');
         }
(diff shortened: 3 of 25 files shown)


nilmerg:
Now also deprecates #3961

lippserd (dismissed):
Very, very nice 👍

PR #4276, opened 2020-11-16
Fix handling of multi valued ldap attributes

### LINKED ISSUE #4213: LDAP handling of mulitple UID
LDAP authentication should be expanded to handle data that could be entered into a LDAP server by default. This includes the ability to handle when an user has multiple UID set. UID by default allows for multiple values in LDAP servers [LDAP UID IETF](https://tools.ietf.org/html/rfc4519#page-18). 

Icinga LDAP implementation currently expects one to change their LDAP configuration to only allow a single value. If multiple values are found it will cause an error. The error message is "strtolower() expects parameter 1 to be string, array". This also applies to groups as well with their identifier, which most commonly is set to CN.

One should not have to change their LDAP configuration to match the software using it. But the software should conform to the LDAP standards as defined.

nilmerg:
fixes #4213


vijayasaradhi-danda:
We applied the changes manually to suggested files. we don't see exception happening any more

PR #4293, opened 2020-12-07
Drop new parameter from ActionController::setAutorefreshInterval()

nilmerg:
We can't change its interface because the Director [overrides it](https://github.com/Icinga/icingaweb2-module-director/blob/master/library/Director/Web/Controller/ActionController.php#L132-L143) to accomplish something that could have also been patched into Icinga Web 2, but nvm..

The alternative solution for controllers not wanting to be subjected to the user's preferences is the one proposed here. Though, it requires to bypass the setter, of which I'm not a huge fan of. Though, it's the only *not-awkward* (without an over-complicated interface) solution I have.

I'm open for other alternatives.

### DIFF
--- a/application/controllers/ApplicationStateController.php
+++ b/application/controllers/ApplicationStateController.php
@@ -17,6 +17,8 @@ class ApplicationStateController extends Controller
 {
     protected $requiresAuthentication = false;
 
+    protected $autorefreshInterval = 60;
+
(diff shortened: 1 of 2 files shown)


lippserd (changes requested):
The solution looks good to me. But could we please document this behavior in the method and property?

PR #4289, opened 2020-12-02
Support PHP 8

Al2Klimov:
fixes #4287

PR #4298, opened 2020-12-21
Update JShrink for PHP 8 support.

sebastic:
As mentioned in https://github.com/Icinga/icingaweb2/pull/4289#issuecomment-747896920, JShrink needs to be updated for PHP 8 as well.

Because there is no release with those changes yet, a commit ref is used as version.

### DIFF
--- a/library/vendor/JShrink/Minifier.php
+++ b/library/vendor/JShrink/Minifier.php
@@ -39,6 +39,13 @@ class Minifier
     protected $input;
 
     /**
+     * Length of input javascript.
+     *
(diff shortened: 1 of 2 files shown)


sklaes:
@sebastic JShrink 1.4 with PHP support has been released:

https://github.com/tedious/JShrink/releases/tag/v1.4.0

sebastic:
`SOURCE` has been updated to use the 1.4.0 version instead, but `Minifier.php` has not been changed since the commit that was used before.

PR #4305, opened 2021-01-22
IDO: Add config option to use the customvariables table for fetching …

xam-stephan:
workaround for icinga2 issue #5621

same solution as
https://github.com/Icinga/icingaweb2/commit/1ed2ebc1916d6d83a62cf649de0f5b981c27fc4b

This commit introduces a config option to use the customvariables table instead:
/etc/icingaweb2/modules/monitoring/config.ini

[ido]
use_customvar_status_table = 0

### DIFF
--- a/modules/monitoring/library/Monitoring/Backend/Ido/Query/IdoQuery.php
+++ b/modules/monitoring/library/Monitoring/Backend/Ido/Query/IdoQuery.php
@@ -8,6 +8,7 @@ use Zend_Db_Expr;
 use Icinga\Application\Icinga;
 use Icinga\Application\Hook;
 use Icinga\Application\Logger;
+use Icinga\Application\Config;
 use Icinga\Data\Db\DbQuery;
 use Icinga\Data\Filter\Filter;
 use Icinga\Data\Filter\FilterExpression;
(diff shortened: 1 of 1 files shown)


nilmerg:
Hi, I can see the reasoning behind this change, but the issue https://github.com/Icinga/icinga2/issues/5621 doesn't seem to be related?

xam-stephan:
not directly.
The problem is, that some objects has no custom vars / the filters does not work, if some entries are missing.

-----

mysql> SELECT COUNT(*) FROM icinga_customvariables cvs
    ->   LEFT JOIN icinga_objects o ON o.object_id = cvs.object_id
    ->    AND o.is_active = 1
    ->  WHERE o.object_id IS NULL;

SELECT COUNT(*) FROM icinga_customvariablestatus cvs
  LEFT JOIN icinga_objects o ON o.object_id = cvs.object_id
   AND o.is_active = 1
 WHERE o.object_id IS NULL;
+----------+
| COUNT(*) |
+----------+
|   **121446** |
+----------+
1 row in set (1.65 sec)

-----

mysql> SELECT COUNT(*) FROM icinga_customvariablestatus cvs
    ->   LEFT JOIN icinga_objects o ON o.object_id = cvs.object_id
    ->    AND o.is_active = 1
    ->  WHERE o.object_id IS NULL;
+----------+
| COUNT(*) |
+----------+
|   **121199** |
+----------+
1 row in set (1.04 sec)

lippserd:
In Icinga 2 we have problems with config hashes after upgrading. This should be fixed with version 2.12.1.
In order to reset the hashes, run the following statements in the Icinga 2 database and restart Icinga 2 afterwards:

```
UPDATE icinga_hosts SET config_hash=NULL;
UPDATE icinga_services SET config_hash=NULL;
```

I would not try to fix this in Web 2.

nilmerg:
Ah, so you're not using Icinga 1.x or disabled table `customvariablestatus` somehow. I expected that to be the case, not another issue with Icinga 2 for which this config option is not designed for.

So, yeah, that's definitely something that doesn't need a fix in Icinga Web 2, but in Icinga 2.

xam-stephan:
ok, then we have to plan a change to update to 2.12.x   (current 2.11.4)

PR #4306, opened 2021-01-22
Authorization enhancements

### LINKED ISSUE #3885: possibility to negate permissions in role configuration
There are use cases where it is desired to negate a single permission out of a wider group.

For example I want to block a role from editing roles authorisation-related settings (users, groups, roles, ...) but I still want to allow viewing them because it really helps debugging permissions-related problems and also give them full powers on everything else.

Right now I have to explicitly look in the code for all "siblings" under `config/` and allow them.

Full access to everything would be:
```
[admins]
users = "foo, bar, baz"
permissions = "*"
```
And to implement what I described above needs something like (newlines for readability):
```
[admins]
users = "foo, bar, baz"
permissions = "config/application/*,
               config/authentication/groups/show,
               config/authentication/roles/show,
               config/authentication/users/show,
               config/modules/*,
               admin,
               application/*,
               grafana/*,
               module/*,
               monitoring/*"
```

The problem is that I must look for all the permissions that would be included by the simple `"*"` and explicitly set them (this usually means `fgrep -r assertPermission` in all the modules I'm using), also I'll have to keep track of any new permission introduced in future versions or any module that I'll add later.

I would like a way to negate permissions (similar to what is already in place for `monitoring/filter/objects` restrictions) so I could achieve what I want with something like:
```
[admins]
users = "foo, bar, baz"
permissions = "*,
               !config/authentication,
               config/authentication/groups/show,
               config/authentication/roles/show,
               config/authentication/users/show"
```

In my example the logic would be "the last match wins", which IMHO is desirable because:
* it's completely backward compatibile with the current logic
* it's really similar to the current logic in how it is expressed (every element of the comma-separated list adds something)
* introduces just a single new operator (the `!` for negation)
* supports easy negations: `*` allows everything, but then `!config/authentication` blocks authentication-related features
* supports exceptions: `!config/authentication` blocks auth-related features, but then `config/authentication/roles/show` allows read-only access to roles for easy debugging

## Describe alternatives you've considered
I'm not aware of a real alternative, currently I have to explicitly list all permissions (like in the example above) but that does not account for any new permission namespace introduced in future changes (which an `*` would have taken care of).

## Additional context
My example is on security-related functions but the request could be easily rewritten for a lot of different use cases. The classic pattern I keep hearing is "leave full access by default but make just `$something` read-only" and it would greatly benefit from this proposed feature.

### LINKED ISSUE #4270: Allow for inheritance in roles
A customer of ours uses a very fine grained permission set and has lots of users access different parts of the monitored objects. Every time, they change something, install a new module or an update brings new options they have to go through a whole lot of roles to reflect these changes. More so, if they want to make a change temporarily it's the same, twice.

I'd like to see roles that can inherit from other roles. This way new settings, temporary changes or changes in the company permission structure can more easily be managed.

Since roles have a representation in config files or the database we could use a script to apply the changes. Not only is that cumbersome but I really don't want to have an error prone script mess with permissions directly in the backend. Setting permissions via API could be an option but to be honest I didn't check if that's possible after all because it's just a small bit less complicated than making changes directly in the backend.

nilmerg:
Supplements https://github.com/Icinga/icingadb-web/pull/158 and adds the following features:

* **Role refusals**
  The picture below shows that all command permissions are granted. Though, not those to add or remove comments.
Screenshot from 2021-02-01 15-25-59
* **Role inheritance**
  No picture. This has only an effect behind the scenes. Single inheritance only. No special behavior, memberships are just shared between roles this way.
* **Unrestricted Roles**
  A new option in addition to `Administrative Access`. If enabled, this causes the restrictions of the role and any other one to be ignored.
* **Password changes not allowed by default anymore**
  The fake refusal `no-user/password-change` has now been changed to a grant `user/password-change`. Any user that had `no-user/password-change` previously still cannot change passwords. Though any user that didn't have this *permission*, needs to be granted `user/password-change` now in order to change passwords.
* **Restructured general permissions**
  Most of the general permissions were re-structured so that they're less cluttered and relate better to their actual purpose. There's no migration required. Any previous role configuration is still valid. Once adjusted in the UI though, it's automatically migrated to the new format.
Screenshot from 2021-02-18 09-35-41

resolves #3885 
resolves #4270

### DIFF
--- a/application/controllers/AccountController.php
+++ b/application/controllers/AccountController.php
@@ -43,7 +43,7 @@ class AccountController extends Controller
         $config = Config::app()->getSection('global');
         $user = $this->Auth()->getUser();
         if ($user->getAdditional('backend_type') === 'db') {
-            if ($user->can('*') || ! $user->can('no-user/password-change')) {
+            if ($user->can('user/password-change')) {
                 try {
                     $userBackend = UserBackend::create($user->getAdditional('backend_name'));
                 } catch (ConfigurationError $e) {
--- a/application/controllers/AnnouncementsController.php
+++ b/application/controllers/AnnouncementsController.php
@@ -61,7 +61,7 @@ class AnnouncementsController extends Controller
      */
     public function newAction()
     {
-        $this->assertPermission('admin');
+        $this->assertPermission('application/announcements');
 
         $form = $this->prepareForm()->add();
         $form->handleRequest();
@@ -73,7 +73,7 @@ class AnnouncementsController extends Controller
      */
     public function updateAction()
     {
-        $this->assertPermission('admin');
+        $this->assertPermission('application/announcements');
 
         $form = $this->prepareForm()->edit($this->params->getRequired('id'));
         try {
@@ -89,7 +89,7 @@ class AnnouncementsController extends Controller
      */
     public function removeAction()
     {
-        $this->assertPermission('admin');
+        $this->assertPermission('application/announcements');
 
         $form = $this->prepareForm()->remove($this->params->getRequired('id'));
         try {
--- a/application/controllers/ConfigController.php
+++ b/application/controllers/ConfigController.php
@@ -35,24 +35,33 @@ class ConfigController extends Controller
     public function createApplicationTabs()
     {
         $tabs = $this->getTabs();
-        $tabs->add('general', array(
-            'title' => $this->translate('Adjust the general configuration of Icinga Web 2'),
-            'label' => $this->translate('General'),
-            'url'   => 'config/general',
-            'baseTarget' => '_main'
-        ));
-        $tabs->add('resource', array(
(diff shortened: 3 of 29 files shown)


chas0rde:
Hello
I just realized that this breaks https://github.com/Icinga/icingaweb2/pull/3553/commits/6ad74ef1cccfda8bb462e3ececd2a625085f1740 if I am not mistaken
We use this feature to allow host/service-var based filtering of what a user sees in monitoring.
The removal of the code from above merge also removes this option (unless I fail to see how it needs to be configured after your change)

Our filter:
_host_responsible=$user:local_name$|_host_deputy=$user:local_name$|_host_informational_users=*$user:local_name$*|_service_responsible=$user:local_name$|_service_deputy=$user:local_name$|_service_informational_users=*$user:local_name$*

lippserd:
Hi,

The code was just moved. If it does not work for you, please open a dedicated issue.

All the best,
Eric

PR #4318, opened 2021-02-17
WIP

Yonas-net:
This PR loads all available dashboards the active Icinga Web 2 user has access to from the database and appends them to the dashboards loaded from an ``ini`` file. Just almost everything is hard coded but might work something like this.


nilmerg:
Hey Yonas,

just noticed that it's not only the name of the PR but also the branch name which has no real name. Please don't push branches or create PRs on here with just *WIP* as the name.

That's a very ambiguous name and if even the branch is called this way, no-one knows for what it is and from whom it is. If you want to mark a PR as work in progress, create it as a draft.

So, please push your changes in a branch with a proper name and then create a new PR also with a proper name. Thank you!

Yonas-net:
OK, Will be done.

Yonas-net:
Replaced by #4323

Issue #4322, opened 2021-02-23
Poor SQL query performance ("CASE WHEN" considered harmful)


Bu-Ble:
For our large Icinga environment (approx. 100k host objects, 1M services) I am evaluating Icingaweb2 with MariaDB as a replacement for Thruk (with LMD, without DB).
The Icingaweb UI seems to be quite slow, so I enabled the slow_query_log and found a few slow SQL queries (10-20 seconds for each execution).

One of the most frequent queries is:
```
MariaDB [icinga]> SELECT COUNT(*) AS services_critical_unhandled
    -> FROM (
    ->     SELECT so.name2 AS service_description
    ->     FROM icinga_objects AS so
    ->     INNER JOIN icinga_services AS s
    ->     ON s.service_object_id = so.object_id AND so.is_active = 1 AND so.objecttype_id = 2
    ->     INNER JOIN icinga_hoststatus AS hs
    ->     ON hs.host_object_id = s.host_object_id
    ->     INNER JOIN icinga_servicestatus AS ss
    ->     ON ss.service_object_id = so.object_id
    ->     WHERE (
    ->         CASE WHEN (ss.problem_has_been_acknowledged + ss.scheduled_downtime_depth + COALESCE(hs.current_state, 0)) > 0 THEN 1 ELSE 0 END = 0
    ->         AND CASE WHEN ss.has_been_checked = 0 OR ss.has_been_checked IS NULL THEN 99 ELSE ss.current_state END = 2
    ->     )
    -> ) AS problems;
+-----------------------------+
| services_critical_unhandled |
+-----------------------------+
|                          77 |
+-----------------------------+
1 row in set (10.88 sec)
```

I am not sure if I understood the WHERE CASE WHEN clause correctly, but rewriting it as simple WHERE conditions resulted in a 1000 times faster execution with the same result:

````
MariaDB [icinga]> SELECT COUNT(*) AS services_critical_unhandled
    -> FROM (
    ->     SELECT so.name2 AS service_description
    ->     FROM icinga_objects AS so
    ->     INNER JOIN icinga_services AS s
    ->     ON s.service_object_id = so.object_id AND so.is_active = 1 AND so.objecttype_id = 2
    ->     INNER JOIN icinga_hoststatus AS hs
    ->     ON hs.host_object_id = s.host_object_id
    ->     INNER JOIN icinga_servicestatus AS ss
    ->     ON ss.service_object_id = so.object_id
[...]

PR #4324, opened 2021-02-24
Increase SQL performance by avoiding "case when"

sklaes:
Only partially resolves the issue
refs #4322


sklaes:
@nilmerg may i ask you to review this pull request - Thanks!

nilmerg:
@lippserd will do so if he got the time. I'm currently busy with non-db stuff and can't wrap my head around this as well.

lippserd:
Hi @sklaes,

Do you have benchmarks that prove that avoiding case when is actually faster?
I did a quick test and found no noticeable changes.

I'm not quite sure what ANSI SQL-92 defines for aggregate functions with boolean expressions, e.g. `SUM(boolean)`
but it seems this only works in MySQL.

All the best,
Eric

sklaes:
Hey @lippserd,
as @Bu-Ble already pointed out in #4322 there is a huge penalty by using **CASE WHEN** in SQL, as it is forcing a full table scan.
His optimization - directly in the sql querys - "resulted in a 1000 times faster execution".
I was not 100% sure how the meta language in php transforms in actual SQL queries, so I only optimized the “SUM(CASE WHEN […]” queries. I have implemented them one day ahead of the pull request in our production environment (using MariaDB 5.5.68) and I do not have any complaints.
In my environment the most impressive improvement was the loading time of contactgroups. Previously I was unable to load them at all (timeout) and now it loads in about 3 seconds:
image

I haven’t tested the it on any other database, so it may not be the correct solution. But I can tell for sure that there is a performance increase by avoiding **CASE WHEN’**

leeclemens:
I think it's worth pointing out that the "huge penalty' is because it is not using indexes (at least in the referenced issue's specific example/comparison). I'm not familiar with any issue with the `CASE` function in general and some searching didn't yield any blogs, etc saying to always avoid that function.

That said, if converting from `CASE` to `WHERE` will enable the use of indexes - that's may be a good thing (or create appropriate indexes). However, that would need to be evaluated for each query being changed individually.

PR #4337, opened 2021-03-18
Feature/icinga dark theme login

### LINKED ISSUE #4394: Solarized dark theme: reconsider text-background contrast
## Describe the bug

At least unhandled critical list items and menu badges (black on red) and success notifications (didn’t-figure-out on yellow) are hard to read.

## To Reproduce

1. Switch to the solarized dark theme
2. List unhandled criticals
3. Have a look at "CRITICAL since ..." and the menu badges
4. Ack one
5. Have a look at the notification at the bottom

## Expected behavior

White on red and (let's say) black on yellow.

## Screenshots

Bildschirmfoto 2021-06-19 um 01 33 31

## Your Environment

Latest stable in Docker container.

## Additional context

Wanted to take care of my eyes (dev after all), switched my IW2 theme and... was disappointed.

flourish86:
fixes #4394


nilmerg (changes requested):
Adjusted a few things. Style is not changed. Only the login in the light theme is fixed and now resembles the one from dark mode completely.

Most of the themes look okay, the high contrast theme and the colorblind theme though, not quite.

@flourish86 Please fix the high contrast theme.
@theFeu Please take a look at the colorblind theme with these changes. I don't know what's wrong or right there. You should, you've made this initially after all. :smiley:

nilmerg:
Oh, and what I've just noticed, the setup wizard is also broken.

flourish86:
> @flourish86 Please fix the high contrast theme.

I think, it's to discuss, what we exactly want here.

The main problem here seems to be, that the text-colors are hard coded values that are still based on the old theme.

So there are basically two options, as I see it:

1) We adjust the text-colors and keep it based on the dark theme
2) We'll have a completely different, means new, theme optimized for "real" high contrast appearance.

I'm not sure at this point what the original foundation for the high contrast theme were. Does it follow any standards or guidelines or was it just made to provide better contrast?

nilmerg:
It's probably based on https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html

flourish86:
Setup Wizard is fixed with 4147e4ef

flourish86:
@theFeu will take care of the colorblind theme

For the high contrast theme, we should reset it to the given one, since we had to do design work for making it work with the new colors.

One thing I learned while talking to @theFeu was that it has to be a light theme and the contrast values have to be at least a certain value (probably 85%).

Even if we used the light theme as a basis, we'd have the code overhead anyway. So why not stick to the working theme, since most high contrast users theme will probably have no real benefit from a fancy new theme.

theFeu (changes requested):
As discussed with Flo, for the colourblind theme we need the unhandled states to always have a white font.
image


I also found that in the tactical overview the there 0 in the middle is hard to read: 
image

The service list also doesn't apply the light green properly to all elements:
image

flourish86:
The high-contrast theme adjustments should now be completed with a6a4692.

flourish86:
Fixed further details for high-contrast.

9f94a2a
c144818

nilmerg:
"Dark theme" should also include the following for the datetime picker: https://github.com/Icinga/icingaweb2/blob/master/public/css/themes/solarized-dark.less#L60-L85

And subsequently, the light theme needs the default style...

public/css/icinga/main.less
```diff
@@ -196,7 +196,7 @@ a:hover > .icon-cancel {
 }
 
 .name-value-table > tbody > tr > th {
-  color: @text-color-light;
+  color: @gray;
```
nilmerg, line 199: The control-label-group of the object features now use `@text-color-light` instead of `@gray`. This now does the exact opposite, can you please decide which to use in both cases?

nilmerg (changes requested):

application/forms/Authentication/LoginForm.php
```diff
@@ -27,6 +27,13 @@ class LoginForm extends Form
      */
     const REDIRECT_URL = 'dashboard';
 
+    public static $defaultElementDecorators = [
+        ['ViewHelper', ['separator' => '']],
+        ['Help', []],
+        ['Errors', ['separator' => '']],
+        ['HtmlTag', ['tag' => 'div', 'class' => 'control-group']]
+    ];
```
nilmerg, line 35: We can't do that. Due to two things:

a) `.control-group` is a structural class and not responsible for styling, hence should never be used for only styling reasons and here it is.
b) Changing the markup of the login form surely breaks third party themes in a big way. If it's only done to re-use our form styles, either make the form styles less specific or copy the style.

flourish86: I removed the .control-group elements and adjusted the styles with 86f8f0d.
✅

public/css/icinga/main.less
flourish86: Changed to `@text-color-light` in both cases with dc5a427.

Also see https://github.com/Icinga/icingadb-web/pull/216/commits/e584755f485e05c298b19725ecd8c148ee637cc5

flourish86: ✅

application/forms/Authentication/LoginForm.php
```diff
@@ -27,6 +27,12 @@ class LoginForm extends Form
      */
     const REDIRECT_URL = 'dashboard';
 
+    public static $defaultElementDecorators = [
+        ['ViewHelper', ['separator' => '']],
+        ['Help', []],
+        ['Errors', ['separator' => '']],
+    ];
+
```
nilmerg, line 35: You've removed the control-group, but why is this still necessary?

nilmerg (changes requested):
The loading animation doesn't look good now:

Screenshot from 2021-06-21 09-21-11

application/forms/Authentication/LoginForm.php
flourish86: Makes sense. Removing this also solves #4395.

flourish86: ✅

application/forms/Authentication/LoginForm.php
```diff
@@ -50,7 +50,7 @@ public function createElements(array $formData)
                 'autocapitalize'    => 'off',
                 'autocomplete'      => 'username',
                 'class'             => false === isset($formData['username']) ? 'autofocus' : '',
-                'label'             => $this->translate('Username'),
+                'placeholder'             => $this->translate('Username'),
```
nilmerg, line 53: The alignment of `=>` got somehow broken here and below.

nilmerg (approved):

PR #4348, opened 2021-04-12
Add basic module documentation

nilmerg:


nilmerg:
@theFeu Please have a look regarding comprehensibility for not-me users. :laughing:

doc/01-About.md
```diff
@@ -55,15 +42,41 @@ They can be stored either in a database or in .ini files.
 See [here](07-Preferences.md#preferences) for more information about a user's preferences
 and how to configure their storage type.
 
-## Documentation <a id="about-documentation"></a>
+## Modules
+
+Modules extend Icinga Web 2 with additional functionality. They allow the integration of
+capabilities into existing views and even other modules. Whether it's a graph provider such
```
theFeu, line 48: "Whether it's a graph provider such
as [Graphite](https://github.com/Icinga/icingaweb2-module-graphite) or a UI for the Icinga 2
configuration like the [Director](https://github.com/Icinga/icingaweb2-module-director)."

Might want to complete into a full sentence.

doc/01-About.md
```diff
@@ -55,15 +42,41 @@ They can be stored either in a database or in .ini files.
...
+> This module is only for the IDO backend. Use [Icinga DB Web](https://github.com/Icinga/icingadb-web) instead.
+
+This is the core module for most Icinga Web 2 users.
+
+It provides an intuitive user interface for monitoring with Icinga 2.
+There are lots of list and detail views (e.g. for hosts and services)
+you can sort and filter depending on what you want to see.
+
+You can also control the monitoring process itself by sending external commands to Icinga.
+Most such actions (like rescheduling a check) can be done with just a single click.
```
theFeu, line 67: It's probably best to hint where the click is (in the UI / the detail view)

doc/08-Modules.md
```diff
@@ -0,0 +1,68 @@
+# Modules
+
+## Installation
+
+A module should be installed in one of the [configured module paths](03-Configuration.md#general-configuration).
+The default path in most installations is `/usr/share/icingaweb2/modules`.
+
+Each directory in there contains the files for a particular module. The directory's name has to be the one
```
theFeu, line 8: An example would be good here for this case.

doc/08-Modules.md
```diff
@@ -0,0 +1,68 @@
...
+that is provided by the module's documentation. If there is none provided, it is usually the name of the
+module in all lowercase. Some modules may use a name prefixed with `icingaweb2-module-` (e.g. `icingaweb2-module-map`).
```
theFeu, line 10: An example would be good here for this case too.

(e.g. `icingaweb2-module-map` turns into `map`).

doc/08-Modules.md
```diff
@@ -0,0 +1,68 @@
...
+A module may require further installation steps. Whether these need to be performed before enabling the module,
+should be provided by the module's documentation. In any case, don't forget to apply these as well, otherwise
+the module will most likely not function correctly.
+
+### Examples
+
+There are sample installation instructions provided for your convenience:
+
+**Sample Tarball installation**
+
```
theFeu, line 34: It might seem obvious, but do we need to specify where to run the commands?

theFeu (changes requested):
I left some notes, but in general it reads very nicely 👍

doc/01-About.md
nilmerg: > Be it a graph provider such as [Graphite](https://github.com/Icinga/icingaweb2-module-graphite) or a UI for the Icinga 2 configuration like the [Director](https://github.com/Icinga/icingaweb2-module-director).

Better?

doc/01-About.md
nilmerg: It's a introduction, not a usage example. I don't think that suits there.

doc/08-Modules.md
nilmerg: An example for a directory with files in it?

doc/08-Modules.md
nilmerg: No? The only path in there is absolute. If the user will use a relative path, he/she/it should know what that implies. :thinking:

doc/01-About.md
theFeu: I more meant like, that you could just finish the sentence with "can be done with just a single click in the UI." 
Because the way you left it open there puts me under the impression that what follows will answer that implicit mystery and then I was confused when you continued with something else :)

doc/08-Modules.md
theFeu: "So if you want to install the reporting module, it should be installed under `usr/share/icingaweb2/modules/reporting`".

doc/08-Modules.md
theFeu: Fair enough :)

doc/01-About.md
nilmerg: Ah, well, added ...`in the UI` to the end :laughing:

doc/08-Modules.md
nilmerg: Hm, none of the content goes that directly into detail about a specific example. Did it like so: `icingaweb2-module-map` turns into `/usr/share/icingaweb2/modules/map`

theFeu (approved):
Lgtm!

PR #4354, opened 2021-04-15
Add datetime picker widget

### LINKED ISSUE #4107: Ship flatpickr as Date/Time Picker
Our JavaScript should ship and use [flatpickr](https://flatpickr.js.org/) as date/time picker.
The only thing we have to take care about is localisation, i.e. date/time format and translation.
flatpickr ships its own translation files as JavaScript code, which is why this feature depends on #4106.

TODOs:

- [x] Ship flatpickr in vendor
- [x] Provide a JavaScript behavior which enables flaptpickr respecting set data attributes for configuration
- [x] Include JavaScript localisation file according to the set locale
- [x] Set date/time format according to [intl](https://www.php.net/manual/de/intldateformatter.getpattern.php)

nilmerg:
**Default style with default locale in the monitoring module**
Screenshot from 2021-04-16 11-19-24

**Default style with default locale in Icinga DB Web**
Screenshot from 2021-04-16 11-19-01

**Dark style with german locale in Icinga DB Web**
Screenshot from 2021-04-16 11-21-15

blocked by https://github.com/Icinga/ipl-web/pull/15

resolves #4107
refs #3416

### DIFF
--- a/application/forms/PreferenceForm.php
+++ b/application/forms/PreferenceForm.php
@@ -94,6 +94,7 @@ class PreferenceForm extends Form
         $this->preferences = new Preferences($this->store ? $this->store->load() : array());
 
         $oldTheme = $this->preferences->getValue('icingaweb', 'theme');
+        $oldLocale = $this->preferences->getValue('icingaweb', 'language');
 
         $webPreferences = $this->preferences->get('icingaweb', array());
         foreach ($this->getValues() as $key => $value) {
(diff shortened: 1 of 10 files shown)


nilmerg:
flatpickr has a rather hefty memory leak, I don't know if it should be merged if that's not fixed soon :thinking: (https://github.com/flatpickr/flatpickr/issues/1908)

PR #4356, opened 2021-04-20
fixed powertools repo name for CentOS 8

log1-c:
name for the powertools changed to lowercase

### DIFF
--- a/doc/02-Installation.md
+++ b/doc/02-Installation.md
@@ -129,7 +129,7 @@ CentOS 8 additionally needs the PowerTools repository for EPEL:
 
 ```
 dnf install 'dnf-command(config-manager)'
-dnf config-manager --set-enabled PowerTools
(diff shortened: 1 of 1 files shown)


nilmerg:
If I search for this, I find other installation instructions which show both variations. Does this depend on a specific 8.x version or is somehow else dependent on something else? Do you have a source showing the case change, a news, changelog or such?

log1-c:
It seems indeed being bound to specific versions, i.e. versions newer than 8.3.2011, which makes it event more annoying.
https://wiki.centos.org/Manuals/ReleaseNotes/CentOS8.2011#Yum_repo_file_and_repoid_changes.

So maybe having both variants might be the right approach.

nilmerg:
> So maybe having both variants might be the right approach.

Fine, then consider this a change request :grin:

nilmerg (approved):
LGTM!

Issue #2840, opened 2017-05-12
Can't configure custom user and group backends


swizzly:
Hi
Having a Custom-Group-Backend without GUI causes a problem in Icingaweb2 (see below).
Maybe Icingaweb2 should make a small message indicating that the Backend has no GUI, or cannot show the list of groups.
Thanks

Message that currently appears:
Invalid backend type "mybackendname" provided

```
#0 /usr/share/icingaweb2/application/forms/Config/UserGroup/UserGroupBackendForm.php(179): Icinga\Forms\Config\UserGroup\UserGroupBackendForm->getBackendForm('mybackendname')
#1 /usr/share/php/Icinga/Web/Form.php(772): Icinga\Forms\Config\UserGroup\UserGroupBackendForm->createElements(Array)
#2 /usr/share/php/Icinga/Web/Form.php(1097): Icinga\Web\Form->create(Array)
#3 /usr/share/icingaweb2/application/forms/Config/UserGroup/UserGroupBackendForm.php(191): Icinga\Web\Form->populate(Array)
#4 /usr/share/php/Icinga/Web/Form.php(1183): Icinga\Forms\Config\UserGroup\UserGroupBackendForm->onRequest()
#5 /usr/share/icingaweb2/application/controllers/UsergroupbackendController.php(92): Icinga\Web\Form->handleRequest()
#6 /usr/share/icingaweb2/library/vendor/Zend/Controller/Action.php(507): Icinga\Controllers\UsergroupbackendController->editAction()
#7 /usr/share/php/Icinga/Web/Controller/Dispatcher.php(76): Zend_Controller_Action->dispatch('editAction')
#8 /usr/share/icingaweb2/library/vendor/Zend/Controller/Front.php(937): Icinga\Web\Controller\Dispatcher->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
#9 /usr/share/php/Icinga/Application/Web.php(389): Zend_Controller_Front->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
#10 /usr/share/php/Icinga/Application/webrouter.php(109): Icinga\Application\Web->dispatch()
#11 /usr/share/icingaweb2/public/index.php(4): require_once('/usr/share/php/...')
#12 {main}
```

lippserd:
Hi,

[...]

PR #4359, opened 2021-04-21
Fixes Icinga/icingaweb2#2840

gianlucapiccolo:


nilmerg:
Hey, you should update your master and rebase the branch with it, otherwise the checks don't run. (At least I think that's the reason..)

gianlucapiccolo:
> Hey, you should update your master and rebase the branch with it, otherwise the checks don't run. (At least I think that's the reason..)
Hi! yes you're allright, I fixed it.

PR #4365, opened 2021-04-28
Make configuration of custom user backends possible

gianlucapiccolo:
refs #2840

### DIFF
--- a/application/forms/Config/UserBackendConfigForm.php
+++ b/application/forms/Config/UserBackendConfigForm.php
@@ -38,6 +38,13 @@ class UserBackendConfigForm extends ConfigForm
     protected $backendToLoad;
 
     /**
+     * The loaded custom backends list
+     *
+     * @var array
+     */
+    protected $customBackends;
+
+    /**
      * Initialize this form
      */
     public function init()
@@ -45,6 +52,7 @@ class UserBackendConfigForm extends ConfigForm
         $this->setName('form_config_authbackend');
         $this->setSubmitLabel($this->translate('Save Changes'));
         $this->setValidatePartial(true);
+        $this->customBackends = UserBackend::getCustomBackendConfigForms();
     }
 
     /**
@@ -114,6 +122,11 @@ class UserBackendConfigForm extends ConfigForm
                 $form = new ExternalBackendForm();
                 break;
             default:
+                foreach ($this->customBackends as $backendType => $backendData) {
+                    if ($backendType === $type) {
+                        return new $backendData['formClassPath']();
+                    }
+                }
                 throw new InvalidArgumentException(
                     sprintf($this->translate('Invalid backend type "%s" provided'), $type)
                 );
@@ -272,6 +285,10 @@ class UserBackendConfigForm extends ConfigForm
         );
         if ($backendType === 'external' || empty($externalBackends)) {
             $backendTypes['external'] = $this->translate('External');
+        }
+
+        foreach ($this->customBackends as $customBackendType => $backendData) {
+            $backendTypes[$customBackendType] = $backendData['name'];
         }
 
         if ($backendType === null) {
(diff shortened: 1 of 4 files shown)


gianlucapiccolo:
@nilmerg hey, I created a new PR because it was a mess to clean the other old one branch history. Can you please ignore the old one and give a look at this one?

nilmerg (changes requested):
Hey Gianluca,

I'm not quite a fan of how you're retrieving/identifying the custom backend config forms. The class `UserBackendConfigForm` shouldn't require the knowledge about this, it rather should know where to ask for it.

So please implement the static method `Icinga\Authentication\User\UserBackend::getCustomBackendConfigForms()` which returns a map of `$identifier => $formClassPath`. Then use this method in class `UserBackendConfigForm` when required. (Preferably only once, in `UserBackendConfigForm::init()`.)

application/forms/Config/UserBackendConfigForm.php
```diff
@@ -45,6 +52,12 @@ public function init()
         $this->setName('form_config_authbackend');
         $this->setSubmitLabel($this->translate('Save Changes'));
         $this->setValidatePartial(true);
+        $this->setCustomBackend();
+    }
+
+    protected function setCustomBackend()
+    {
+        $this->customBackends = UserBackend::getCustomBackendConfigForms();
```
nilmerg, line 60: Aaah, my eyes bleed :sob: :joy:

Do I understand this correct, that you did that just to avoid the error you otherwise get in the test you've patched as well?

This is bad. Designing code in such a way, that it's testable is never a good thing. Do that without this *setter* directly in the init method.

I've added a suggestion below how to make the test working.

test/php/application/forms/Config/UserBackendReorderFormTest.php
```diff
@@ -17,6 +17,11 @@ public function save()
         self::$newConfig = $this->config;
         return false;
     }
+
+    public function setCustomBackend()
+    {
+        return false;
+    }
```
nilmerg, line 24: Remove that. Setup a mock for the functionality the test requires instead.

Add the following to the test case where the request object is already being mocked:

`Icinga::app()->shouldReceive('getModuleManager->getLoadedModules')->andReturn([]);`

nilmerg (changes requested):

application/forms/Config/UserBackendConfigForm.php
gianlucapiccolo: Yes you're totally right, let me fix this.

test/php/application/forms/Config/UserBackendReorderFormTest.php
gianlucapiccolo: I didn't know that mockery allows to concatenate methods in shouldReceive. Good to know...

nilmerg: It's called demeter chains :grin:

nilmerg (approved):

nilmerg (changes requested):
Hey Gianluca,

sorry for re-visiting this, but I've did the same for user group backends now in #4388.

While doing this, I've initially wanted to do it the same as you did in your PR here. Though then I noticed some problems:

* To get a form, you'll have to instantiate a backend first. You do this with an empty config object. This implies that the form may be delivered based on the backend's config. But by passing an empty config, this isn't even possible. So it's best to make the `getConfigurationFormClass` method static.
* You call `getName()` in order to show a nice name to the user in the dropdown. The name however is what the user configures in the first place. It's not a label of sorts for the backend's identifier. The dropdown can only show the identifier.

So please adjust your changes analogous to mine in #4388.

Thanks!

nilmerg:
Hey Gianluca,

if you don't have time for this, please enable maintainer edits so that I can adjust the code myself. The release of v2.9 is coming closer and I'd like to finish this here. :smiley:

bgroeber:
Hi @nilmerg, 

sorry for the late response.
@gianlucapiccolo is currently unavailable for a few weeks and I guess the maintainer edit flag can only be set by the author, so I re-submit this PR with #4393, and the maintainer edit flag enabled, and should substitute this one.

Thanks for your effort!

nilmerg:
Superseded by #4393

PR #4369, opened 2021-05-11
Format: Add support for more UoMs

Yonas-net:


Yonas-net:
It contains a few changes that I didn't actually make at all.

Yonas-net:
It is no longer needed!

PR #4389, opened 2021-06-08
RELEASE.md: add Docker preparations

Al2Klimov:


nilmerg:
That's [already there](https://github.com/Icinga/icingaweb2/blob/master/RELEASE.md#docker)?

nilmerg:
Or, why should we have a note here that a different repo *may* need updates?

Following this, we'd also need a note that we should ensure monitoring module still works with the IDO.
Or that Icinga DB Web works with Icinga DB.

I see no point in this.

Al2Klimov:
No, at that time you're just replicating the already finalized tag which may have bugs like docker-icingaweb2#41 .

This one encourages you to fix such before the release GHA even run.

nilmerg:
That's too late. Such should happen waaaay before the actual release. Just as it happened today.

Al2Klimov:
Even better. :)

PR #4390, opened 2021-06-11
Feature/auto dark and light mode adjustments#4386

### LINKED ISSUE #4386: Automatic dark/light mode adjustments
With #4337 we'll introduce a dark and light theme. The dark theme is the new default. The light theme is a builtin theme which can be chosen alternatively.

This is crying out for an automatic adjustment on browsers/systems which support it. So let's do it!

For this we need a basis as follows:

* The preferences get a new option to choose one of the following modes:
  * Light
  * Auto
  * Dark
* Our default theme will provide functionality for all three modes
  * There will be no separate *light* theme to choose
* Any other theme also should be able provide functionality for all three modes

# How this can be achieved

## Our default theme uses [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) in places where it would use LESS variables

This will look like this: `background-color: var(--body-bg-color, @body-bg-color);`

`--body-bg-color` is the custom CSS property. `@body-bg-color` is the LESS variable that's been used previously and is now defined as default in case the custom property isn't defined.

Since we still have to provide support for IE11, we have to provide a fallback for that as well:

```css
background-color: @body-bg-color;
background-color: var(--body-bg-color, @body-bg-color);
```

This however should be abstracted away using a LESS mixin:

```less

.var(@property, @variable) {
  @{property}: @variable;
  @{property}: var(--@variable, @variable);
}

.var(background-color, body-bg-color); // These might need be strings
```

For this to work, we have to migrate our LESS parser to https://github.com/wikimedia/less.php as https://github.com/leafo/lessphp is not maintained anymore and [doesn't provide a required feature for this](https://github.com/leafo/lessphp/issues/585). We have to make sure [it works with PHP 5.6](https://github.com/wikimedia/less.php/issues/50#issuecomment-747593459) though.

### Compatibility

Whether it's the IE11 fallback that also applies to all other [browsers not supporting CSS custom properties](https://caniuse.com/css-variables) or the use of each LESS variable as default property value, we should not produce any breaking changes to themes with this. (At least not more than what we'll introduce with #4337)

### Caching

This doesn't require a change in the caching implementation. It's native CSS/LESS after all.

## Themes not only need to use the `prefers-color-scheme` media query, but also a custom *extension* so to say

We have to provide the possibility to switch off any automatic adjustments. That's what the two modes `Light` and `Dark` are meant for. Choosing either will switch off the automatism and only enable `Light` or `Dark`.

This can only be achieved by extending the media query with additional conditions. Since it's not possible to add generic comparisons to media queries, we have to utilize a condition that is still generic enough:

```less
@media (min-height: @prefer-light-color-scheme), (prefers-color-scheme: light) and (min-height: @enable-color-preference) {
  :root {
    --body-bg-color: white;
  }
}

@media (min-height: @prefer-dark-color-scheme), (prefers-color-scheme: dark) and (min-height: @enable-color-preference) {
  :root {
    --body-bg-color: black;
  }
}
```

The three LESS variables should be interpreted as booleans, where `min-height: 0px == true` and `min-height: 999999px == false`.

`@prefer-light-color-scheme` will be set to `0px` if the user chooses `Light` as mode, `999999px` otherwise. The same applies to `@prefer-dark-color-scheme` but for the `Dark` mode. `@enable-color-preference` will be only be set to `0px` if `Auto` is chosen as mode, `999999px` otherwise.

### Compatibility

Existing themes are not subject to this in any way. New themes not using the extension, will always be in mode `Auto`. Themes not using the media query, have no mode at all.

:question: We may show the user a warning in the preferences if a theme is chosen that doesn't apply the extension.
:exclamation: No warning though for no media query. But we may hide/disable the option for the mode.

The default theme must not apply any mode if a theme is active that has none.
We should ensure it's not possible for themes or modules to override any of the three LESS variables.

### Caching

The three LESS variables `@enable-color-preference`, `@prefer-light-color-scheme` and `@prefer-dark-color-scheme` need to be put in a separate file for each combination. Only one of them is served based on the user's preference though.

**Mode `Light`**

```less
@enable-color-preference: 999999px;
@prefer-light-color-scheme: 0px;
@prefer-dark-color-scheme: 999999px;
```

**Mode `Auto`**

```less
@enable-color-preference: 0px;
@prefer-light-color-scheme: 999999px;
@prefer-dark-color-scheme: 999999px;
```

**Mode `Dark`**

```less
@enable-color-preference: 999999px;
@prefer-light-color-scheme: 999999px;
@prefer-dark-color-scheme: 0px;
```

**No Mode**

```less
@enable-color-preference: 999999px;
@prefer-light-color-scheme: 999999px;
@prefer-dark-color-scheme: 999999px;
```

# What our `default` and `light` theme does

The `default` theme isn't a real theme and is `Dark` by default. There's no need to change this in #4337. But wherever a LESS variable is used for a color, we have to replace that with a CSS custom property.

The `light` theme however needs to be integrated into the `default` theme with the appropriate media query. This is required because if there ever is another theme providing a `Dark` and `Light` mode, it should be able to rely on the fact that the `default` theme applies its own *light* rules as well. Themes were handled **in addition** to our default theme since ever, this must not change.

sukhwinder33445:
resolves #4386


library/Icinga/Web/StyleSheet.php
```diff
@@ -236,4 +250,32 @@ public function __toString()
...
     }
+
+    /**
+     * Get the path to the current LESS theme file
+     *
+     * @param $theme
+     *
+     * @return  string|null Return null if self::DEFAULT_THEME is set as theme, path otherwise
+     */
+    public static function getThemeFile($theme)
```
nilmerg, line 261: You should also use this where you've got it from.

library/Icinga/Web/StyleSheet.php
```diff
@@ -149,6 +149,20 @@ protected function collect()
...
+
+            if (($theme === null || $theme === self::DEFAULT_THEME) && empty($mode)) {
+                $mode = 'dark';
+            } elseif (empty($mode)) {
+                $mode = 'none';
+            }
+
+            $this->lessCompiler->addLessFile($this->pubPath . '/css/modes/'. $mode .'.less');
+        } else {
+            $this->lessCompiler->addLessFile($this->pubPath . '/css/modes/none.less');
```
nilmerg, line 164: We should introduce a constant like `self::DEFAULT_THEME` analogous to the the mode. So please add `self::DEFAULT_MODE` and use it.

Also, don't apply the mode `none` for anonymous users, apply the default mode.

library/Icinga/Web/StyleSheet.php
```diff
@@ -149,6 +149,20 @@ protected function collect()
...
             }
         }
+
+        if ($user = Auth::getInstance()->getUser()) {
+            $mode = $user->getPreferences()->getValue('icingaweb', 'mode');
+
+            if (($theme === null || $theme === self::DEFAULT_THEME) && empty($mode)) {
+                $mode = 'dark';
+            } elseif (empty($mode)) {
+                $mode = 'none';
```
nilmerg, line 159: Remember, you can define a default value for `getValue` above.

library/Icinga/Web/StyleSheet.php
```diff
@@ -149,6 +149,20 @@ protected function collect()
                 $this->lessCompiler->setTheme($this->pubPath . '/css/themes/' . $theme . '.less');
             }
         }
+
+        if ($user = Auth::getInstance()->getUser()) {
+            $mode = $user->getPreferences()->getValue('icingaweb', 'mode');
+
+            if (($theme === null || $theme === self::DEFAULT_THEME) && empty($mode)) {
```
nilmerg, line 156: Why are you checking the theme here?

application/forms/PreferenceForm.php
```diff
@@ -202,6 +203,25 @@ public function createElements(array $formData)
...
+        ) {
+            $this->addElement(
+                'radio',
+                'mode',
+                array(
+                    'label'         => $this->translate('Mode'),
+                    'multiOptions'  => [
+                        'dark'  => 'Dark',
+                        'light' => 'Light',
+                        'auto'  => 'Auto'
```
nilmerg, line 218: Please mark the labels for translation.

application/forms/PreferenceForm.php
```diff
@@ -202,6 +203,25 @@ public function createElements(array $formData)
...
         }
 
+        if (! isset($formData['theme'])
+            || $formData['theme'] === StyleSheet::DEFAULT_THEME
+            || strpos(file_get_contents(StyleSheet::getThemeFile($formData['theme'])),'prefers-color-scheme') !== false
+        ) {
+            $this->addElement(
+                'radio',
+                'mode',
+                array(
```
nilmerg, line 213: Please don't use the legacy array syntax :)

application/forms/PreferenceForm.php
```diff
@@ -202,6 +203,25 @@ public function createElements(array $formData)
             }
         }
 
+        if (! isset($formData['theme'])
+            || $formData['theme'] === StyleSheet::DEFAULT_THEME
+            || strpos(file_get_contents(StyleSheet::getThemeFile($formData['theme'])),'prefers-color-scheme') !== false
```
nilmerg, line 208: Only checking for `prefers-color-scheme` isn't enough. You'd need to check also for the extensions using `min-height`.

See, a theme not using the `min-height` extensions will always be in mode `auto`. If the user chooses `dark` or `light` then, it has no effect. I'd rather make only the `auto` option selectable in this case.

nilmerg (changes requested):

library/Icinga/Web/StyleSheet.php
sukhwinder33445: because when you log in again, the default theme and dark mode are automatically selected. But `$mode = $user->getPreferences()->getValue('icingaweb', 'mode', 'none');` always returns `‘none'`. I suspect this is because dark mode is shown selected but has not yet been saved in form. Is there another way to do this?

nilmerg: I don't understand this. It is of course required to save before logging out and back in. If you get `none` using the call above, it is either configured this way or saving the form didn't persist the chosen mode.

application/forms/PreferenceForm.php
```diff
@@ -202,6 +203,33 @@ public function createElements(array $formData)
             }
         }
 
+        if (! isset($formData['theme'])
+            || $formData['theme'] === StyleSheet::DEFAULT_THEME
+            || strpos(file_get_contents(StyleSheet::getThemeFile($formData['theme'])), 'prefers-color-scheme') !== false
+            || strpos(file_get_contents(StyleSheet::getThemeFile($formData['theme'])), 'min-height') === false
```
nilmerg, line 209: You can't just search for `min-height`, this is a standard CSS property, that can occur anywhere in the file.

application/forms/PreferenceForm.php
```diff
@@ -202,6 +203,33 @@ public function createElements(array $formData)
...
+        ) {
+            $modes = [
+                'dark'  => $this->translate('Dark'),
+                'light' => $this->translate('Light'),
+                'auto'  => $this->translate('Auto')
+            ];
+            if (strpos(file_get_contents(StyleSheet::getThemeFile($formData['theme'])), 'min-height') === false) {
```
nilmerg, line 216: You should check this only once.

nilmerg: I also would like to see the other options disabled, not invisible. This should work with this element option `'disable' => ['dark', 'light']`.

library/Icinga/Web/StyleSheet.php
```diff
@@ -137,17 +144,20 @@ protected function collect()
...
-                    $this->lessCompiler->setTheme($module->getCssDir() . '/themes/' . $theme . '.less');
-                }
-            } else {
-                $this->lessCompiler->setTheme($this->pubPath . '/css/themes/' . $theme . '.less');
+        if ($themePath = self::getThemeFile($theme)) {
+                $this->lessCompiler->setTheme($themePath);
+        }
+
+        if ($user = Auth::getInstance()->getUser()) {
+            $mode = $user->getPreferences()->getValue('icingaweb', 'mode', 'none');
```
nilmerg, line 152: Can't you just use the default mode as default value here? `none` shouldn't be possible to configure anyway.

library/Icinga/Web/StyleSheet.php
```diff
@@ -137,17 +144,20 @@ protected function collect()
...
+
+            if (($theme === null || $theme === self::DEFAULT_THEME) && $mode === 'none') {
+                $mode = self::DEFAULT_MODE;
             }
+
+            $this->lessCompiler->addLessFile($this->pubPath . '/css/modes/'. $mode . '.less');
```
nilmerg, line 158: This will never serve the file for mode `none`. You need to check the active theme here as well, in order to decide whether to respect the user's preference or apply `none` as mode if the chosen theme has no support for it.

nilmerg (changes requested):
Please also check the status checks.

library/Icinga/Web/StyleSheet.php
```diff
@@ -137,17 +144,20 @@ protected function collect()
...
-            } else {
-                $this->lessCompiler->setTheme($this->pubPath . '/css/themes/' . $theme . '.less');
+        if ($themePath = self::getThemeFile($theme)) {
+                $this->lessCompiler->setTheme($themePath);
+        }
+
+        if ($user = Auth::getInstance()->getUser()) {
+            $mode = $user->getPreferences()->getValue('icingaweb', 'mode', self::DEFAULT_MODE);
+
+            if ($mode === null) {
```
nilmerg, line 154: I think I now understand why you defined `none` as default above previously. Because currently `none` still isn't applied, as `null` is only returned if you don't define a default value above.

But I'm not that fond of using just the fact that nothing is configured as reason for the `none` mode anyway. Configuration may not only be written by the `PreferenceForm`, but also by other tools. These tools probably won't verify that a theme supports modes at all. Also, it's possible to define a global default theme (application-wide) so then it's not under control of the user and the theme may also one that doesn't support any modes.

You really need to perform the same check as in the `PreferenceForm` (the regex) for the theme that is being applied above.

application/forms/PreferenceForm.php
```diff
@@ -202,6 +203,35 @@ public function createElements(array $formData)
...
 
+        $regEx = '/@media\s*\(\s*min-height\s*:\s*@prefer-(light|dark)-color-scheme\s*\)\s*,\s*\(\s*prefers-color-scheme\s*:\s*(light|dark)\s*\)\s*and\s*\(\s*min-height\s*:\s*@enable-color-preference\s*\)/';
+        $regEx2 = '/\(\s*min-height\s*:\s*@enable-color-preference\s*\)/';
+        $enableAuto = [];
+        if (isset($formData['theme']) && $formData['theme'] !== StyleSheet::DEFAULT_THEME) {
+            $file = file_get_contents(StyleSheet::getThemeFile($formData['theme']));
+            $hasMode = (bool) preg_match_all($regEx, $file, $matches);
+            $onlyAuto = $hasMode ? null : preg_match($regEx2, $file);
+            $enableAuto = ($hasMode || $onlyAuto) ? ['auto'] : [];
+        }
```
nilmerg, line 214: Uh oh, I don't think it needs to be that complicated. :smile: 

You only need to check for the light mode query: `@media\s*\(\s*min-height\s*:\s*@prefer-light-color-scheme\s*\)\s*,\s*\(\s*prefers-color-scheme\s*:\s*light\s*\)\s*and\s*\(\s*min-height\s*:\s*@enable-color-preference\s*\)`

If it matches, this means that all three modes are available. If it doesn't match, check for any mode query: `@media.*prefers-color-scheme`

If this matches, only the `Auto` mode is available.

application/forms/PreferenceForm.php
```diff
@@ -202,6 +203,35 @@ public function createElements(array $formData)
...
+
+        $modes = [
+            'dark'      => $this->translate('Dark'),
+            'light'     => $this->translate('Light'),
+            'auto'      => $this->translate('Auto')
+        ];
+
+        $disabled = isset($matches) ? array_diff(array_keys($modes), $matches[1], $enableAuto) : [];
```
nilmerg, line 222: I guess `$onlyAuto ? ['dark', 'light'] : []` suffices here :wink: 

You could also do that below in the option array.

application/forms/PreferenceForm.php
nilmerg: Oh and please make the regex patterns constants in class `StyleSheet`.

nilmerg (changes requested):
Besides my comments please adjust the following:

* Use `theme_mode` as option name and `Theme Mode` as label
* Don't use `Auto` as label, but `System`
* Completely remove the legacy parser and put the new one it its place
* Set the mode on the `LessCompiler` (Just like it's being done with `setTheme` for the theme) and render the file *after* the theme
  * We have to make sure to prevent the mode vars from being overidden

public/css/icinga/layout.less
```diff
@@ -82,6 +82,7 @@
 
 #layout {
   background-color: @body-bg-color;
+  background-color: var(--body-bg-color, @body-bg-color);
```
nilmerg, line 85: Please remove that. It's been for tests only.

nilmerg (changes requested):
Please also remove the `Modes` theme. That's been for testing only as well.

library/Icinga/Web/LessCompiler.php
```diff
@@ -118,6 +125,10 @@ public function getLessFiles()
     {
         $lessFiles = $this->lessFiles;
 
+        if ($this->themeMode !== null) {
+            $lessFiles[] = $this->themeMode;
+        }
+
```
nilmerg, line 131: Please put that at the very end so that it's applied **after** the theme. This will prevent that themes can override the mode's LESS variables.

nilmerg (changes requested):

nilmerg (commented):
Nice! Thank you!

It's now @flourish86's turn to adjust the input styling.

flourish86:
> Nice! Thank you!
> 
> It's now @flourish86's turn to adjust the input styling.

✅ with c2e77a1

public/css/icinga/forms.less
```diff
@@ -517,3 +517,26 @@ form.icinga-form .form-info {
 .flyover-content .control-label-group {
   text-align: left;
 }
+
+.control-group > label {
+  text-align: center;
+}
```
nilmerg, line 523: This is

a) too broad. All form styles are wrapped in either `.icinga-form` (layout) or `.icinga-controls` (style).
b) There are other cases where a label is directly in a `.control-group` and this doesn't make any sense

Remove this and wrap the theme mode's labels with `span`s and style those instead.

public/css/icinga/forms.less
```diff
@@ -517,3 +517,26 @@ form.icinga-form .form-info {
...
+
+.theme-mode-input {
+  display: none;
+  text-align: center;
```
nilmerg, line 527: Doesn't make any sense, the input is hidden, remove this.

application/forms/PreferenceForm.php
```diff
@@ -202,6 +209,46 @@ public function createElements(array $formData)
...
+                        ) . $this->translate('Light'),
+                    'system'    => new HtmlElement(
+                        'img', [
+                            'src' => $this->getView()->href('img/theme-mode-thumbnail-system.svg')
+                            ]
+                        ) . $this->translate('System')
+                ],
+                'value'         => isset($value) ? $value : '',
+                'disable'       => isset($disabled) ? $disabled : [],
+                'escape' => false
```
nilmerg, line 248: Woah, check your alignment. Also, use `HtmlElement::create()` instead of `new HtmlElement()`, the latter changed recently and it won't work anymore how you're using it here.

nilmerg (changes requested):
I like the preview pictures!!!1 Though, please have a look at my comments..

Also, why are the pictures SVG with embedded binary stuff? (png I suppose) Please make them either non-SVG's or real SVG's.

nilmerg:
There's also no visible change between enabled/disabled. Just select e.g. the high contrast theme and you can't interact with the controls, but that's not visible in any way.

public/css/icinga/forms.less
flourish86: ✅

application/forms/PreferenceForm.php
flourish86: ✅

flourish86:
> There's also no visible change between enabled/disabled. Just select e.g. the high contrast theme and you can't interact with the controls, but that's not visible in any way.

Good point. Fixed ✅

public/css/icinga/forms.less
flourish86: ✅

nilmerg (approved):

nilmerg (approved):

PR #4409, opened 2021-06-30
Url: Implement hasParams() method

Yonas-net:


nilmerg:
Is it too complicated to call `$url->getParams()->isEmpty()` ? :smile:

Yonas-net:
It isn't, but it sucks when sometimes you have to do a line break because of it. You can close it if you don't want to.

nilmerg:
Line breaks are not a valid reason to introduce redundancy. :wink:

Issue #4421, opened 2021-07-02
Checkboxes and permission names overlap in Safari


Yonas-net:
## Describe the bug
When the permission has a long name, the Cehckboxes overlaps it, instead of making a line break as in Chrome. See screenshot.

<img width="850" alt="Bildschirmfoto 2021-07-02 um 17 36 54" src="https://user-images.githubusercontent.com/57616252/124298243-55546300-db5c-11eb-8dec-df4de0957a57.png">

PR #4425, opened 2021-07-05
CSS: Make labels wrap in Safari

flourish86:
refs #4421

### DIFF
--- a/public/css/icinga/widgets.less
+++ b/public/css/icinga/widgets.less
@@ -231,6 +231,10 @@ form.role-form {
     font-style: normal;
   }
 
+  .control-label > * {
+    display: inline-block;
+  }
+
   h3 {
     > :first-child {
       display: inline-block;


public/css/icinga/widgets.less
```diff
@@ -224,6 +224,8 @@ table.multiselect tr[href] td {
 form.role-form {
   &.icinga-form .control-label-group {
     width: 20em;
+    // make labels wrap in Safari
+    line-break: anywhere;
```
nilmerg, line 228: I don't like the use of `line-break` here. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/line-break) this is a locale specific property which just happens to work in our case. What I specifically don't like is the statement `"disregarding any prohibition against line breaks"` which several parts of the permission name intentionally have and hence why this issue in Safari popped up to begin with.

I can't find a case where a `white-space: nowrap` isn't respected though, not with Chromium or Firefox. So this may only be a case of mis-interepretation or a different browser behavior despite what MDN says.

Is there a chance that another property has the same effect? Is MDN indeed wrong and the spec/browser behavior is entirely different? (Source pls)

public/css/icinga/widgets.less
```diff
@@ -231,6 +231,10 @@ form.role-form {
     font-style: normal;
   }
 
+  .control-label > * {
+    display: inline-block;
```
nilmerg, line 235: Okay. That results in Safari breaking on the zero-width-space? But it's the `em`s which get this now, they don't contain the zero-width-spaces at all...

Please explain this. :sweat_smile:

nilmerg (approved):

PR #4429, opened 2021-07-07
Dashboard: Add ability to use IPL Icons

flourish86:
refs Icinga/icingadb-web/#59

### DIFF
--- a/application/views/scripts/navigation/dashboard.phtml
+++ b/application/views/scripts/navigation/dashboard.phtml
@@ -1,3 +1,8 @@
+<?php
+
+use ipl\Web\Widget\Icon;
+
+?>
 <div class="controls">
     <?= $tabs ?>
 </div>
@@ -5,7 +10,13 @@
     <?php foreach ($navigation as $item): /** @var \Icinga\Web\Navigation\NavigationItem $item */?>
     <a class="dashboard-link" href="<?= $this->url($item->getUrl(), $item->getUrlParameters()) ?>"<?= $this->propertiesToString($item->getAttributes()) ?>>
         <div class="link-icon">
-            <?= $this->icon($item->getIcon() ?: 'forward', null, array('aria-hidden' => true)) ?>
+        <?php
+            if (substr($item->getUrl()->getPath(), 0, 9) === 'icingadb/') {
+                echo new Icon($item->getIcon(), [ 'aria-hidden' => 1]);
+            } else {
+                echo $this->icon($item->getIcon() ?: 'forward', null, array('aria-hidden' => true));
+            }
+        ?>
         </div>
         <div class="link-meta">
             <div class="link-label"><?= $this->escape($item->getLabel()) ?></div>
--- a/public/css/icinga/widgets.less
+++ b/public/css/icinga/widgets.less
@@ -125,10 +125,16 @@
   display: table-cell;
   padding-right: .5em;
   vertical-align: middle;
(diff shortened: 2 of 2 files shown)


flourish86:
With this you're able to use 'ipl\Web\Widget\Icon' in configuration files like this:

```
    $section->add(N_('Tactical Overview'), [
        'url' => 'icingadb/tactical',
        'priority' => 110,
        'icon'     => new Icon('chart-pie')
    ]);
```

nilmerg:
To me this is not only a thing for the navigation dashboard, but for the entire menu. Top level menu entries also have icons. Overall, this is also an entire Web 2 thing, because the ipl (where the new Icons come from) is now a core dependency. So we should use and support the new icons anyway, everywhere. But that's out of scope here and should be part of the theme sprint.

We shouldn't attempt to support something here what we wouldn't in the future. You proposal includes such changes. I'd want to support the new icons only for Icinga DB Web here, no-one else should be able to use them until v2.10. Then we can still decide how to properly implement the support.

I'll provide suggestions for that in a separate review now.

application/views/scripts/navigation/dashboard.phtml
```diff
@@ -1,11 +1,22 @@
...
 <div class="controls">
     <?= $tabs ?>
 </div>
 <div class="content">
     <?php foreach ($navigation as $item): /** @var \Icinga\Web\Navigation\NavigationItem $item */?>
     <a class="dashboard-link" href="<?= $this->url($item->getUrl(), $item->getUrlParameters()) ?>"<?= $this->propertiesToString($item->getAttributes()) ?>>
         <div class="link-icon">
-            <?= $this->icon($item->getIcon() ?: 'forward', null, array('aria-hidden' => true)) ?>
+        <?php
+            if (gettype($item->getIcon()) == 'object' && get_class($item->getIcon()) == 'ipl\Web\Widget\Icon') {
```
nilmerg, line 14: Please check the url to decide whether to use the old or new icons.

`if (substr($item->getUrl(), 0, 9) === 'icingadb/') {`

This way only Icinga DB Web should able to use them.

By the way, `$item->getIcon() instanceof \ipl\Web\Widget\Icon` would have been the correct check and loose comparisons (`==`) are discouraged most of the time, use strict comparsions (`===`) instead :wink:

application/views/scripts/navigation/dashboard.phtml
```diff
@@ -1,11 +1,22 @@
...
+                echo (new HtmlString($item->getIcon()));
```
nilmerg, line 15: You shouldn't need `HtmlString` here. Also expect the icon to be still a string, so it should be `new Icon($item->getIcon())` instead.

And please add the `aria-hidden` attribute just like it's done for the old icons.

Why a string? `configuration.php` files should be as lightweight as possible. Menu entries are such a common thing, it shouldn't be required to create instances of `Icon` explicitly.

nilmerg (changes requested):

application/views/scripts/navigation/dashboard.phtml
flourish86: ✅

application/views/scripts/navigation/dashboard.phtml
flourish86: ✅

nilmerg (approved):

PR #4432, opened 2021-07-08
Revert the monitoring preview picture to light theme

theFeu:
I don't know whether you might want to switch it to another screenshot, but I just reverted the change here :)


nilmerg:
I'm not sure if Github is the primary location where it's required to reflect exactly the official style. Isn't that icinga.com? :wink: 

The master still has the dark theme. It's part of the next major and... again, this single picture isn't even enough to show how Web 2 looks and what features are provided. So it's hardly important.

I did think about the same, reverting it, but dismissed it due to the above. If at all I'd merge this into the branch that gets the 2.9 tag, so if anyone looks at said tag or branch gets the correct preview picture. :man_shrugging:

theFeu:
Ah yes, fair enough :)
I was of the opinion that the readme would also be mirrored to the docs, but after checking it now, that's not the case.

Then this can be closed, sorry!

PR #4472, opened 2021-07-21
Rememberme compatibility with php version 5.6+

sukhwinder33445:


library/Icinga/Crypt/AesCrypt.php
```diff
@@ -217,4 +232,35 @@ public function encryptToBase64($data)
...
+        $encrypt = openssl_encrypt($data, $this->method, $this->getkey(), 0, $this->getIV());
+
+        if (is_bool($encrypt) && $encrypt === false) {
+            throw new RuntimeException('Encryption failed');
+        }
+
+        $hmac = hash_hmac('sha256', $encrypt, $this->getkey(), $as_binary=true);
+
+        return base64_encode($this->getIV() . $hmac . $encrypt);
+    }
```
nilmerg, line 265: * Please don't define variables in calls (e.g. `$sha2len` and `$as_binary`, the latter even doesn't use camel case notation)
* You can drop the `is_bool` check when using the strict comparison operator for the variable in question. Strict comparison also checks the type.
* Why is there `$iv` and `$this->getIV()`? Are both the same? Can't `$data` have only the `$hmac` if the IV is already known otherwise?
* Please remove the `get` prefix from both methods, they're no getters

nilmerg (changes requested):

nilmerg (approved):

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -55,12 +55,16 @@ class AesCrypt
 
     const GCM_SUPPORT_VERSION = '7.1';
 
-    public function __construct($random_bytes_len = 128)
+    public function __construct($forceMethod = null, $random_bytes_len = 128)
```
nilmerg, line 58: I'd like to see a setter here instead.

nilmerg (changes requested):

library/Icinga/Crypt/AesCrypt.php
sukhwinder33445: I wanted to do it with setter too, but the problem is, iv and key depend on the method. Or is there a way to call setter before the constructor?

nilmerg: Then, instead of throwing a RuntimeError in case of a missing iv and key, generate them on demand in their getter?

sukhwinder33445: Sorry, only the iv depends on the method.

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -51,15 +51,31 @@ class AesCrypt
...
+
         $this->key = random_bytes($random_bytes_len);
     }
 
+    /**
+     * Force set the method
+     *
+     * @return $this
+     */
+    public function forceSetMethod($method)
```
nilmerg, line 72: Just `setMethod` please.

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -110,7 +126,9 @@ public function setIV($iv)
     public function getIV()
     {
         if (empty($this->iv)) {
-            throw new RuntimeException('No iv set');
+            $len = openssl_cipher_iv_length($this->method);
+
+            return $this->iv = random_bytes($len);
```
nilmerg, line 131: Why do you return here? The method does so anyway in the next line.

nilmerg (changes requested):

nilmerg (approved):

PR #4496, opened 2021-07-28
Feature/redesign about view

### LINKED ISSUE #4328: Redesign System/About View
It's ugly and the inner container scrolls, when exceeding viewport. So we should come up with a visual redesign.

Screen Shot 2021-03-01 at 15 23 49

flourish86:
Resolves #4328

### DIFF
--- a/application/views/scripts/about/index.phtml
+++ b/application/views/scripts/about/index.phtml
@@ -1,25 +1,140 @@
+<?php
+
+use ipl\Html\HtmlElement;
+use ipl\Web\Widget\Icon;
+
+?>
 <div class="controls">
     <?= $tabs ?>
 </div>
-<div id="about" class="content content-centered">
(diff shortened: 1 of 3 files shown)


nilmerg:
Please reference the issue with `resolves`

flourish86:
Resolves #4496

nilmerg (approved):

flourish86:
Thanks for taking over.

It's good how the width grows, with more content in the cells.

Can we make the "configure" buttons in `@icinga-blue`? See #4683

I also noticed another thing: If there's really long cell content the content width exceeds the viewport's width. Maybe this is not a problem in most cases, but it could be glitchy on narrower screen sizes.

Screen Shot 2022-02-28 at 09 32 15

nilmerg:
> I also noticed another thing

The commit id is of fixed length, that won't happen.

flourish86:
> > I also noticed another thing
> 
> The commit id is of fixed length, that won't happen.

Ok, seems to work in all screen widths, then.

PR #4500, opened 2021-08-03
Utilize multiple encryption ciphers for remember me

### LINKED ISSUE #4493: Stay Logged In - Unknown cipher algorithm
Hello, 

when i open the icingaweb2 in chrome i get the error "open_ssl_decrypt(): Unknown ciper algoritm"
If I do this with Edge or Firefox I don't have this problem
I clear also the cache of chrome ! 

Firefox 87
Edge Version 92.0.902.55 
Google Chrome : Version 92.0.4515.107 (Official Build) (64-bit)

In the /var/log/message I got the error :
Jul 28 13:25:56 xxxx icingaweb2[10347]: ErrorException in /usr/share/php/Icinga/Crypt/AesCrypt.php:186 with message: openssl_decrypt(): Unknown cipher algorithm   

#0 [internal function]: Icinga\Application\ApplicationBootstrap->Icinga\Application\{closure}(Integer, String, String, Integer, Array)   
#1 /usr/share/php/Icinga/Crypt/AesCrypt.php(186): openssl_decrypt(String, String, String, Integer, String, String)   
#2 /usr/share/php/Icinga/Crypt/AesCrypt.php(206): Icinga\Crypt\AesCrypt->decrypt(String)   
#3 /usr/share/php/Icinga/Web/RememberMe.php(211): Icinga\Crypt\AesCrypt->decryptFromBase64(String)   
#4 /usr/share/icingaweb2/application/controllers/AuthenticationController.php(52): Icinga\Web\RememberMe->authenticate()   
#5 /usr/share/icingaweb2/library/vendor/Zend/Controller/Action.php(507): Icinga\Controllers\AuthenticationController->loginAction()   
#6 /usr/share/php/Icinga/Web/Controller/Dispatcher.php(76): Zend_Controller_Action->dispatch(String)   
#7 /usr/share/icingaweb2/library/vendor/Zend/Controller/Front.php(937): Icinga\Web\Controller\Dispatcher->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))   
#8 /usr/share/php/Icinga/Application/Web.php(304): Zend_Controller_Front->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))   
#9 /usr/share/php/Icinga/Application/webrouter.php(107): Icinga\Application\Web->dispatch()   
#10 /usr/share/icingaweb2/public/index.php(4): require_once(String)   
#11 {main}

See here extra info : 
# icinga2 --version

icinga2 - The Icinga 2 network monitoring daemon (version: 2.12.5-1)
Copyright (c) 2012-2021 Icinga GmbH (https://icinga.com/)
License GPLv2+: GNU GPL version 2 or later <http://gnu.org/licenses/gpl2.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

System information:
  Platform: Red Hat Enterprise Linux Server
  Platform version: 7.9 (Maipo)
  Kernel: Linux
  Kernel version: 3.10.0-1160.31.1.el7.x86_64
  Architecture: x86_64

Build information:
  Compiler: GNU 4.8.5
  Build host: runner-hh8q3bz2-project-505-concurrent-0
  OpenSSL version: OpenSSL 1.0.2k-fips  26 Jan 2017

Application information:

General paths:
  Config directory: /etc/icinga2
  Data directory: /var/lib/icinga2
  Log directory: /var/log/icinga2
  Cache directory: /var/cache/icinga2
  Spool directory: /var/spool/icinga2
  Run directory: /run/icinga2

Old paths (deprecated):
  Installation root: /usr
  Sysconf directory: /etc
  Run directory (base): /run
  Local state directory: /var

Internal paths:
  Package data directory: /usr/share/icinga2
  State path: /var/lib/icinga2/icinga2.state
  Modified attributes path: /var/lib/icinga2/modified-attributes.conf
  Objects path: /var/cache/icinga2/icinga2.debug
  Vars path: /var/cache/icinga2/icinga2.vars
  PID path: /run/icinga2/icinga2.pid
  
  
  Is this a bug or is there something i'm overlooking ?
  
  regards,
  Geert

### LINKED ISSUE #4497: openssl_cipher_iv_length(): Unknown cipher algorithm
## Describe the bug
Using the RememberMe function we get this error
immagine

We have identified that come from this file /usr/share/php/Icinga/Crypt/AesCrypt.php where the upper case of the cipher cause this function to fail

`$len = openssl_cipher_iv_length($this->method);`

to solve the error, in the constructor we have changed the method to lowercase

NOT WORKING

```php
    /** @var string The cipher method */
    private $method = 'AES-128-GCM';

    const GCM_SUPPORT_VERSION = '7.1';

    public function __construct($random_bytes_len = 128)
    {
        if (version_compare(PHP_VERSION, self::GCM_SUPPORT_VERSION, '<')) {
            $this->method = 'AES-128-CBC';
        }

        $this->key = random_bytes($random_bytes_len);
    }
```
WORKING
```php
    /** @var string The cipher method */
    private $method = 'aes-128-gcm';

    const GCM_SUPPORT_VERSION = '7.1';

    public function __construct($random_bytes_len = 128)
    {
        if (version_compare(PHP_VERSION, self::GCM_SUPPORT_VERSION, '<')) {
            $this->method = 'AES-128-CBC';
        }

        $this->key = random_bytes($random_bytes_len);
    }
```
## Your Environment

* Icinga Web 2 version and modules (System - About): 2.9.2
* Web browser used: Chrome
* Icinga 2 version used (`icinga2 --version`): 2.12.5
* PHP version used (`php --version`): 7.3.20 and 7.1
* Server operating system and version: Red Hat Enterprise Linux Server release 7.8 (Maipo)

sukhwinder33445:
fix #4493

### DIFF
--- a/library/Icinga/Crypt/AesCrypt.php
+++ b/library/Icinga/Crypt/AesCrypt.php
@@ -14,7 +14,7 @@ use RuntimeException;
  * ```php
  *
  * // Encryption
- * $encryptedData = new AesCrypt()->encrypt($data); // Accepts a string
+ * $encryptedData = (new AesCrypt())->encrypt($data); // Accepts a string
  *
  *
  * // Encrypt and encode to Base64
@@ -23,7 +23,7 @@ use RuntimeException;
  *
  * // Decryption
  * $aesCrypt = (new AesCrypt())
- *          ->setTag($tag)
+ *          ->setTag($tag)  // if exists
  *          ->setIV($iv)
  *          ->setKey($key);
  *
@@ -35,14 +35,24 @@ use RuntimeException;
  *          ->setIV($iv)
  *          ->setKey($key);
  *
- * $decryptedData = $aesCrypt->->decryptFromBase64($data);
+ * $decryptedData = $aesCrypt->decryptFromBase64($data);
  * ```
  *
  */
 class AesCrypt
 {
+    /** @var array The list of cipher methods */
+    const METHODS = [
+        'aes-256-gcm',
+        'aes-256-cbc',
+        'aes-256-ctr'
+    ];
+
     /** @var string The encryption key */
     private $key;
+
+    /** @var int The length of the key */
+    private $keyLength;
 
     /** @var string The initialization vector which is not NULL */
     private $iv;
@@ -51,21 +61,15 @@ class AesCrypt
     private $tag;
 
     /** @var string The cipher method */
-    private $method = 'AES-128-GCM';
+    private $method;
 
-    const GCM_SUPPORT_VERSION = '7.1';
-
-    public function __construct($random_bytes_len = 128)
+    public function __construct($keyLength = 128)
     {
-        if (version_compare(PHP_VERSION, self::GCM_SUPPORT_VERSION, '<')) {
-            $this->method = 'AES-128-CBC';
-        }
-
-        $this->key = random_bytes($random_bytes_len);
+        $this->keyLength = $keyLength;
     }
 
     /**
-     * Force set the method
+     * Set the method
      *
      * @return $this
      */
@@ -74,6 +78,45 @@ class AesCrypt
         $this->method = $method;
 
         return $this;
+    }
+
+    /**
+     * Get the method
+     *
+     * @return string
+     */
+    public function getMethod()
+    {
+        if ($this->method === null) {
+            $this->method = $this->getSupportedMethod();
+        }
+
+        return $this->method;
+    }
+
+    /**
+     * Get supported method
+     *
+     * @return string
+     *
+     * @throws RuntimeException If none of the methods listed in the METHODS array is available
+     */
+    protected function getSupportedMethod()
+    {
+        $availableMethods = openssl_get_cipher_methods();
+        $methods = self::METHODS;
+
+        if (! $this->isAuthenticatedEncryptionSupported()) {
+            unset($methods[0]);
+        }
+
+        foreach ($methods as $method) {
+            if (in_array($method, $availableMethods)) {
+                return $method;
+            }
+        }
+
+        throw new RuntimeException('No supported method found');
     }
 
     /**
@@ -93,12 +136,11 @@ class AesCrypt
      *
      * @return string
      *
-     * @throws RuntimeException If the key is not set
      */
     public function getKey()
     {
         if (empty($this->key)) {
-            throw new RuntimeException('No key set');
+            $this->key = random_bytes($this->keyLength);
         }
 
         return $this->key;
@@ -121,12 +163,11 @@ class AesCrypt
      *
      * @return string
      *
-     * @throws RuntimeException If the IV is not set
      */
     public function getIV()
     {
         if (empty($this->iv)) {
-            $len = openssl_cipher_iv_length($this->method);
+            $len = openssl_cipher_iv_length($this->getMethod());
             $this->iv = random_bytes($len);
         }
 
@@ -137,9 +178,20 @@ class AesCrypt
      * Set the Tag
      *
      * @return $this
+     *
+     * @throws RuntimeException If a tag is available but authenticated encryption (AE) is not supported.
+     *
+     * @throws UnexpectedValueException If tag length is less then 16
      */
     public function setTag($tag)
     {
+        if (! $this->isAuthenticatedEncryptionSupported()) {
+            throw new RuntimeException(sprintf(
+                "The given decryption method is not supported in php version '%s'",
+                PHP_VERSION
+            ));
+        }
+
         if (strlen($tag) !== 16) {
             throw new UnexpectedValueException(sprintf(
                 'expects tag length to be 16, got instead %s',
@@ -169,7 +221,7 @@ class AesCrypt
     }
 
     /**
-     * Decrypt the given data using the key, iv and tag
+     * Decrypt the given string
      *
      * @param string $data
      *
@@ -179,11 +231,11 @@ class AesCrypt
      */
     public function decrypt($data)
     {
-        if ($this->method === 'AES-128-CBC') {
-            return $this->decryptCBC($data);
+        if (! $this->isAuthenticatedEncryptionRequired()) {
+            return $this->nonAEDecrypt($data);
         }
 
-        $decrypt = openssl_decrypt($data, $this->method, $this->getKey(), 0, $this->getIV(), $this->getTag());
+        $decrypt = openssl_decrypt($data, $this->getMethod(), $this->getKey(), 0, $this->getIV(), $this->getTag());
 
         if ($decrypt === false) {
             throw new RuntimeException('Decryption failed');
@@ -193,11 +245,13 @@ class AesCrypt
     }
 
     /**
-     * Decode from Base64 and decrypt the given data using the key, iv and tag
+     * Decode from Base64 and decrypt the given string
      *
      * @param string $data
      *
-     * @return string decrypted data
+     * @return string The base64 decoded and decrypted string
+     *
+     * @deprecated Use decrypt() instead as it also returns a base64 decoded string
      *
      * @throws RuntimeException If decryption fails
      */
@@ -207,21 +261,21 @@ class AesCrypt
     }
 
     /**
-     * Encrypt the given data using the key, iv and tag
+     * Encrypt the given string
      *
      * @param string $data
      *
-     * @return string encrypted data
+     * @return string encrypted string
      *
      * @throws RuntimeException If decryption fails
      */
     public function encrypt($data)
     {
-        if ($this->method === 'AES-128-CBC') {
-            return $this->encryptCBC($data);
+        if (! $this->isAuthenticatedEncryptionRequired()) {
+            return $this->nonAEEncrypt($data);
         }
 
-        $encrypt = openssl_encrypt($data, $this->method, $this->getkey(), 0, $this->getIV(), $this->tag);
+        $encrypt = openssl_encrypt($data, $this->getMethod(), $this->getKey(), 0, $this->getIV(), $this->tag);
 
         if ($encrypt === false) {
             throw new RuntimeException('Encryption failed');
@@ -231,11 +285,13 @@ class AesCrypt
     }
 
     /**
-     * Encrypt the given string using the the key, iv, tag and encode to Base64
+     * Encrypt the given string and encode to Base64
      *
      * @param string $data
      *
-     * @return string encrypted and encoded to Base64 data
+     * @return string encrypted and base64 encoded string
+     *
+     * @deprecated Use encrypt() instead as it also returns a base64 encoded string
      *
      * @throws RuntimeException If encryption fails
      */
@@ -244,18 +300,23 @@ class AesCrypt
         return base64_encode($this->encrypt($data));
     }
 
-    private function decryptCBC($data)
+    /**
+     * Decrypt the given string with non Authenticated encryption (AE) cipher method
+     *
+     * @param string $data
+     *
+     * @return string decrypted string
+     *
+     * @throws RuntimeException If decryption fails
+     */
+    private function nonAEDecrypt($data)
     {
-        if (strlen($this->getIV()) !== 16) {
-            throw new RuntimeException('Decryption failed');
-        }
-
         $c = base64_decode($data);
         $hmac = substr($c, 0, 32);
         $data = substr($c, 32);
 
-        $decrypt = openssl_decrypt($data, $this->method, $this->getKey(), 0, $this->getIV());
-        $calcHmac = hash_hmac('sha256', $data, $this->getKey(), true);
+        $decrypt = openssl_decrypt($data, $this->getMethod(), $this->getKey(), 0, $this->getIV());
+        $calcHmac = hash_hmac('sha256', $this->getIV() . $data, $this->getKey(), true);
 
         if ($decrypt === false || ! hash_equals($hmac, $calcHmac)) {
             throw new RuntimeException('Decryption failed');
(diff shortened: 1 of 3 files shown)


library/Icinga/Crypt/AesCrypt.php
```diff
@@ -50,22 +60,27 @@ class AesCrypt
...
-            $this->method = 'AES-128-CBC';
+        foreach (self::METHODS as $key => $method) {
+            if ($key === 0 && version_compare(PHP_VERSION, self::GCM_SUPPORT_VERSION, '<')) {
+                continue;
+            }
+            if (in_array($method, openssl_get_cipher_methods())) {
+                $this->method = $method;
+                break;
+            }
         }
```
nilmerg, line 77: Please put that into the getter, and only detect it in case there's no method set yet.

Also:
* Drop the `$key === 0` and php version check, that's redundant
* Cache the result of `openssl_get_cipher_methods()` in a variable

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -50,22 +60,27 @@ class AesCrypt
...
 
-        $this->key = random_bytes($random_bytes_len);
+        $this->key = random_bytes(128);
```
nilmerg, line 79: Why is this constant now? I see this change problematic, since it's been previously a parameter that may have already been used. (Backwards incompatible change)

Is this really necessary?

library/Icinga/Web/RememberMe.php
```diff
@@ -88,27 +93,25 @@ public static function fromCookie()
...
-                    ->setMethod('AES-128-CBC')
-                    ->setKey(hex2bin($rs->passphrase))
-                    ->setIV($iv);
-
-                $data[0] = $tag; // encryptedPass
-            } else {
-                $rememberMe->aesCrypt->setTag(base64_decode($tag));
+        if (! empty($data)) {
+            // php downgrade
+            if (version_compare(PHP_VERSION, AesCrypt::GCM_SUPPORT_VERSION, '<')) {
```
nilmerg, line 104: Use the method in `AesCrypt` that tells you if a tag is required here as well.

library/Icinga/Web/RememberMe.php
```diff
@@ -217,27 +220,44 @@ public function authenticate()
...
      * @return $this
      */
     public function persist($iv = null)
     {
         if ($iv) {
-            $this->remove(bin2hex($iv));
+            self::remove(bin2hex($iv));
         }
 
         $this->getDb()->insert(static::TABLE, [
```
nilmerg, line 263: Just an observation: This now *may* establish a connection to the database twice. (`remove` uses `getDb()` as well but maintains a different instance of `ipl\Sql\Connection`)

nilmerg (changes requested):

library/Icinga/Crypt/AesCrypt.php
sukhwinder33445: I wanted to make sure that the user does not extend the length because the rememberme table cannot handle it if the length is greater than 128.

It's not really necessary.

nilmerg: It's the RememberMe class which accesses the database. AesCrypt is a utility class which doesn't impose storage in the database :wink:

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -50,22 +57,18 @@ class AesCrypt
     /** @var string The authentication tag which is passed by reference when using AEAD cipher mode */
     private $tag;
 
-    /** @var string The cipher method */
```
nilmerg, line 53: Why have you removed the docblock?

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -76,6 +79,28 @@ public function setMethod($method)
         return $this;
     }
 
+    /**
+     * Get supported method
+     *
+     * @return string
+     */
+    public function getSupportedMethod()
```
nilmerg, line 87: I'd rather make this protected. At least, I see no need why it should be of use externally?

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -76,6 +79,28 @@ public function setMethod($method)
...
+    {
+        $availableMethods = openssl_get_cipher_methods();
+        foreach (self::METHODS as $method) {
+            if ($this->isRequiredTag() && ! $this->isAESupported()) {
+                $this->method = null;
+                continue;
+            }
```
nilmerg, line 94: Why is this part of the loop? It does nothing with `$method`, even resets `$this->method` which the entire function body doesn't touch.

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -50,22 +57,18 @@ class AesCrypt
...
 
     const GCM_SUPPORT_VERSION = '7.1';
 
     public function __construct($random_bytes_len = 128)
     {
-        if (version_compare(PHP_VERSION, self::GCM_SUPPORT_VERSION, '<')) {
-            $this->method = 'AES-128-CBC';
-        }
-
+        $this->method = $this->getSupportedMethod();
```
nilmerg, line 66: Didn't I suggest to put this into the getter for the cipher?

The reason why is that constructors should perform as less work as possible. Also, if someone uses `setMethod` first, it won't run unnecessarily.

nilmerg (changes requested):
Didn't look at everything yet again. Remaining comments will follow.

library/Icinga/Crypt/AesCrypt.php
sukhwinder33445: It was a mistake. Fixed

library/Icinga/Crypt/AesCrypt.php
sukhwinder33445: Yes, you are right. Fixed

library/Icinga/Crypt/AesCrypt.php
sukhwinder33445: This checks if the method set by default supports the php version, if not, it simply jumps to the next index 1. So we can override the default method if php does not support it.

nilmerg: I know what it does. But it has no connection to the current iteration in any way. If the first cipher isn't the GCM one, it ignores it which would be wrong.

If you want to invalidate the default method, don't set one. Isn't `getSupportedMethod()` designed to perform this job? If PHP doesn't allow us to use GCM, it shouldn't report it.

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -276,4 +330,26 @@ private function encryptCBC($data)
...
     }
+
+    /**
+     * whether the given method requires a tag
+     *
+     * @param string $method
+     *
+     * @return bool True if required false otherwise
+     */
+    public function isRequiredTag($method)
```
nilmerg, line 341: I'd rather call this `isAuthenticatedEncryptionRequired()` (`isAESupported` would then be `isAuthenticatedEncryptionSupported()`). Also why is this requiring the method with a parameter? I can't see why it should not just use `getMethod()` internally.

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -51,21 +58,15 @@ class AesCrypt
...
-    const GCM_SUPPORT_VERSION = '7.1';
+    private $method;
 
     public function __construct($random_bytes_len = 128)
     {
-        if (version_compare(PHP_VERSION, self::GCM_SUPPORT_VERSION, '<')) {
-            $this->method = 'AES-128-CBC';
-        }
-
         $this->key = random_bytes($random_bytes_len);
```
nilmerg, line 65: While you're at it lightening the constructor... :smile: , why not also generating the key lazily? `getKey()` also throws in case of a missing key, which never will be the case since it's set up here in the constructor. Generating the key in `getKey()` if there's none is then a clear possibility to me. (Of course, you'd need a property for the key length then, but so what?)

library/Icinga/Web/RememberMe.php
```diff
@@ -92,19 +99,18 @@ public static function fromCookie()
...
-                $data[0] = $tag; // encryptedPass
-            } else {
-                $rememberMe->aesCrypt->setTag(base64_decode($tag));
+        if (count($data) > 1) {
+            // php downgrade
+            if (! $rememberMe->aesCrypt->isAESupported()) {
+                throw new RuntimeException(sprintf(
+                    "The given decryption method is not supported in php version '%s'",
+                    PHP_VERSION
+                ));
```
nilmerg, line 108: I'd rather throw this in `setTag()`. Setting a tag for `AesCrypt` on PHP < 7.1 should not be possible.

library/Icinga/Web/RememberMe.php
```diff
@@ -217,20 +223,37 @@ public function authenticate()
             $user->setDomain(Config::app()->get('authentication', 'default_domain'));
         }
 
-        $authenticated = $authChain->authenticate($user, $password);
-        if ($authenticated) {
-            $auth->setAuthenticated($user);
+        foreach (AesCrypt::METHODS as $method) {
+            $this->aesCrypt->setMethod($method);
```
nilmerg, line 227: I believe we said that we don't want a fallback anymore? If `decryptFromBase64()` doesn't work, the user isn't logged in.

nilmerg (changes requested):
You should now use `getMethod()` everywhere. You're still accessing `$this->method` in method `getIV()` at least.

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -177,6 +181,14 @@ public function getIV()
      */
     public function setTag($tag)
     {
+        // php downgrade
```
nilmerg, line 184: Remove this comment. It has no clear relation in this method.

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -177,6 +181,14 @@ public function getIV()
...
+        if (! $this->isAuthenticatedEncryptionSupported()) {
+            throw new RuntimeException(sprintf(
```
nilmerg, line 186: Don't forget the `@throws` annotation for both exceptions :wink:

library/Icinga/Web/RememberMe.php
```diff
@@ -67,6 +67,7 @@ public static function forget()
 
         return (new Cookie(static::COOKIE))
             ->setHttpOnly(true)
+            ->setPath('/')
```
nilmerg: You still have a test related change here.

library/Icinga/Web/RememberMe.php
```diff
@@ -165,12 +158,13 @@ public function getCookie()
...
         ];
 
-        if ($this->aesCrypt->isRequiredTag($this->aesCrypt->getMethod())) {
+        if ($this->aesCrypt->isAuthenticatedEncryptionRequired()) {
             array_splice($values, 1, 0, base64_encode($this->aesCrypt->getTag()));
         }
 
         return (new Cookie(static::COOKIE))
             ->setExpire($this->getExpiresAt())
+            ->setPath('/')
```
nilmerg: Another test related change :wink:

library/Icinga/Web/RememberMe.php
```diff
@@ -214,7 +208,7 @@ public function setExpiresAt($expiresAt)
      */
     public function authenticate()
     {
-        $authenticated = false;
+        $password = $this->aesCrypt->decryptFromBase64($this->encryptedPassword);
```
nilmerg: Please move this directly above line no. 220. You don't need it earlier and only there.

library/Icinga/Web/RememberMe.php
```diff
@@ -265,6 +242,7 @@ public function persist($iv = null)
 
         $this->getDb()->insert(static::TABLE, [
             'username'          => $this->username,
+            'method'          => $this->aesCrypt->getMethod(),
```
nilmerg: Test related change :wink: (I guess)

nilmerg (changes requested):

library/Icinga/Web/RememberMe.php
sukhwinder33445: I think if the password decryption fails, we don't need the rest.

nilmerg (dismissed):
Code wise, LGTM!

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -207,21 +261,21 @@ public function decryptFromBase64($data)
...
     public function encrypt($data)
     {
-        if ($this->method === 'AES-128-CBC') {
-            return $this->encryptCBC($data);
+        if (! $this->isAuthenticatedEncryptionRequired()) {
+            return $this->nonAEEncrypt($data);
         }
 
-        $encrypt = openssl_encrypt($data, $this->method, $this->getkey(), 0, $this->getIV(), $this->tag);
+        $encrypt = openssl_encrypt($data, $this->getMethod(), $this->getkey(), 0, $this->getIV(), $this->tag);
```
nilmerg, line 278: Please use the correct casing for `getKey`

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -244,18 +300,23 @@ public function encryptToBase64($data)
...
-        }
-
         $c = base64_decode($data);
         $hmac = substr($c, 0, 32);
         $data = substr($c, 32);
 
-        $decrypt = openssl_decrypt($data, $this->method, $this->getKey(), 0, $this->getIV());
-        $calcHmac = hash_hmac('sha256', $data, $this->getKey(), true);
+        $decrypt = openssl_decrypt($data, $this->getMethod(), $this->getKey(), 0, $this->getIV());
+        $calcHmac = hash_hmac('sha256', $this->getIV().$data, $this->getKey(), true);
```
nilmerg, line 319: Add a space before and after concatenation operators.

library/Icinga/Crypt/AesCrypt.php
```diff
@@ -264,16 +325,45 @@ private function decryptCBC($data)
...
+     * @param string $data
+     *
+     * @return string encrypted string
+     *
+     * @throws RuntimeException If encryption fails
+     */
+    private function nonAEEncrypt($data)
     {
-        $encrypt = openssl_encrypt($data, $this->method, $this->getkey(), 0, $this->getIV());
+        $encrypt = openssl_encrypt($data, $this->getMethod(), $this->getkey(), 0, $this->getIV());
```
nilmerg, line 339: Please use the correct casing for `getKey`

nilmerg (changes requested):

library/Icinga/Crypt/AesCrypt.php
sukhwinder33445: How is that even possible 🤔 . I always use `PhpStorm's` autocomplete.

nilmerg (approved):

PR #4501, opened 2021-08-04
If appropriate, set the API parameter all_services for schedule-downtime

lippserd:
Since Icinga 2.11.0 the schedule-downtime API supports the
all_services parameter. So far we've always sent a separate
request for scheduling service downtimes. As of Icinga 2.13.0,
these service downtimes are automatically removed when the host
downtimes are removed. Of course, this doesn't work if we don't
use the all_services parameter but send a separate request.
With this commit we set this parameter if the transport is API
and Icinga is equal to or greater than 2.11.0.

In addition, if child_options and all_services were previously set,
a request was sent per host and service. This is now also only a
single request if an API command transport is requested or only
API command transports are configured.

### DIFF
--- a/modules/monitoring/library/Monitoring/Command/Object/ApiScheduleHostDowntimeCommand.php
+++ b/modules/monitoring/library/Monitoring/Command/Object/ApiScheduleHostDowntimeCommand.php
@@ -1,0 +1,40 @@
+<?php
+/* Icinga Web 2 | (c) 2021 Icinga GmbH | GPLv2+ */
+
+namespace Icinga\Module\Monitoring\Command\Object;
+
+/**
+ * Schedule host downtime command for API command transport and Icinga >= 2.11.0 that
+ * sends all_services and child_options in a single request
+ */
+class ApiScheduleHostDowntimeCommand extends ScheduleHostDowntimeCommand
+{
+    /** @var int Whether no, triggered, or non-triggered child downtimes should be scheduled */
+    protected $childOptions;
+
+    protected $forAllServicesNative = true;
+
+    /**
+     * Get child options, i.e. whether no, triggered, or non-triggered child downtimes should be scheduled
+     *
+     * @return int
+     */
+    public function getChildOptions()
+    {
+        return $this->childOptions;
+    }
+
+    /**
+     * Set child options, i.e. whether no, triggered, or non-triggered child downtimes should be scheduled
+     *
+     * @param int $childOptions
+     *
+     * @return $this
+     */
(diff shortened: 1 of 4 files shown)


modules/monitoring/library/Monitoring/Command/Object/ApiScheduleHostDowntimeCommand.php
```diff
@@ -0,0 +1,37 @@
+<?php
+// $name | (c) 2021 Icinga GmbH | GPLv2
```
nilmerg, line 2: Icinga Web 2, GPLv2+

nilmerg (changes requested):

modules/monitoring/library/Monitoring/Command/Renderer/IcingaApiCommandRenderer.php
```diff
@@ -156,25 +158,40 @@ public function renderScheduleDowntime(ScheduleServiceDowntimeCommand $command)
...
-            );
+
+        if ($allServicesCompat) {
+            $commandData = $data + [
+                    'type'        => 'Service',
+                    'filter'      => 'host.name == host_name',
+                    'filter_vars' => [
+                        'host_name' => $command->getObject()->getName()
+                    ]
+                ];
```
nilmerg, line 191: This looks too far indented, but the checks don't fail :man_shrugging:

nilmerg (dismissed):
Way better now :smiley:

modules/monitoring/library/Monitoring/Command/Renderer/IcingaApiCommandRenderer.php
lippserd: Yeah I though so too, but PHPStorm also always reformats that section.

nilmerg: Don't get me going with PHPStorm and formatting indentation, that's broken for me since some time.

lippserd: Do you want me to change it?

nilmerg: I'd do it, but again, the checks don't mind. So it's just my opinion.

lippserd: Done.

nilmerg (approved):

PR #4511, opened 2021-08-20
Check if the column in the row is set for filters

raviks789:
When the column is not set for the filters undefined property exception is thrown in icinga-director. This is a quick fix to solve the issue.

fixes icinga/icingaweb2-module-director#1795


nilmerg (changes requested):
Please don't re-add the TODO. It has never been resolved and won't now.

nilmerg (approved):

PR #4521, opened 2021-08-30
Role: Ensure granted permission doesn't exist within the refusals

### LINKED ISSUE #4520: Permissions doesn't work as expected
## Describe the bug
As you can see in the screenshot, `icinga2` user is not allowed to access any command actions. However, the user can still schedule `Downtimes`, add/remove `Comments`, reschedule `Checks`. I don't think this is my fault, cause as I can see it, the user shouldn't be able to perform such actions.

Bildschirmfoto 2021-08-26 um 14 41 37

Yonas-net:


nilmerg:
I'm afraid but this is not the correct fix. `Role::grants()` only has to report if a permission is granted or not. Although if granted, the permission may also be refused however. But the refusal has to be checked in parallel, as it's done by e.g. `User::can()`.

Think of `Role::grant()` and `Role::denies()` as getters for the respective configuration option. Both may be true at the same time, but this only has a meaning once evaluated and that's not done by `Role::grant()` nor by `Role::denies()`.

Yonas-net:
> But the refusal has to be checked in parallel, as it's done by e.g. `User::can()`

Also, how would be it to check for denies in `ObjectAuthorization::checkGrants()` or does a new method need to be implemented that checks grants and refusals in parallel?

nilmerg:
Nope, that's the exact fix we need :)

PR #4526, opened 2021-09-02
NavigationItem: Rename `resolveMacros()` method to `expandMacros()`

Yonas-net:
As icingadb-web has now introduced its own macro class as a trait and
since the object actions are inheriting from this class, there is
unfortunately an error message that the `resolveMacros()` method
of icingadb-web isn't compatible to this one.


nilmerg:
You can change the visibility of trait methods and this error doesn't occur: https://www.php.net/manual/en/language.oop5.traits.php#language.oop5.traits.visibility

PR #4532, opened 2021-09-09
Vagrant: Puppet: icingaweb2_dev: don't let icingacli run on PHP 5.4

Al2Klimov:
... to avoid failures due to Composer platform checks.


Al2Klimov:
## Before

```
==> default: Notice: /Stage[main]/Icingaweb2_dev/Exec[enable-monitoring-module]/returns: PHP Fatal error:  Composer detected issues in your platform: Your Composer dependencies require a PHP version ">= 5.6.0". You are running 5.4.16. in /usr/share/icinga-php/ipl/vendor/composer/platform_check.php on line 25
```

## After

Purrs.

PR #4535, opened 2021-09-14
JS: Don't close the right column when submitting a form if not re…

Yonas-net:
…quested

resolves https://github.com/Icinga/icingadb-web/issues/284


nilmerg (commented):
I'm not sure about this. The intended effect is basically that of the recently introduced `__CLOSE__` redirect rule. (e65ec1900a7fd2f0b811c10c382ad26fdbbe3953)

I think we should investigate if we can change all cases where redirects are expected to have that effect, to use `__CLOSE__` instead. This isn't a problem for Icinga Web 2 itself and core modules. Other modules such as the Director though, may depend on this as well in which case we cannot remove it.

Yonas-net:
> I think we should investigate if we can change all cases where redirects are expected to have that effect, to use `__CLOSE__` instead.

So how does this help us fixing our problem with the view mode switcher? After all, the right column will still be closed anyway.

nilmerg:
> So how does this help us fixing our problem with the view mode switcher?

The part of the JS that closes the right column and loads the redirect in the left one is removed then, because once everything uses `__CLOSE__` it's obsolete.

PR #4538, opened 2021-09-16
Fix multi modal windows and submit

### LINKED ISSUE #4525: Clicking multiple times on a quick action opens several overlapping modal windows
## Describe the bug
1. Clicking multiple times on a quick action opens several empty overlapping modal windows.
2. Clicking on the submit button multiple times add multiple entries.

## Screenshot
Screenshot 2021-09-01 at 14 49 36

sukhwinder33445:
fix #4525

### DIFF
--- a/public/js/icinga/behavior/modal.js
+++ b/public/js/icinga/behavior/modal.js
@@ -42,6 +42,11 @@
         var $modal = _this.$ghost.clone();
         var $urlTarget = _this.icinga.loader.getLinkTargetFor($a, false);
 
+        _this.modalOpener = event.currentTarget;
+
+        // Disable pointer events to block further function calls
+        _this.modalOpener.style.pointerEvents = 'none';
+
         // Add showCompact, we don't want controls in a modal
         url = _this.icinga.utils.addUrlFlag(url, 'showCompact');
 
@@ -189,6 +194,10 @@
      * @param $modal {jQuery} The modal element
      */
     Modal.prototype.hide = function($modal) {
+        // Remove pointerEvent none style to make the button clickable again
+        this.modalOpener.style.pointerEvents = '';
+        this.modalOpener = null;
+
         $modal.removeClass('active');
         // Using `setTimeout` here to let the transition finish
         setTimeout(function () {
--- a/public/js/icinga/behavior/modal.js
+++ b/public/js/icinga/behavior/modal.js
@@ -90,6 +90,8 @@
         var _this = event.data.self;
         var $form = $(event.currentTarget).closest('form');
         var $modal = $form.closest('#modal');
+        // otherwise the form is submitted several times by clicking the "Submit" button several times
+        $form.find('input:not(:disabled)').prop('disabled', true);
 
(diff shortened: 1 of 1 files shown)


nilmerg (changes requested):

public/js/icinga/behavior/modal.js
```diff
@@ -90,6 +93,8 @@
         var _this = event.data.self;
         var $form = $(event.currentTarget).closest('form');
         var $modal = $form.closest('#modal');
+        // otherwise the form is submitted several times by clicking the "Submit" button several times
+        $form.find('input[type=submit]').prop('disabled', true);
```
nilmerg, line 97: Please disable all form inputs like [here](https://github.com/Icinga/icingaweb2/blob/master/public/js/icinga/loader.js#L133). It may not just be the submit button that is submitting the form.

public/js/icinga/behavior/modal.js
```diff
@@ -36,6 +36,9 @@
      * @returns {boolean}
      */
     Modal.prototype.onModalToggleClick = function(event) {
+        // Disable pointer events to block further function calls
+        event.currentTarget.style.pointerEvents = 'none';
```
nilmerg, line 40: Please move that further below. It's not part of the initialization procedures after all.

public/js/icinga/behavior/modal.js
```diff
@@ -187,6 +192,10 @@
      * @param $modal {jQuery} The modal element
      */
     Modal.prototype.hide = function($modal) {
+        var actionUrl = $modal.find('[action]').attr('action');
+        actionUrl = actionUrl.split('&').slice(0, -1).join(" ");
+        $('a[href="' + actionUrl + '"]').removeAttr("style");
```
nilmerg, line 197: Eeh.. I think now it's the time to consider storing a reference to the modal opener temporarily on the instance while the modal is open. There's just a single modal at a time after all.

Also, don't use jQuery to reset the style. You already didn't need it to define it.

nilmerg (changes requested):

public/js/icinga/behavior/modal.js
```diff
@@ -29,6 +29,11 @@
 
     Modal.prototype = new Icinga.EventListener();
 
+    /**
+     * Element that has opened the modal
+     */
+    let modalOpener;
```
nilmerg, line 35: That's not an instance property. Also, `let` is ECMA2015 only and we still have to support IE.

public/js/icinga/behavior/modal.js
```diff
@@ -187,6 +199,8 @@
      * @param $modal {jQuery} The modal element
      */
     Modal.prototype.hide = function($modal) {
+        modalOpener.removeAttribute('style');
```
nilmerg, line 202: This removes the entire style attribute. You should only reset/remove what you've added. (i.e. `pointerEvents`)

nilmerg (changes requested):

nilmerg (approved):

PR #4558, opened 2021-10-13
Allow module dependencies with OR operator

sukhwinder33445:


flourish86:
With something like this should make clear, what is meant here.

I couldn't reproduce the view you sent me in RC.

Desktop Copy 2

PR #4564, opened 2021-10-20
Migrate: Add possibility to migrate monitoring navigation items

### LINKED ISSUE #4578: Migration for `monitoring`'s host- and service-actions
Once Icinga/icingadb-web#303 is done, we provide our own navigation items `icingadb-host-action` and `icingadb-service-action`.

We should provide a migration command which will migrate `host-action` items to `icingadb-host-action` and `service-action` to `icingadb-service-action`. Macros and filters should be automatically transformed.

Yonas-net:
resolves #4578

### DIFF
--- a/modules/migrate/application/clicommands/NavigationCommand.php
+++ b/modules/migrate/application/clicommands/NavigationCommand.php
@@ -1,0 +1,193 @@
+<?php
+
+namespace Icinga\Module\Migrate\Clicommands;
+
+use Icinga\Application\Config;
+use Icinga\Application\Icinga;
+use Icinga\Application\Logger;
+use Icinga\Cli\Command;
+use Icinga\Data\ConfigObject;
+use Icinga\Exception\NotReadableError;
+use Icinga\Exception\NotWritableError;
+use Icinga\Module\Icingadb\Compat\UrlMigrator;
+use Icinga\Util\DirectoryIterator;
+use Icinga\Web\Request;
+use ipl\Web\Filter\QueryString;
+use ipl\Web\Url;
+
+class NavigationCommand extends Command
+{
+    /**
+     * Migrate local user monitoring navigation items to the Icinga DB Web actions
+     *
+     * USAGE
+     *
+     *  icingacli migrate navigation [options]
+     *
+     * OPTIONS:
+     *
+     *  --user=<username>  Migrate monitoring navigation items only for
+     *                     the given user. (Default *)
+     *
+     *  --delete           Remove the legacy files after successfully
+     *                     migrated the navigation items.
+     */
+    public function indexAction()
+    {
+        $moduleManager = Icinga::app()->getModuleManager();
+        if (! $moduleManager->hasEnabled('icingadb')) {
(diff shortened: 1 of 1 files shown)


modules/migrate/application/clicommands/PreferencesCommand.php
```diff
@@ -102,4 +107,175 @@ public function indexAction()
...
+
+            $urlString = $configObject->get('url');
+            if ($urlString !== null) {
+                $url = Url::fromPath($urlString, [], new Request());
+
+                try {
+                    $urlString = UrlMigrator::transformUrl($url)->getAbsoluteUrl();
+                    $configObject->url = rawurldecode($urlString);
+                } catch (\Exception $err) {
+                    Logger::error($err->getMessage());
```
nilmerg, line 219: I don't think we should show these errors. The user cant' do anything to solve them anyway. Though, please only catch `InvalidArgumentException`, this way real errors really interrupt the process.

modules/migrate/application/clicommands/PreferencesCommand.php
```diff
@@ -102,4 +107,175 @@ public function indexAction()
...
+
+            $legacyFilter = $configObject->get('filter');
+            if ($legacyFilter !== null) {
+                $filter = QueryString::parse($legacyFilter);
+
+                try {
+                    $filter = UrlMigrator::transformFilter($filter);
+                    $configObject->filter = rawurldecode(QueryString::render($filter));
+                } catch (\Exception $err) {
+                    Logger::error($err->getMessage());
```
nilmerg, line 231: This won't throw a `InvalidArgumentException`, so please remove the try..catch block.

nilmerg (changes requested):
Really nice! Only one further thing:

`icingacli migrate preferences` shows now this:

```
Available actions:

  index           Migrate local INI user preferences to a database
  navigation      Migrate local user monitoring navigation items to the Icinga DB Web actions
```

This interferes with the [upgrading documentation](https://icinga.com/docs/icinga-web-2/latest/doc/80-Upgrading/#upgrading-to-icinga-web-2-29x) and is also a not very useful command name: `icingacli migrate preferences index`

I'd rather want to see this as command: `icingacli migrate navigation`

modules/migrate/application/clicommands/NavigationCommand.php
```diff
@@ -0,0 +1,189 @@
...
+
+            $urlString = $configObject->get('url');
+            if ($urlString !== null) {
+                $url = Url::fromPath($urlString, [], new Request());
+
+                try {
+                    $urlString = UrlMigrator::transformUrl($url)->getAbsoluteUrl();
+                    $configObject->url = rawurldecode($urlString);
+                } catch (\InvalidArgumentException $err) {
+                    Logger::error($err->getMessage());
```
nilmerg, line 130: I also said that showing those errors is useless. Do nothing here. (Signal this with a comment: `// pass`)

nilmerg (changes requested):

nilmerg (approved):

PR #4587, opened 2021-11-06
Implement client cert auth for the API client and verify the certificates

dasJ:
This is probably only half-done because it lacks docs and the UI is not perfect (you can currently enter no username or certificate and the config is saved).
The code works and I'm currently running it in production. The second commit fixes a [CWE-295](https://cwe.mitre.org/data/definitions/295.html).


dasJ:
I will not sign the CLA, feel free to create a proper PR with my code

nilmerg:
If you don't sign the CLA, we can't accept your contribution.

PR #4613, opened 2021-12-01
Replace compiled Less colors with CSS var() function calls and generate light mode calls

lippserd:


nilmerg:
By the way, this must not run for IE users. There must be a way to disable the automatism globally.

lippserd:
@nilmerg Ready to review!

library/Icinga/Less/ColorProp.php
```diff
@@ -0,0 +1,100 @@
...
+    {
+        return $this->index;
+    }
+
+    /**
+     * @param int $index
+     *
+     * @return $this
+     */
+    public function setIndex(int $index)
```
nilmerg, line 59: Not compatible with PHP 5.6

library/Icinga/Less/ColorProp.php
```diff
@@ -0,0 +1,100 @@
...
+    {
+        return $this->name;
+    }
+
+    /**
+     * @param string $name
+     *
+     * @return $this
+     */
+    public function setName(string $name)
```
nilmerg, line 79: Not compatible with PHP 5.6

library/Icinga/Less/LightModeDefinition.php
```diff
@@ -0,0 +1,74 @@
...
+    {
+        return $this->name;
+    }
+
+    /**
+     * @param string $name
+     *
+     * @return $this
+     */
+    public function setName(string $name)
```
nilmerg, line 43: Not compatible with PHP 5.6

library/Icinga/Less/ColorProp.php
```diff
@@ -0,0 +1,100 @@
...
+    /** @var Less_Tree_Color Color with which we created the ColorProp */
+    protected $color;
+
+    /** @var int */
+    protected $index;
+
+    /** @var string Color variable name */
+    protected $name;
+
+    private function __construct()
```
nilmerg, line 25: Not compatible with PHP 5.6 (access level change)

nilmerg (changes requested):
Every file is missing a license header. I thought we do this only for ipl content? Here every file has one, why should these differ?

Please fully ensure this doesn't run for IE like it's being done for Javascript [here](https://github.com/Icinga/icingaweb2/blob/master/library/Icinga/Application/webrouter.php#L71-L72).

library/Icinga/Application/webrouter.php
```diff
@@ -56,10 +56,12 @@
 
     switch ($path) {
         case 'css/icinga.css':
-            Stylesheet::send();
+            $forIe11 = (bool) preg_match('/Trident\/7.0;.*rv:11/', $_SERVER['HTTP_USER_AGENT']);
+            Stylesheet::send(false, $forIe11);
```
nilmerg, line 60: `$forIe11` should ne negated, the one below as well

nilmerg (changes requested):

nilmerg (approved):

nilmerg (approved):

PR #4616, opened 2021-12-07
Fix migrating navigation items only for the given user doesn't work

### LINKED ISSUE #4617: Migrating navigation items for a given user also migrates other user's navigation items
## Describe the bug

When you have some directories which are ordered alphabetically and you want to migrate e.g. navigation items for a user, which occur in the middle of these alphabetically arranged directories, all following user directories will be migrated too, because we set the given username to null after each iteration and have no chance to filter the remaining directories.

## Your Environment

* Web browser used:
* Icinga 2 version used (`icinga2 --version`):
* PHP version used (`php --version`):
* Server operating system and version:

yhabteab:
fixes #4617

### DIFF
--- a/modules/migrate/application/clicommands/NavigationCommand.php
+++ b/modules/migrate/application/clicommands/NavigationCommand.php
@@ -50,10 +50,11 @@ class NavigationCommand extends Command
         }
 
         $rc = 0;
-        $username = $this->params->get('user');
+        $user = $this->params->get('user');
         $directories = new DirectoryIterator($preferencesPath);
 
         foreach ($directories as $directory) {
+            $username = $user;
             if ($username !== null && $directories->key() !== $username) {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


nilmerg:
What does this fix? It seems to migrate the user's config just fine without your changes.

Please open an issue and add a detailed description what this will fix and how to reproduce it.

yhabteab:
Let's say you have two users `Anton` and `Berta` and you want to migrate only the ones of `Anton`. The directories are now ordered alphabetically, i.e. Anton comes before Berta. However, [here](https://github.com/Icinga/icingaweb2/blob/master/modules/migrate/application/clicommands/NavigationCommand.php#L69) we reset the given username to null and keep going until the iterator is empty. You are just lucky that the user you tested it with is the last directory.

nilmerg (approved):

PR #4660, opened 2022-02-07
Drop the unhandled services counter in host lists

### LINKED ISSUE #3921: "unhandled services" count host host view differences with different query filters
## Describe the bug
in icingaweb2, when I go to "Tactical Overview" -> "Hosts" and filter for a specific host group, I have found one Host who has no unhandled services shown (the red text in braces).
But when I add the specific host name to the filter (or filter only for this specific host) I get the (correct) number of unhandled services for this host.
Strange: As far, as I can see, this behaviour exists only for this host (can't go through all hosts and check it manually)

Summarizing: You can't rely for the info in braces about unhandled services in "Tactical Overview" -> "Hosts"

## To Reproduce
Don't know how to reproduce. Currently I found only this host with this behaviour

## Expected behavior
the information about unhandled services in host view should always the same correct one 

## Screenshots
Screenshot from 2019-08-15 11-43-18_filter_with_hg
Screenshot from 2019-08-15 11-44-09_filter_with_hg_and_host

## Your Environment

* Icinga Web 2 version and modules (System - About): 2.6.3
* Web browser used: Firefox 68
* Icinga 2 version used (`icinga2 --version`): 2.8.2
* PHP version used (`php --version`): 7.1
* Server operating system and version: rhel 7 (64 Bit)

nilmerg:
It has too much problems. It wasn't even visible since the initial release, at least not if there were many results. It being visible was more a case of chance than guarantee. We now also identified a major issue with restrictions and the way we can assemble queries here. In short, there are too much technical difficulties that we deemed it not worthwhile to keep.

closes #3921


lippserd (commented):
Restoring order makes perfect sense. However, I also think that the grouping is not necessary. Do you remember why you added it?

nilmerg:
Not sure, actually. But:

> This restores it, and puts the columns into the group by definition, though I'm not sure whether the latter is really required. It also works with PostgreSQL in both cases.

So I'll remove it in an instant if you think it's unncessary.

lippserd:
> So I'll remove it in an instant if you think it's unncessary.

I think it's fine if we remove it.

PR #4698, opened 2022-03-09
Set modal title from response header on auto/formsubmit

yhabteab:


library/Icinga/Web/Controller/ActionController.php
```diff
@@ -506,7 +506,7 @@ protected function postDispatchXhr()
                 rawurlencode($this->view->title . ' :: ' . $this->view->defaultTitle),
                 true
             );
-        } else {
+        } elseif ($resp->getHeader('X-Icinga-Title') === null) {
```
nilmerg, line 509: I believe we should always avoid overriding this header. Please encapsulate the entire control structure with this condition. So that even if `$this->view->title` is set the header isn't overridden.

public/js/icinga/behavior/modal.js
```diff
@@ -109,6 +109,9 @@
         req.addToHistory = false;
         req.$redirectTarget = $modal.data('redirectTarget');
         req.done(function (data, textStatus, req) {
+            var title = req.getResponseHeader('X-Icinga-Title');
+            _this.setTitle($modal, decodeURIComponent(title).replace(/\s::\s.*/, ''));
```
nilmerg, line 113: Only set a title if you got one.

nilmerg (changes requested):

public/js/icinga/behavior/modal.js
```diff
@@ -109,6 +109,11 @@
         req.addToHistory = false;
         req.$redirectTarget = $modal.data('redirectTarget');
         req.done(function (data, textStatus, req) {
+            var title = req.getResponseHeader('X-Icinga-Title');
+            if (typeof title !== undefined) {
```
nilmerg, line 113: The `typeof` instruction returns a string. `typeof title !== 'undefined'` would work. But this won't just be `undefined` if the header is missing. The header might exist, but it might not have a value. Use `!! title` instead. This way you'll continue only if whatever it is can be casted to a boolean true.

nilmerg (changes requested):

nilmerg (dismissed):

nilmerg (dismissed):

lippserd:
Wait a moment before merging this PR please. During testing I also got some errors with these line https://github.com/Icinga/icingaweb2/pull/4698/files#diff-daa41d383a500dc3662c13054e5f52570bd82f696be25c8fe7960c80b4a4ffaaR63.

nilmerg:
Is nearly a week enough to wait for your update? :laughing:

yhabteab:
> Wait a moment before merging this PR please. During testing I also got some errors with modal.js:63.

I think I just encountered the same error 😞. Prevents the modal views from being rendered.
Bildschirmfoto 2022-04-14 um 12 17 54

nilmerg:
That's fixed in the master.

PR #4702, opened 2022-03-11
Introduce some enhanced dashboards functionality

### LINKED ISSUE #3809: Enhanced Dashboards
Every authenticated user in Icinga Web 2 is allowed to create own dashboards. Those dashboards contain dashlets which render the configured URL. Every URL/view in Icinga Web 2 is allowed to be rendered in a dashlet. The configured dashboards are saved to per-user INI files. There are some limitations though:

* Having more than one Icinga Web 2 server requires the manual synchronization of the user created INI-files.
* Modules can’t provide dashlets.
* It’s not possible to override/set default dashboards for users.
* It’s not possible share dashboards.
* It’s not possible to reorder dashlets in the frontend.
* Also, it’s not possible to resize them.

yhabteab:
resolves #3809

### DIFF
--- a/application/controllers/DashboardsController.php
+++ b/application/controllers/DashboardsController.php
@@ -1,0 +1,375 @@
+<?php
+
+/* Icinga Web 2 | (c) 2022 Icinga GmbH | GPLv2+ */
+
+namespace Icinga\Controllers;
+
+use GuzzleHttp\Psr7\ServerRequest;
+use Icinga\Forms\Dashboard\DashletForm;
+use Icinga\Forms\Dashboard\HomePaneForm;
+use Icinga\Forms\Dashboard\RemoveDashletForm;
+use Icinga\Forms\Dashboard\RemoveHomePaneForm;
+use Icinga\Forms\Dashboard\WelcomeForm;
+use Icinga\Model\ModuleDashlet;
+use Icinga\Web\Dashboard\Dashboard;
+use Icinga\Web\Dashboard\Dashlet;
+use Icinga\Web\Dashboard\Settings;
+use Icinga\Web\Dashboard\Setup\SetupNewDashboard;
+use Icinga\Web\Navigation\DashboardHome;
+use Icinga\Web\Widget\Tabextension\DashboardSettings;
+use ipl\Html\HtmlElement;
+use ipl\Web\Compat\CompatController;
+use ipl\Web\Url;
+use ipl\Web\Widget\ActionLink;
+
+class DashboardsController extends CompatController
+{
+    /** @var Dashboard */
+    protected $dashboard;
+
+    public function init()
+    {
+        parent::init();
+
+        $this->dashboard = new Dashboard();
+        $this->dashboard->setUser($this->Auth()->getUser());
+        $this->dashboard->setTabs($this->getTabs());
+        $this->dashboard->load();
+    }
+
+    public function indexAction()
+    {
+        $this->createTabs();
+
+        $activeHome = $this->dashboard->getActiveHome();
+        if (!$activeHome || ! $activeHome->hasPanes()) {
+            $this->getTabs()->add('dashboard', [
+                'active' => true,
+                'title' => $this->translate('Welcome'),
+                'url' => Url::fromRequest()
+            ]);
+
+            // Setup dashboard introduction form
+            $welcomeForm = new WelcomeForm($this->dashboard);
+            $welcomeForm->on(WelcomeForm::ON_SUCCESS, function () use ($welcomeForm) {
+                $this->redirectNow($welcomeForm->getRedirectUrl());
+            })->handleRequest(ServerRequest::fromGlobals());
+
+            $this->dashboard->setWelcomeForm($welcomeForm);
+        } elseif (empty($activeHome->getPanes(true))) {
+            // TODO(TBD): What to do when the user has only disabled dashboards? Should we render the welcome screen?
+        } else {
+            if ($this->getParam('pane')) {
+                $pane = $this->getParam('pane');
+                $this->getTabs()->activate($pane);
+            }
+        }
+
+        $this->content = $this->dashboard;
+    }
+
+    /**
+     * Display all the dashboards assigned to a Home set in the `home` request param
+     *
+     * If no pane param is submitted, the default pane is displayed (usually the first one)
+     */
+    public function homeAction()
+    {
+        $home = $this->params->getRequired('home');
+        if (! $this->dashboard->hasHome($home)) {
+            $this->httpNotFound(sprintf(t('Home "%s" not found'), $home));
+        }
+
+        $this->createTabs();
+
+        $activeHome = $this->dashboard->getActiveHome();
+        if (! $activeHome || empty($activeHome->getPanes(true))) {
+            $this->getTabs()->add($home, [
+                'active'    => true,
+                'title'     => $home,
+                'url'       => Url::fromRequest()
+            ]);
+        } elseif (($pane = $this->getParam('param'))) {
+            $this->dashboard->activate($pane);
+        }
+
+        $this->content = $this->dashboard;
+    }
+
+    public function renameHomeAction()
+    {
+        $this->setTitle(t('Update Home'));
+
+        $home = $this->params->getRequired('home');
+        if (! $this->dashboard->hasHome($home)) {
+            $this->httpNotFound(sprintf($this->translate('Home "%s" not found'), $home));
+        }
+
+        $homeForm = new HomePaneForm($this->dashboard);
+        $homeForm->on(HomePaneForm::ON_SUCCESS, function () {
+            $this->redirectNow('__CLOSE__');
+        })->handleRequest(ServerRequest::fromGlobals());
+
+        $homeForm->load($this->dashboard->getActiveHome());
+        $this->addContent($homeForm);
+    }
+
+    public function removeHomeAction()
+    {
+        $this->setTitle(t('Remove Home'));
+
+        $home = $this->params->getRequired('home');
+        if (! $this->dashboard->hasHome($home)) {
+            $this->httpNotFound(sprintf($this->translate('Home "%s" not found'), $home));
+        }
+
+        $homeForm = (new RemoveHomePaneForm($this->dashboard))
+            ->on(RemoveHomePaneForm::ON_SUCCESS, function () {
+                $this->redirectNow('__CLOSE__');
+            })
+            ->handleRequest(ServerRequest::fromGlobals());
+
+        $this->addContent($homeForm);
+    }
+
+    public function editPaneAction()
+    {
+        $this->setTitle(t('Update Pane'));
+
+        $pane = $this->params->getRequired('pane');
+        $home = $this->params->getRequired('home');
+
+        if (! $this->dashboard->hasHome($home)) {
+            $this->httpNotFound(sprintf($this->translate('Home "%s" not found'), $home));
+        }
+
+        if (! $this->dashboard->getActiveHome()->hasPane($pane)) {
+            $this->httpNotFound(sprintf($this->translate('Pane "%s" not found'), $pane));
+        }
+
+        $paneForm = new HomePaneForm($this->dashboard);
+        $paneForm->on(HomePaneForm::ON_SUCCESS, function () use ($pane) {
+            $this->redirectNow('__CLOSE__');
+        })->handleRequest(ServerRequest::fromGlobals());
+
+        $paneForm->load($this->dashboard->getActiveHome()->getPane($pane));
+
+        $this->addContent($paneForm);
+    }
+
+    public function removePaneAction()
+    {
+        $this->setTitle(t('Remove Pane'));
+
+        $home = $this->params->getRequired('home');
+        $paneParam = $this->params->getRequired('pane');
+
+        if (! $this->dashboard->hasHome($home)) {
+            $this->httpNotFound(sprintf($this->translate('Home "%s" not found'), $home));
+        }
+
+        if (! $this->dashboard->getActiveHome()->hasPane($paneParam)) {
+            $this->httpNotFound(sprintf($this->translate('Pane "%s" not found'), $paneParam));
+        }
+
+        $paneForm = new RemoveHomePaneForm($this->dashboard);
+        $paneForm->populate(['org_name' => $paneParam]);
+        $paneForm->on(RemoveHomePaneForm::ON_SUCCESS, function ()  {
+            $this->redirectNow('__CLOSE__');
+        })->handleRequest(ServerRequest::fromGlobals());
+
+        $paneForm->getElement('btn_remove')->setLabel(t('Remove Pane'));
+        $paneForm->prependHtml(HtmlElement::create('h1', null, sprintf(
+            t('Please confirm removal of dashboard pane "%s"'),
+            $paneParam
+        )));
+
+        $this->addContent($paneForm);
+    }
+
+    public function newDashletAction()
+    {
+        $this->setTitle(t('Add Dashlet To Dashboard'));
+
+        $dashletForm = new DashletForm($this->dashboard);
+        $dashletForm->populate($this->getRequest()->getPost());
+        $dashletForm->on(DashletForm::ON_SUCCESS, function () use ($dashletForm) {
+            $this->redirectNow('__CLOSE__');
+        })->handleRequest(ServerRequest::fromGlobals());
+
+        $params = $this->getAllParams();
+        if ($this->getParam('url')) {
+            $params['url'] = rawurldecode($this->getParam('url'));
+        }
+
+        $dashletForm->populate($params);
+
+        $this->addContent($dashletForm);
+    }
+
+    public function editDashletAction()
+    {
+        $this->setTitle(t('Edit Dashlet'));
+
+        $pane = $this->validateDashletParams();
+        $dashlet = $pane->getDashlet($this->getParam('dashlet'));
+
+        $dashletForm = new DashletForm($this->dashboard);
+        $dashletForm->on(DashletForm::ON_SUCCESS, function () use ($dashletForm, $pane) {
+            $this->redirectNow('__CLOSE__');
+        })->handleRequest(ServerRequest::fromGlobals());
+
+        $dashletForm->getElement('submit')->setLabel(t('Update Dashlet'));
+
+        $dashletForm->load($dashlet);
+        $this->addContent($dashletForm);
+    }
+
+    public function removeDashletAction()
+    {
+        $this->validateDashletParams();
+        $this->setTitle(t('Remove Dashlet'));
+
+        $removeForm = (new RemoveDashletForm($this->dashboard))
+            ->on(RemoveDashletForm::ON_SUCCESS, function ()  {
+                $this->redirectNow('__CLOSE__');
+            })
+            ->handleRequest(ServerRequest::fromGlobals());
+
+        $this->addContent($removeForm);
+    }
+
+    public function reorderDashletsAction()
+    {
+        $this->assertHttpMethod('post');
+        if (! $this->getRequest()->isApiRequest()) {
+            $this->httpBadRequest('No API request');
+        }
+
+        if (
+            ! preg_match('/([^;]*);?/', $this->getRequest()->getHeader('Content-Type'), $matches)
+            || $matches[1] !== 'application/json'
+        ) {
+            $this->httpBadRequest('No JSON content');
+        }
+
+        $dashboards = $this->getRequest()->getPost();
+        foreach ($dashboards as $home => $panes) {
+            if (! $this->dashboard->hasHome($home)) {
+                $this->httpNotFound(sprintf($this->translate('Dashboard home "%s" not found'), $home));
+            }
+
+            $home = $this->dashboard->getHome($home);
+            $home->setActive();
+            $home->loadDashboardsFromDB();
+
+            foreach ($panes as $pane => $dashlets) {
+                if (! $home->hasPane($pane)) {
+                    $this->httpNotFound(sprintf($this->translate('Dashboard pane "%s" not found'), $pane));
+                }
+
+                $pane = $home->getPane($pane);
+                foreach ($dashlets as $order => $dashlet) {
+                    if (! $pane->hasDashlet($dashlet)) {
+                        $this->httpNotFound(sprintf($this->translate('Dashlet "%s" not found'), $pane));
+                    }
+
+                    Dashboard::getConn()->update(Dashlet::TABLE, ['priority' => $order], [
+                        'id = ?'            => $pane->getDashlet($dashlet)->getUuid(),
+                        'dashboard_id = ?'  => $pane->getUuid()
+                    ]);
+                }
+            }
+        }
+
+        exit;
+    }
+
+    /**
+     * Provides a mini wizard which guides a new user through the dashboard creation
+     * process and helps them get a first impression of Icinga Web 2.
+     */
+    public function setupDashboardAction()
+    {
+        if (isset($this->getRequest()->getPost()['btn_next'])) {
+            // Set compact view to prevent the controls from being
+            // rendered in the modal view when redirecting
+            $this->view->compact = true;
+
+            $this->getResponse()->setHeader('X-Icinga-Title', 'Configure Dashlets', true);
+        } else {
+            $this->setTitle($this->translate('Add Dashlet'));
+        }
+
+        $query = ModuleDashlet::on(Dashboard::getConn());
+
+        $setupForm = new SetupNewDashboard($this->dashboard);
+        $setupForm->initDashlets(Dashboard::getModuleDashlets($query));
+        $setupForm->on(SetupNewDashboard::ON_SUCCESS, function () use ($setupForm) {
+            if ($setupForm->getPopulatedValue('btn_cancel')) {
+                $this->redirectNow('__CLOSE__');
+            }
+
+            $this->redirectNow($setupForm->getRedirectUrl());
+        })->handleRequest(ServerRequest::fromGlobals());
+
+        $this->addContent($setupForm);
+    }
+
+    public function settingsAction()
+    {
+        $this->createTabs();
+        // TODO(yh): This may raise an exception when the given tab name doesn't exist.
+        //      But as ipl::Tabs() doesn't offer the possibility to check this beforehand, just ignore it for now!!
+        $this->dashboard->activate('dashboard_settings');
+
+        $this->addControl(new ActionLink(
+            $this->translate('Add new Home'),
+            Url::fromPath(Dashboard::BASE_ROUTE . '/new-dashlet'),
+            'plus',
+            [
+                'class'                 => 'add-home',
+                'data-icinga-modal'     => true,
+                'data-no-icinga-ajax'   => true
+            ]
+        ));
+
+        $this->content = new Settings($this->dashboard);
+    }
+
+    /**
+     * Create tab aggregation
+     */
+    private function createTabs()
+    {
+        $tabs = $this->dashboard->getTabs();
+        $activeHome = $this->dashboard->getActiveHome();
+        if ($activeHome && $activeHome->hasPanes()) {
+            $tabs->extend(new DashboardSettings());
+        }
+
+        return $tabs;
+    }
+
+    private function validateDashletParams()
+    {
+        $home = $this->params->getRequired('home');
+        $pane = $this->params->getRequired('pane');
+        $dashlet = $this->params->getRequired('dashlet');
+
+        if (! $this->dashboard->hasHome($home)) {
+            $this->httpNotFound(sprintf($this->translate('Home "%s" not found'), $home));
+        }
+
+        if (! $this->dashboard->getActiveHome()->hasPane($pane)) {
+            $this->httpNotFound(sprintf($this->translate('Pane "%s" not found'), $pane));
+        }
+
+        $pane = $this->dashboard->getActiveHome()->getPane($pane);
+        if (! $pane->hasDashlet($dashlet)) {
+            $this->httpNotFound(sprintf($this->translate('Dashlet "%s" not found'), $dashlet));
+        }
+
+        return $pane;
+    }
+}
--- a/application/controllers/SearchController.php
+++ b/application/controllers/SearchController.php
@@ -3,23 +3,21 @@
 
 namespace Icinga\Controllers;
 
-use Icinga\Web\Controller\ActionController;
-use Icinga\Web\Widget;
 use Icinga\Web\Widget\SearchDashboard;
+use ipl\Web\Compat\CompatController;
 
 /**
  * Search controller
  */
-class SearchController extends ActionController
+class SearchController extends CompatController
 {
     public function indexAction()
     {
         $searchDashboard = new SearchDashboard();
         $searchDashboard->setUser($this->Auth()->getUser());
-        $this->view->dashboard = $searchDashboard->search($this->params->get('q'));
 
-        // NOTE: This renders the dashboard twice. Remove this once we can catch exceptions thrown in view scripts.
-        $this->view->dashboard->render();
+        $this->controls->setTabs($searchDashboard->getTabs());
+        $this->content = $searchDashboard->search($this->getParam('q'));
     }
 
     public function hintAction()
--- a/library/Icinga/Application/Modules/Module.php
+++ b/library/Icinga/Application/Modules/Module.php
@@ -13,11 +13,14 @@ use Icinga\Exception\IcingaException;
 use Icinga\Exception\ProgrammingError;
 use Icinga\Module\Setup\SetupWizard;
 use Icinga\Util\File;
+use Icinga\Web\Dashboard;
 use Icinga\Web\Navigation\Navigation;
-use Icinga\Web\Widget;
 use ipl\I18n\GettextTranslator;
 use ipl\I18n\StaticTranslator;
 use ipl\I18n\Translation;
+use ipl\Web\Url;
+use RecursiveDirectoryIterator;
+use RecursiveIteratorIterator;
 use Zend_Controller_Router_Route;
 use Zend_Controller_Router_Route_Abstract;
 use Zend_Controller_Router_Route_Regex;
@@ -214,9 +217,16 @@ class Module
     /**
      * A set of Pane elements
      *
-     * @var array
+     * @var Dashboard\Pane[]
      */
-    protected $paneItems = array();
+    protected $paneItems = [];
+
+    /**
+     * A set of Dashlet elements
+     *
+     * @var Dashboard\Dashlet[]
+     */
+    protected $dashletItems = [];
 
     /**
      * A set of objects representing a searchUrl configuration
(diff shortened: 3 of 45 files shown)


yhabteab:
For the titles of the modals to work properly in the mini wizard of the first introduction, you will need these changes #4698
! Apart from that you have to drop all Database dashboard tables and import from the `dashboards.sql` file again and you should be good to go! :)

nilmerg (changes requested):
First batch, I need a break now. :sweat_smile:

nilmerg:
Oh, and please fix the style according to the tests.

yhabteab:
> Oh, and please fix the style according to the tests.

I'm not sure what I'm supposed to have done wrong there!!!

nilmerg:
Without looking at the code, I guess you have abstract methods written this way: `public abstract function`
Instead of: `abstract public function`

public/js/icinga/behavior/dashboards.js
```diff
@@ -0,0 +1,215 @@
+/*! Icinga Web 2 | (c) 2022 Icinga GmbH | GPLv2+ */
+
+;(function (Icinga, $) {
+
+    'use strict';
+
+    try {
+        var Sortable = require('icinga/icinga-php-library/vendor/Sortable')
```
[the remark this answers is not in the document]
yhabteab: Those keywords are *block-scoped* and can only be used within this try catch block.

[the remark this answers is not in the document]
nilmerg: Then initialize it outside. `var` is so legacy... :sweat_smile:

public/js/icinga/behavior/dashboards.js
```diff
@@ -0,0 +1,215 @@
...
+    Dashboard.prototype.onRendered = function (e) {
+        let _this = e.data.self;
+        e.target.querySelectorAll('.dashboard-settings, .dashboard.content, .dashboard-item-list, .dashlet-item-list')
+            .forEach(sortable => {
+                let groupName = _this.getTypeFor(sortable),
+                    draggable,
+                    handle;
+
+                switch (groupName) {
+                    case WIDGET_TYPES.DashboardHome: {
```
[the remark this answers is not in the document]
yhabteab: So that I can define and use the same variable names repeatedly 🙈. Don't you like it that way?

application/controllers/SearchController.php
```diff
@@ -3,23 +3,21 @@
...
     public function indexAction()
     {
         $searchDashboard = new SearchDashboard();
         $searchDashboard->setUser($this->Auth()->getUser());
-        $this->view->dashboard = $searchDashboard->search($this->params->get('q'));
 
-        // NOTE: This renders the dashboard twice. Remove this once we can catch exceptions thrown in view scripts.
-        $this->view->dashboard->render();
+        $this->controls->setTabs($searchDashboard->getTabs());
+        $this->content = $searchDashboard->search($this->getParam('q'));
```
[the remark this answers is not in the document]
yhabteab: I don't want to add a content but rather replace the default content by that from search dashboard.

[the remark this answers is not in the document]
nilmerg: Which default content?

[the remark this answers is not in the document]
nilmerg: And even then, there is `setContent()`.

public/js/icinga/behavior/dashboards.js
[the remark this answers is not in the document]
nilmerg: Ah, no. That's a valid use-case. Wouldn't even have used it myself :sweat_smile:

application/controllers/SearchController.php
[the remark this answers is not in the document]
yhabteab: > Which default content?

See [here](https://github.com/Icinga/ipl-web/blob/master/src/Compat/CompatController.php#L60).

> And even then, there is `setContent()`.

Couldn’t found that in `CompatController`.

nilmerg:
The database tables should be in the icingaweb2 database, not a separate one. Adjust the full schema, add an 2.11 upgrade script and adjust the code accordingly.

nilmerg:
Ah oh, it's only the sql file that is outdated...

yhabteab:
> Ah oh, it's only the sql file that is outdated...

Does it occur when creating the foreign keys in `dashlet_system` table? I forgot to move the table to the end!

application/controllers/SearchController.php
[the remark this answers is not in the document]
nilmerg: Aaah, of course `setContent` isn't there. But anyway, the controllers `$content` shouldn't be overridden. There is already an id associated with it, which gets lost now. This may break various other things.

Use `addContent`. Wrap the dashlets in another div and add there `dashboard` as class. Update `widgets.less:122` and your JS accordingly.

nilmerg:
No, yes.., I mean: There's no upgrade file. The full schema file is missing the tables.

And... the dashboards.sql file isn't working as the last two create statements need to be swapped.

nilmerg:
The drag&drop in the editor doesn't work for me. Though, I have no errors in the console and the JS is loaded fine. :thinking: Do I need to enable this somehow first?

yhabteab:
No, actually you don't have to do anything. But the events will be delayed a bit. Maybe that's why?

yhabteab:
Ohh, wait it's actually something wrong with the dashboard reording if there is no dashboard home rendered.

application/controllers/SearchController.php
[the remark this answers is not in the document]
yhabteab: Wouldn't it be simpler to just re-add the id if it's so essential and not to adjust all the js code where all the dashlets are loaded and css styles?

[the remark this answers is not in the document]
nilmerg: And what about it's not just the id but another attribute or class? This isn't a property of the base class without any reason.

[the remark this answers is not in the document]
nilmerg: Also, something may expect `$content` to be an instance of `Content`, which your dashboard is not.

nilmerg:
It works now. But why is there a delay? Feels weird.
Also, I don't think the page needs to be re-rendered (redirected) in case only a dashlet is moved. Or is there technical reason I'm overlooking?

yhabteab:
> It works now. But why is there a delay? Feels weird.

To prevent the events from being triggered immediately, when you actually only want to edit the dashboard/home and click on edit/header.

> Also, I don't think the page needs to be re-rendered (redirected) in case only a dashlet is moved. Or is there technical reason I'm overlooking

We need to render whenever an action is performed. Otherwise we end up in some weird state with the url `*/reorder-widgets` in the search bar, since we are now using the `loader.loadUrl()` method.

nilmerg:
The cancel button in the modals should be `type=button`. Currently it tries to submit the form and the browser shows a validation message for a split-second.

yhabteab:
> Added a dashboard (test), but I cannot add a dashlet?

You have to collapse it first :)

nilmerg:
:man_facepalming:

nilmerg:
The button on the upper right leads to the wrong modal I guess. It looks like it thinks I want to add a dashlet :thinking:

nilmerg:
Oh and, the respective dashboard is not pre-selected if I really want to add a dashlet to it.

nilmerg:
If there is some error, the menu is unusable. Since this feature requires a database upgrade, many users will miss that I'm sure. So a broken menu must be prevented.

Screenshot from 2022-04-08 08-29-39

application/controllers/DashboardsController.php
```diff
@@ -0,0 +1,483 @@
...
+                        t('%s dashlet "%s" successfully'),
+                        $orgPane ? 'Moved' : 'Updated',
+                        $dashlet->getTitle()
+                    ));
+                }
+            }
+        }
+
+        if ($widgetType !== 'Dashlets' || ($orgHome && $orgPane)) {
+            $this->redirectNow($reroutePath);
```
nilmerg, line 388: Replace this with this:

```suggestion
            $this->createTabs();
            $this->dashboard->activate('dashboard_settings');
            $this->sendMultipartUpdate();
```

This will only update the tabs. The dummy content and the header below can also be removed then.
Make also sure to always set `addToHistory=false` and `scripted=true` when calling `loadUrl()` in your JS.

public/css/icinga/dashboards.less
```diff
@@ -0,0 +1,341 @@
...
+    background-color: @tr-active-color;
+  }
+}
+
+.dashlet-item-list.empty-list {
+  margin: 0 0 .9em;
+}
+
+// Dashboard manager
+:not(.dashboard) > .container > .content {
```
nilmerg, line 140: This applies to all other views as well. Please change it.

nilmerg:
Module config pages show this:
```
Uncaught Error: Class "Icinga\Application\Modules\Widget" not found in /icingaweb2/library/Icinga/Application/Modules/Module.php:1054
Stack trace:
#0 /icingaweb2/application/controllers/ConfigController.php(162): Icinga\Application\Modules\Module->getConfigTabs()
#1 /icingaweb2/library/vendor/Zend/Controller/Action.php(507): Icinga\Controllers\ConfigController->moduleAction()
#2 /icingaweb2/library/Icinga/Web/Controller/Dispatcher.php(76): Zend_Controller_Action->dispatch('moduleAction')
#3 /icingaweb2/library/vendor/Zend/Controller/Front.php(937): Icinga\Web\Controller\Dispatcher->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
#4 /icingaweb2/library/Icinga/Application/Web.php(304): Zend_Controller_Front->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
#5 /icingaweb2/library/Icinga/Application/webrouter.php(109): Icinga\Application\Web->dispatch()
#6 /icingaweb2/public/index.php(4): require_once('/icingaweb2/lib...')
#7 {main}
thrown


```

yhabteab:
> Module config pages show this:

How is that even possible 😟?? See https://github.com/Icinga/icingaweb2/pull/4702/commits/e61899fa330267086af711be669adf19e1ceab1b

library/Icinga/Web/Widget/Tabextension/DashboardSettings.php
```diff
@@ -1,39 +1,43 @@
 <?php
-/* Icinga Web 2 | (c) 2014 Icinga Development Team | GPLv2+ */
+/* Icinga Web 2 | (c) 2014-2022 Icinga Development Team | GPLv2+ */
```
nilmerg, line 2: We don't show a from-to, only the from, which is the year the file got introduced.

library/Icinga/Application/Modules/Module.php
```diff
@@ -371,6 +381,37 @@ protected function dashboard($name, array $properties = array())
...
+    /**
+     * Add or get a dashlet
+     *
+     * @param string $name
+     * @param Url|string $url
+     * @param array $properties
+     *
+     * @return DashletContainer
+     */
+    protected function provideDashlet($name, $url, array $properties = [])
```
nilmerg, line 404: Please add a type declaration to param `$name`

library/Icinga/Application/Modules/Module.php
```diff
@@ -307,7 +317,7 @@ public function getSearchUrls()
     /**
      * Return this module's dashboard
      *
-     * @return  Navigation
+     * @return Navigation
```
nilmerg, line 320: This doesn't **have to** return an instance of this. This method is internal, so you can return whatever you see fit for your enhanced dashboards.

nilmerg (changes requested):

library/Icinga/Application/Modules/Module.php
yhabteab: Should I have to add param type declaration of `string, int, bool` in other code places too? I was thinking we do this only `icingadb-web`!

nilmerg: icingadb-web already ditches support for php 5.6 7.0 and 7.1. Web 2 will as well with the update to 2.11.

public/css/icinga/dashboards.less
```diff
@@ -0,0 +1,333 @@
...
+  }
+}
+
+.form-list-control.multi-dashlets:not(.collapsed) span {
+  margin-left: 1.4em;
+}
+
+// Make the submit buttons a bit bigger than the normal size
+.modal-form.icinga-controls .control-group.form-controls input[type="submit"],
+.control-group.form-controls .modal-cancel {
```
nilmerg, line 296: This is a reason to refactor the general form control styles in my opinion.

You have a button inside `.form-controls`. This makes your button a form control, as such it should follow the same layout rules as all the other form controls.
Since your `.modal-cancel` button is a `button[type=button]`, this doesn't happen. `button[type=button]` should also get the same layout rules if inside `.form-controls`. Adjust the file form.less for this and remove the this section here.

public/css/icinga/dashboards.less
```diff
@@ -0,0 +1,333 @@
...
+form.icinga-form .control-group.form-controls .remove-button {
+  .cancel-button();
+  margin-right: auto;
+}
+
+.modal-form .control-group.form-controls .modal-cancel {
+  padding: .5em 1em;
+}
+
+.control-group.form-controls .modal-cancel {
```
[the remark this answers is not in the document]
nilmerg: There's also `.btn-cancel` already defined by form.less. This should also apply to `button[type=button]`. Then you don't need `.modal-cancel` and you'll get the proper default style.

public/css/icinga/dashboards.less
```diff
@@ -0,0 +1,333 @@
...
+// Make the submit buttons a bit bigger than the normal size
+.modal-form.icinga-controls .control-group.form-controls input[type="submit"],
+.control-group.form-controls .modal-cancel {
+  min-width: 7em;
+  font-size: 1.2em;
+  padding: ~"calc(@{vertical-padding} - 2px)" @horizontal-padding;
+}
+
+// Remove and modal cancel buttons
+form.icinga-form .control-group.form-controls .remove-button {
```
nilmerg, line 303: The class `remove-button` is to common to be used in such a equally common rule. Consider that anyone else might use this class at the same location. You cannot assume that the button then also requires an automatic right margin.

There's already `.btn-remove`, use that instead. You could also add the right margin to its definition in form.less. Why shouldn't remove buttons not always be aligned to the left? There take also a look into the reporting module's report modification form where such a button also exists and it wouldn't look bad (imho) if this applies there as well.

nilmerg (changes requested):

public/css/icinga/dashboards.less
```diff
@@ -0,0 +1,333 @@
...
+    width: 100%;
+  }
+}
+
+.form-list-control.multi-dashlets:not(.collapsed) span {
+  margin-left: 1.4em;
+}
+
+// Make the submit buttons a bit bigger than the normal size
+.modal-form.icinga-controls .control-group.form-controls input[type="submit"],
```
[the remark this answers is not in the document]
yhabteab: Bildschirmfoto 2022-04-12 um 11 38 15

public/css/icinga/dashboards.less
yhabteab: Should the `.cancel-button()` mixin be applied to all `btn-remove` buttons by default instead of the `.button()` mixin. Because it makes no sense at all to have cancel buttons styled **red**, but not remove buttons.

public/css/icinga/dashboards.less
yhabteab: The point here is not that my `.modal-cancel` doesn't have the same layout as all other from-controls, but rather the overall buttons size no matter what kind. All form-controls submit buttons don't look as big by default as in the mockups in the referenced issue.

public/css/icinga/dashboards.less
nilmerg: Forget `.cancel-button()`. It's mis-named, yes. Just apply `btn-remove` and it's fine.

nilmerg:
Just got this:

```
Fatal error: Uncaught TypeError: Icinga\Web\Navigation\Navigation::merge(): Argument #1 ($navigation) must be of type Icinga\Web\Navigation\Navigation, array given, called in /icingaweb2/library/Icinga/Web/Navigation/Navigation.php on line 432 and defined in /icingaweb2/library/Icinga/Web/Navigation/Navigation.php on line 374
--
1 | 0.0016 | 406752 | {main}( ) | .../index.php:0
2 | 0.0022 | 430568 | require_once( '/icingaweb2/library/Icinga/Application/webrouter.php ) | .../index.php:4
3 | 0.0668 | 4354528 | Icinga\Application\Web->dispatch( ) | .../webrouter.php:109
4 | 0.0668 | 4354528 | Zend_Controller_Front->dispatch( $request = class Icinga\Web\Request { protected $_dispatched = TRUE; protected $_module = 'default'; protected $_moduleKey = 'module'; protected $_controller = 'dashboard'; protected $_controllerKey = 'controller'; protected $_action = 'index'; protected $_actionKey = 'action'; protected $_params = ['controller' => 'dashboard', 'action' => 'index', 'module' => 'default']; protected $_paramSources = [0 => '_GET', 1 => '_POST']; protected $_requestUri = '/icingaweb2/dashboard'; protected $_baseUrl = '/icingaweb2'; protected $_basePath = NULL; protected $_pathInfo = '/dashboard'; protected $_rawBody = NULL; protected $_aliases = []; protected $response = NULL; protected $uniqueId = NULL; protected $url = class Icinga\Web\Url { protected $external = NULL; protected $params = class Icinga\Web\UrlParams { protected $separator = '&'; protected $params = [...]; protected $index = [...] }; protected $anchor = ''; protected $path = 'dashboard'; protected $basePath = '/icingaweb2'; protected $host = NULL; protected $port = NULL; protected $scheme = NULL; protected $username = NULL; protected $password = NULL }; protected $user = class Icinga\User { protected $firstname = NULL; protected $lastname = NULL; protected $email = NULL; protected $localUsername = 'icingaadmin'; protected $domain = NULL; protected $additionalInformation = ['backend_name' => 'icingaweb2', 'backend_type' => 'db', 'assigned_roles' => [...], 'icingadb.view_modes' => [...]]; protected $externalUserInformation = []; protected $unrestricted = TRUE; protected $permissions = ['*' => '*']; protected $restrictions = []; protected $groups = ['admin' => 'admin']; protected $roles = [0 => class Icinga\Authentication\Role { ... }]; protected $preferences = class Icinga\User\Preferences { protected $preferences = [...] }; protected $isHttpUser = FALSE } }, $response = class Icinga\Web\Response { protected $_body = []; protected $_exceptions = []; protected $_headers = []; protected $_headersRaw = []; protected $_httpResponseCode = 200; protected $_isRedirect = FALSE; protected $_renderExceptions = FALSE; public $headersSentThrowsException = TRUE; protected $autoRefreshInterval = NULL; protected $cookies = NULL; protected $redirectUrl = NULL; protected $request = NULL; protected $reloadWindow = NULL; protected $reloadCss = NULL; protected $rerenderLayout = FALSE; protected $overrideWindowId = FALSE } ) | .../Web.php:304
5 | 0.0694 | 4472656 | Icinga\Web\Controller\Dispatcher->dispatch( $request = class Icinga\Web\Request { protected $_dispatched = TRUE; protected $_module = 'default'; protected $_moduleKey = 'module'; protected $_controller = 'dashboard'; protected $_controllerKey = 'controller'; protected $_action = 'index'; protected $_actionKey = 'action'; protected $_params = ['controller' => 'dashboard', 'action' => 'index', 'module' => 'default']; protected $_paramSources = [0 => '_GET', 1 => '_POST']; protected $_requestUri = '/icingaweb2/dashboard'; protected $_baseUrl = '/icingaweb2'; protected $_basePath = NULL; protected $_pathInfo = '/dashboard'; protected $_rawBody = NULL; protected $_aliases = []; protected $response = NULL; protected $uniqueId = NULL; protected $url = class Icinga\Web\Url { protected $external = NULL; protected $params = class Icinga\Web\UrlParams { protected $separator = '&'; protected $params = [...]; protected $index = [...] }; protected $anchor = ''; protected $path = 'dashboard'; protected $basePath = '/icingaweb2'; protected $host = NULL; protected $port = NULL; protected $scheme = NULL; protected $username = NULL; protected $password = NULL }; protected $user = class Icinga\User { protected $firstname = NULL; protected $lastname = NULL; protected $email = NULL; protected $localUsername = 'icingaadmin'; protected $domain = NULL; protected $additionalInformation = ['backend_name' => 'icingaweb2', 'backend_type' => 'db', 'assigned_roles' => [...], 'icingadb.view_modes' => [...]]; protected $externalUserInformation = []; protected $unrestricted = TRUE; protected $permissions = ['*' => '*']; protected $restrictions = []; protected $groups = ['admin' => 'admin']; protected $roles = [0 => class Icinga\Authentication\Role { ... }]; protected $preferences = class Icinga\User\Preferences { protected $preferences = [...] }; protected $isHttpUser = FALSE } }, $response = class Icinga\Web\Response { protected $_body = []; protected $_exceptions = []; protected $_headers = []; protected $_headersRaw = []; protected $_httpResponseCode = 200; protected $_isRedirect = FALSE; protected $_renderExceptions = FALSE; public $headersSentThrowsException = TRUE; protected $autoRefreshInterval = NULL; protected $cookies = NULL; protected $redirectUrl = NULL; protected $request = NULL; protected $reloadWindow = NULL; protected $reloadCss = NULL; protected $rerenderLayout = FALSE; protected $overrideWindowId = FALSE } ) | .../Front.php:937
6 | 0.0723 | 4722336 | Icinga\Web\Controller\ActionController->__construct( $request = class Icinga\Web\Request { protected $_dispatched = TRUE; protected $_module = 'default'; protected $_moduleKey = 'module'; protected $_controller = 'dashboard'; protected $_controllerKey = 'controller'; protected $_action = 'index'; protected $_actionKey = 'action'; protected $_params = ['controller' => 'dashboard', 'action' => 'index', 'module' => 'default']; protected $_paramSources = [0 => '_GET', 1 => '_POST']; protected $_requestUri = '/icingaweb2/dashboard'; protected $_baseUrl = '/icingaweb2'; protected $_basePath = NULL; protected $_pathInfo = '/dashboard'; protected $_rawBody = NULL; protected $_aliases = []; protected $response = NULL; protected $uniqueId = NULL; protected $url = class Icinga\Web\Url { protected $external = NULL; protected $params = class Icinga\Web\UrlParams { protected $separator = '&'; protected $params = [...]; protected $index = [...] }; protected $anchor = ''; protected $path = 'dashboard'; protected $basePath = '/icingaweb2'; protected $host = NULL; protected $port = NULL; protected $scheme = NULL; protected $username = NULL; protected $password = NULL }; protected $user = class Icinga\User { protected $firstname = NULL; protected $lastname = NULL; protected $email = NULL; protected $localUsername = 'icingaadmin'; protected $domain = NULL; protected $additionalInformation = ['backend_name' => 'icingaweb2', 'backend_type' => 'db', 'assigned_roles' => [...], 'icingadb.view_modes' => [...]]; protected $externalUserInformation = []; protected $unrestricted = TRUE; protected $permissions = ['*' => '*']; protected $restrictions = []; protected $groups = ['admin' => 'admin']; protected $roles = [0 => class Icinga\Authentication\Role { ... }]; protected $preferences = class Icinga\User\Preferences { protected $preferences = [...] }; protected $isHttpUser = FALSE } }, $response = class Icinga\Web\Response { protected $_body = []; protected $_exceptions = []; protected $_headers = []; protected $_headersRaw = []; protected $_httpResponseCode = 200; protected $_isRedirect = FALSE; protected $_renderExceptions = FALSE; public $headersSentThrowsException = TRUE; protected $autoRefreshInterval = NULL; protected $cookies = NULL; protected $redirectUrl = NULL; protected $request = NULL; protected $reloadWindow = NULL; protected $reloadCss = NULL; protected $rerenderLayout = FALSE; protected $overrideWindowId = FALSE }, $invokeArgs = ['displayExceptions' => '1'] ) | .../Dispatcher.php:59
7 | 0.0794 | 5044696 | Icinga\Controllers\DashboardController->init( ) | .../ActionController.php:170
8 | 0.0799 | 5090192 | Icinga\Web\Widget\Dashboard->load( ) | .../DashboardController.php:35
9 | 0.0809 | 5146256 | Icinga\Web\Navigation\Navigation->load( $type = 'dashboard-pane' ) | .../Dashboard.php:74
10 | 0.0871 | 5281216 | Icinga\Web\Navigation\Navigation->merge( $navigation = [] ) | .../Navigation.php:432
```

nilmerg:
Oh, nevermind, accessed the legacy dashboard route..

public/css/icinga/dashboards.less
nilmerg: Ooof, mockups are an artist's interpretation. Just like artworks. Well, no, artworks are way more sophisticated.. Didn't you notice that some of the button labels are randomly placed? These are quickly assembled, to get a rough idea how something should look like. Most of the time this is about the general layout, colors and such. But of course, this is a part of a broader application, that has it's default styles. These modals are in no way expected to **not** use these default styles. It's the exact opposite, they **have to** to use them, where applicable.

If a mockup looks 100% like the end result. The mockup is good, not the end result.

application/controllers/DashboardsController.php
```diff
@@ -0,0 +1,465 @@
...
+     * process and helps them get a first impression of Icinga Web 2.
+     */
+    public function setupDashboardAction()
+    {
+        if (isset($this->getRequest()->getPost()['btn_next'])) {
+            // Set compact view to prevent the controls from being
+            // rendered in the modal view when redirecting
+            $this->view->compact = true;
+
+            $this->getResponse()->setHeader('X-Icinga-Title', t('Configure Dashlets'), true);
```
nilmerg, line 388: You can use `setTitle()` here now. It will work just fine once you've rebased with master.

public/css/icinga/forms.less
```diff
@@ -305,12 +305,15 @@ form.icinga-form .form-controls {
     .button();
   }
 
-  input[type="submit"].btn-remove {
-    .button(@body-bg-color, @color-critical, @color-critical-accentuated);
+  input[type="submit"].btn-remove,
+  button[type="submit"].btn-remove {
+    margin-right: auto;
+    .cancel-button();
```
nilmerg, line 311: I thought I made it clear. Forget `cancel-button`! It's flawed and doesn't match with the `.button` mixin. Leave the styling as it is please.

public/css/icinga/forms.less
```diff
@@ -329,6 +332,16 @@ form.icinga-form {
...
+
+  // Button elements of type button inside a form controls
+  .form-controls button[type="button"] {
+    .button();
+  }
+
+  .control-group button[type="button"]:hover {
+    color: @text-color-on-icinga-blue;
+    background-color: @icinga-blue;
+  }
```
nilmerg, line 344: You don't need these two blocks. Since you've added the `button[type=button]` above where `.btn-remove` styles are applied, the hover effect is applied correctly and the style anyway. Cancel buttons shouldn't look like standard buttons, they don't change anything after all. Remove this.

nilmerg (changes requested):

nilmerg:
I know, the mockups don't show how adding a dashlet from the editor should look like, but I think this should work the same as in the welcome screen. Otherwise the feature of having the pre-defined dashlets suggested is lost forever after the welcome screen.

nilmerg:
Oh, and what I've just noticed, there is no way to remove the default home. Not unless there is another home. Yeah, we've decided to not show the header for the default home if it's the only home, but it should still be possible to remove it somehow.

nilmerg:
Oh and, also strange is that if I remove the default home while another one is still there, it just gets emptied, the dahslets/dashboards are removed, but the home itself is still there. Attempting to remove this empty default home then also doesn't remove it. Only after I've removed the only remaining custom home, both homes are gone and the welcome screen shows up.

nilmerg:
Just pulled your last two commits, now I see this when accessing /icingaweb2/dashboards:

```
SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'Default Home' for key 'icingaweb_dashboard_home.idx_dashboard_home_name'
#0 /usr/local/src/ipl-sql/src/Connection.php(402): PDOStatement->execute(Array)
#1 /usr/local/src/ipl-sql/src/Connection.php(438): ipl\Sql\Connection->prepexec(String)
#2 /icingaweb2/library/Icinga/Web/Dashboard/Common/DashboardManager.php(176): ipl\Sql\Connection->insert(String, Array)
#3 /icingaweb2/library/Icinga/Web/Dashboard/Common/DashboardManager.php(208): Icinga\Web\Dashboard\Dashboard->manageEntry(Object(Icinga\Web\Dashboard\DashboardHome))
#4 /icingaweb2/library/Icinga/Web/Dashboard/Common/DashboardManager.php(89): Icinga\Web\Dashboard\Dashboard->initGetDefaultHome()
#5 /icingaweb2/library/Icinga/Web/Dashboard/Common/DashboardManager.php(46): Icinga\Web\Dashboard\Dashboard->loadDashboardEntries()
#6 /icingaweb2/application/controllers/DashboardsController.php(39): Icinga\Web\Dashboard\Dashboard->load()
#7 /icingaweb2/library/Icinga/Web/Controller/ActionController.php(170): Icinga\Controllers\DashboardsController->init()
#8 /icingaweb2/library/Icinga/Web/Controller/Dispatcher.php(59): Icinga\Web\Controller\ActionController->__construct(Object(Icinga\Web\Request), Object(Icinga\Web\Response), Array)
#9 /icingaweb2/library/vendor/Zend/Controller/Front.php(937): Icinga\Web\Controller\Dispatcher->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
#10 /icingaweb2/library/Icinga/Application/Web.php(304): Zend_Controller_Front->dispatch(Object(Icinga\Web\Request), Object(Icinga\Web\Response))
#11 /icingaweb2/library/Icinga/Application/webrouter.php(109): Icinga\Application\Web->dispatch()
#12 /icingaweb2/public/index.php(4): require_once(String)
#13 {main}
```

yhabteab:
> Just pulled your last two commits, now I see this when accessing /icingaweb2/dashboards:

There is a schema update.

nilmerg:
Found a new bug:

1. Start on the welcome screen
2. Use the defaults
3. Add a new home
4. Remove the default home
5. Add a dashboard to the new home
6. Add a single dashlet to the dashboard
7. Notice that all dashlets got added to the dashboard

nilmerg:
Oh, and after that the welcome screen still appears. Shouldn't my own home now be the primary home?
Screenshot from 2022-04-14 10-24-31

yhabteab:
> Oh, and after that the welcome screen still appears. Shouldn't my own home now be the primary home?

Well, we agreed with @lippserd that if the user removes all dashboards from the `Default Home`, the welcome screen will be displayed over again. Because the default home is not really gone, the user just doesn't get to see it. And in the base route view only dashboards from the default home is rendered.

yhabteab:
> Found a new bug:

Jap, fixed.

library/Icinga/Application/Modules/DashboardContainer.php
```diff
@@ -35,7 +35,11 @@ public function setDashlets(array $dashlets)
      */
     public function getDashlets()
     {
-        return $this->dashlets ?: array();
+        uasort($this->dashlets, function (array $x, array $y) {
+            return $x['priority'] - $y['priority'];
```
nilmerg, line 39: You should return a -1, 0 or 1 here, as usually expected. There's also the spaceship operator since php 7 which simplifies this: `<=>`

nilmerg (changes requested):

public/js/icinga/behavior/dashboards.js
```diff
@@ -152,33 +141,29 @@
...
-                        invertSwap : true,
-                        delay      : 100,
-                        dataIdAttr : 'id',
-                        direction  : 'vertical',
-                        draggable  : draggable,
-                        handle     : handle,
-                        group      : { name : groupName }
+                        scroll      : true,
+                        invertSwap  : true,
+                        delay       : 100,
```
nilmerg, line 160: The plan was to get rid of this delay then ;)

yhabteab: 🤦‍♂️

library/Icinga/Common/DataExtractor.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace Icinga\Common;
+
+trait DataExtractor
```
nilmerg, line 5: I'm not sure about the justification of this trait. Method `toArray` is used once and not implemented anyway, so can be dropped. `fromArray` is named like a factory method but isn't one (doesn't return a new object). The trait itself is also only used once. I'd like to propose that you remove this trait and add the functionality of `fromArray` as part of the method `BaseDashboard::setProperties` and use that instead.

Though, please only where the code really doesn't know which properties need to be set. There are some usages in the method `loadDashboardEntries` of various classes, which should exactly know which setters to call, so they should. Helps also to know for sure that those setters are called and where if inspected using the IDE.

library/Icinga/Web/Dashboard/Common/ModuleDashlet.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+/* Icinga Web 2 | (c) 2022 Icinga GmbH | GPLv2+ */
+
+namespace Icinga\Web\Dashboard\Common;
+
+trait ModuleDashlet
```
nilmerg, line 7: Is this supposed to be used multiple times? Currently there is a single usage. If it keeps that way, I don't see why these methods should be in a trait.

library/Icinga/Web/Dashboard/Common/DashboardControls.php
```diff
@@ -0,0 +1,134 @@
...
+/* Icinga Web 2 | (c) 2022 Icinga GmbH | GPLv2+ */
+
+namespace Icinga\Web\Dashboard\Common;
+
+use Icinga\Exception\ProgrammingError;
+
+use Icinga\Web\Dashboard\Dashboard;
+use function ipl\Stdlib\get_php_type;
+
+trait DashboardControls
```
nilmerg, line 12: This name is mis-leading. `DashboardEntries` seems more fitting to me.

nilmerg:
> > Added a dashboard (test), but I cannot add a dashlet?

> You have to expand it first :)

I know now why I didn't notice that. Github uses a different implementation but with the same icons.

**Expanded**
Screenshot from 2022-04-21 09-52-19

**Collapsed**
Screenshot from 2022-04-21 09-52-11

Should we do it the same as Github does it? Or is this just a Github thing and not standard? @flourish86

nilmerg:
1. Add a new dashlet
2. Use a custom url
3. Now choose a different home option
4. The dashlet options disappear permanently

library/Icinga/Web/Dashboard/Common/DashboardControls.php
```diff
@@ -0,0 +1,134 @@
...
+        return array_key_exists($name, $this->dashboards);
+    }
+
+    public function getEntries()
+    {
+        // An entry can be added individually afterwards, it might be the case that the priority
+        // order gets mixed up, so we have to sort things here before being able to render them
+        uasort($this->dashboards, function (BaseDashboard $x, BaseDashboard $y) {
+            return $x->getPriority() <=> $y->getPriority();
+        });
```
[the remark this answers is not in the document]
nilmerg: Currently wondering why these aren't sorted by the database already anyway :thinking:

yhabteab:
### Requires

- https://github.com/Icinga/ipl-orm/pull/60
- https://github.com/Icinga/ipl-sql/pull/56

sukhwinder33445 (commented):
I found the following bugs:

- [x] Adding same dashlets for `Icingadb` and `Monitoring` module just add one `Pane`.

  Steps to reproduce:

      * Click on `Add Dashlets now` button on welcome page.
      * Select `Host Problems` from `Icingadb` and `Monitoring` module.
      * Save and you can see, just one Pane have been created.

  The same behavior even if I change the name of a pane when creating it.

- [ ] It should be possible to add and edit a description.
- [ ] The cli command `icingacli migrate dashboards` can add the same pane title multiple times if executed multiple times.
    If the name already exists, it will ask for a new name but use the same title. Either the name should also be used as the title, or it asks for the title as well.

nilmerg:
> It should be possible to add and edit a description.

Don't make this a priority. It's okay if we don't provide this with the first iteration.

gbin2265:
Hello,

I made a copy and tested your dashboard.

It's already fantastic that the info is written into the database.
In our case we have an F5 loadbalancer with 2 x icingaweb2 servers behind it (we will eventually increase this to 4 if necessary)
Every time a dashboard is modified, we have to copy it to all these icingaweb2 servers

What I see is that this is based on users only, or am I mistaken?

We are in an environment of more than 500 users (connected via Microsoft AD) who are in ROLES. (user -> group AD -> ROLE)

I have made a proposal for modifications of icingaweb2
https://github.com/Icinga/icingaweb2/pull/5074
We can now create dashboards that are based on a ROLES.

Are there any plans to also make ROLE based dashboards?

regards,
Geert

yhabteab:
> What I see is that this is based on users only, or am I mistaken?

Hii, that's correct. This PR implements (rewrites the existing dashboard implementation logic) just the basics (stores the dashboards in the database, adds so called `Dashboard Homes` that allow you to have multiple entry points for the dashboards, not just the default one, and many other UI features that weren't possible with the legacy implementation).

> Are there any plans to also make ROLE based dashboards?

Yes! This PR is just the baseline for all these kind of features like sharing/subscribing dashboards from others and also as an Icinga Web 2 admin creating default dashboards based on `Role` and/or user groups. You can also check the referenced issue #3809  for a full feature overview and some nice mockups.

Issue #4710, opened 2022-03-23
Mouseover not working in the dashboard


DasWiesel23:
## Describe the bug
Mouseover is not working anymore on service problems in the dashboard.

## To Reproduce
Have multiple service problems in the dashboard at the same time. Try to move the mouse over one service. It should get highlighted and if you click on it, it should get expanded on the right side of your screen.

## Expected behavior
Mouseover a service problem, click on it and see how the details are getting displayed on the right side.
Furthermore hold shift and select multiple problems with the mouse to acknowledge, downtime or comment multiple problems at the same time.

## Screenshots
Using the following URL you'll see what section I mean in exact => https://ibb.co/dBnsrvW

## Your Environment
Icinga Web 2 Version: 2.10.0
Git commit: f1dc03fd8553fe58ef85ca6fb607fe2922c3e078
PHP Version: 7.4.28
Git commit date: 2022-03-23
icinga2: r2.13.2-1
O/S: Debian 11

moreamazingnick:
my multiselect works but only if access is directly like https://icinga.local/icingaweb2/monitoring/list/services?service_problem=1&sort=service_severity

it doesnt work in dashboards like https://icinga.local/icingaweb2/dashboard

Virsacer:
Also multiselect with Strg or Shift is affected.
But this seems to be limited to the dashboard.

moreamazingnick:
the href attribute in the tablerows in the dashboards are missing

`<table data-base-target="_next" class="table-row-selectable state-table multiselect compact" data-icinga-multiselect-url="/icingaweb2/monitoring/services/show" data-icinga-multiselect-controllers="/icingaweb2/monitoring/services" data-icinga-multiselect-data="service,host">
...
<tbody>
<tr></tr>
`

nilmerg:
Hi,

confirmed, thanks!

PR #4719, opened 2022-03-24
Fix missing href in <tr> for dashboards refs #4710

moreamazingnick:


nilmerg:
There's a different change required. I'm already on it.

PR #4724, opened 2022-03-25
Fix missing global theme check refs #4723

### LINKED ISSUE #4723: Theme Mode switch enabled on theme without dark/light
## Describe the bug
The Theme Mode switch becomes enabled even if the theme does not have dark/light.

## To Reproduce
1. Set a theme globally that does not have dark/light
2. Set the same theme in the user settings

## Expected behavior
The Theme Mode switch should  stay disabled

## Screenshots
image
image

## Your Environment
* Icinga Web 2 version and modules (System - About): 2.10.0
* Web browser used: Firefox 99.0b7
* Icinga 2 version used (`icinga2 --version`): 2.13.2-1
* PHP version used (`php --version`): 8.0.8
* Server operating system and version: Ubuntu 21.10

moreamazingnick:
if  global theme is selected and theme chooser is disabled, ther should be a check on the system default theme which is the selected theme, if there is dark/lightmode support.
fixes #4723

### DIFF
--- a/application/forms/PreferenceForm.php
+++ b/application/forms/PreferenceForm.php
@@ -212,7 +212,7 @@ class PreferenceForm extends Form
         } else {
             $themeFile = StyleSheet::getThemeFile(Config::app()->get('themes', 'default'));
         }
-        
+
         if (isset($formData['theme']) && $formData['theme'] !== StyleSheet::DEFAULT_THEME) {
             $themeFile = StyleSheet::getThemeFile($formData['theme']);
         }
--- a/application/forms/PreferenceForm.php
+++ b/application/forms/PreferenceForm.php
@@ -186,7 +186,7 @@ class PreferenceForm extends Form
             );
         }
 
-        $themeFile = StyleSheet::getThemeFile(Config::app()->get('themes', 'default'));        
+        $themeFile = StyleSheet::getThemeFile(Config::app()->get('themes', 'default'));
         if (! (bool) Config::app()->get('themes', 'disabled', false)) {
             $themes = Icinga::app()->getThemes();
             if (count($themes) > 1) {
--- a/application/forms/PreferenceForm.php
+++ b/application/forms/PreferenceForm.php
@@ -186,7 +186,7 @@ class PreferenceForm extends Form
             );
         }
 
-        $themeFile = null;        
+        $themeFile = StyleSheet::getThemeFile(Config::app()->get('themes', 'default'));        
         if (! (bool) Config::app()->get('themes', 'disabled', false)) {
             $themes = Icinga::app()->getThemes();
             if (count($themes) > 1) {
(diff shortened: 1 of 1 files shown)


application/forms/PreferenceForm.php
```diff
@@ -208,8 +208,15 @@ public function createElements(array $formData)
                     )
                 );
             }
+        } else {
+            $globalThemeFile = Config::app()->get('themes', 'default');
+            $themeFile = StyleSheet::getThemeFile($globalThemeFile);
+            $file = $themeFile !== null ? @file_get_contents($themeFile) : false;
+            if ($file && strpos($file, StyleSheet::LIGHT_MODE_IDENTIFIER) === false) {
+                $disabled = ['', 'light', 'system'];
+            }
```
nilmerg, line 217: This is nearly the same as the structure for the user's theme. Please combine both.

```
$theme = global theme or null;
if (user has chosen theme)
    $theme = user chosen theme

if ($theme !== null && $theme !== default theme)
    ...
```

application/forms/PreferenceForm.php
```diff
@@ -208,8 +208,15 @@ public function createElements(array $formData)
...
         }
-
+
```
nilmerg, line 219: Please don't add superfluous whitespace.

nilmerg (changes requested):
Thanks! Left you some comments :)

application/forms/PreferenceForm.php
```diff
@@ -185,7 +185,8 @@ public function createElements(array $formData)
                 false
             );
         }
-
+        $themeFile = null;
+
```
nilmerg, line 189: We require empty lines after closed control structures if any instructions follow. Though, none when starting control structures, so if you flip both lines around I'm fine with it.

application/forms/PreferenceForm.php
nilmerg: I mean the indentation, not the empty line itself, if you've wondered.

nilmerg: I can still see the indentation: :thinking: 
Screenshot from 2022-03-29 12-45-06

nilmerg (approved):
Thank you!

application/forms/PreferenceForm.php
```diff
@@ -208,11 +209,16 @@ public function createElements(array $formData)
                     )
                 );
             }
+        } else {
+            $themeFile = StyleSheet::getThemeFile(Config::app()->get('themes', 'default'));
```
nilmerg, line 213: Hm, I was too fast. This runs only if the user cannot change the theme. But that shouldn't affect the mode selection anyway. Already initialize `$themeFile` with the global default theme. Don't do it here.

nilmerg (changes requested):

application/forms/PreferenceForm.php
moreamazingnick: I like that even better

nilmerg:
The checks fail, please fix the trailing white-space at line 189.

nilmerg (approved):

PR #4752, opened 2022-04-21
history.js: Properly handle data attributes upon navigation

nilmerg:
This is required so that attributes like `data-icinga-container-id` are properly managed if the user navigates using the history buttons. Previously these were not managed at all as `applyLocationBar` didn't made use of ui.js' capabilities to paste/cut/move contents.

I've rewritten `applyLocationBar` basically. It now also just moves a column if only the location has changed, not the url.

I will add a comment with test-cases the next days.

### DIFF
--- a/public/js/icinga/history.js
+++ b/public/js/icinga/history.js
@@ -224,57 +224,60 @@
          * Read the pane url from the current URL and load the corresponding panes into containers to
          * match the current history state.
          *
-         * @param   {Boolean|Null}  onload  Set to true when the main pane should not be updated, defaults to false
+         * @param   {Boolean}  onload  Set to true when the main pane should not be updated, defaults to false
          */
-        applyLocationBar: function (onload) {
-            var icinga = this.icinga,
-                main,
-                parts;
+        applyLocationBar: function (onload = false) {
+            let col2State = this.getCol2State();
 
-            if (typeof onload === 'undefined') {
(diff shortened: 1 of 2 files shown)


nilmerg:
### landing page tests

:heavy_check_mark:  redirect to login, col2 state must be preserved and shown after login
:heavy_check_mark:  already logged in, only col2 state must be loaded, col1 must be left alone

### navigation tests

:heavy_check_mark:  on the dashboard, click on three menu items in a row, navigate back three times and check if the order is the same but reversed and the result is the dashboard again

:heavy_check_mark:  on the dashboard, directly go to the hosts list and choose a host, navigate back two times and notice the following:
  1) the right column closes, the left column is left alone
  2) the left column is replaced

:heavy_check_mark:  Repeat the same test again, but by navigating forward and notice the following:
 1) the left column is replaced
 2) the right column is opened

:heavy_check_mark:  Stay on the host detail and click on the services tab, navigate back and notice:
 1) the left column is replaced

:heavy_check_mark:  Now click on the services tab again and choose one, navigate back two times and notice:
 1) the left column moves to the right, the left column is loaded
 2) the right colum is replaced

:heavy_check_mark:  Navigate forward two times now and notice:
 1) the left column is replaced by the services tab
 2) the right column moves to the left and the service appears in the left column

PR #4766, opened 2022-05-02
sql: Change charset to `utf8mb4`

### LINKED ISSUE #4680: mysql: use of utf8 vs utf8mb4
## Describe the bug

[The internal tables are currently declared as utf8](https://github.com/Icinga/icingaweb2/blob/master/etc/schema/mysql.schema.sql#L13). This is likely not what was intended here. The problem is, that mysql / mariadb utf8 is not actually utf8, but instead utf8mb3, which means that unicode characters which require 4 bytes to encode, are silently dropped.

This happens most likely when using emojis, many of which happen to require a 4 byte encoding in utf8. Eg. 💩 is a popular example.

At best this is unintended behaviour, but at worst this opens up security problems as it is a way to write something to the database and get something else / shorter back.

## To Reproduce
I have not tried to trigger this yet, as this was found during code review / configuration review.

## Expected behavior
Everywhere in mariadb/ mysql utf8mb4 should be used. Plain utf8 is a legacy encoding from the time when that extra byte was considered a waste of space. But full utf8 support is always always what engineers want nowadays. If not, you should switch to an explicit utf8mb3 to document that intention. And also add a lengthy comment that explains why this choice was made.

## Additional context
See https://github.com/Icinga/docker-icingaweb2/pull/64

sukhwinder33445:
fixes #4680

### DIFF
--- a/doc/80-Upgrading.md
+++ b/doc/80-Upgrading.md
@@ -7,6 +7,12 @@ v2.6 to v2.8 requires to follow the instructions for v2.7 too.
 
 * The Vagrant file and all its assets have been removed.
 
+**Database Schema**
+
+* Please apply the `v2.11.0.sql` upgrade script depending on your database vendor.
+  In package installations this file can be found in `/usr/share/doc/icingaweb2/schema/*-upgrades/`
+  (Debian/Ubuntu: `/usr/share/icingaweb2/etc/schema/*-upgrades/`).
+
 ## Upgrading to Icinga Web 2 2.10.x
 
 **General**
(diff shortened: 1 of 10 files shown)


nilmerg (changes requested):
Please update the upgrading documentation by mentioning the upgrade script.

nilmerg:
Oh and check in the code for defaults (DbStore) and similar cases.

nilmerg:
There's more, some `COLLATE utf8...` instructions for example

nilmerg (approved):

PR #4758, opened 2022-04-27
Remove deprecated classes and methods

sukhwinder33445:

PR #4768, opened 2022-05-03
Change FilterMatch to FilterEqual class for single object filter on object name.

### LINKED ISSUE #4682: Wildcard characters in object names should not be processed in detail queries
## Describe the bug
Having an asterisk in an object name causes it to be processed as wildcard.

## To Reproduce
Have a service or host with an asterisk in the name. Click on a link for this object in the UI. The object is not found.

## Expected behavior
The asterisk is not processed as wildcard in this case.

## Additional context
ref/IP/38281

raviks789:
Correspondingly the rendering of the filter for the db query is also modified to identify the instances of FilterEqual and to render accordingly.

This correctly selects the host, service or contact for the detailed view in case the object name contains wild card characters like "\\*".

depends on #4758
fix #4682

### DIFF
--- a/library/Icinga/Data/Db/DbConnection.php
+++ b/library/Icinga/Data/Db/DbConnection.php
@@ -6,6 +6,8 @@ namespace Icinga\Data\Db;
 use DateTime;
 use DateTimeZone;
 use Exception;
+use Icinga\Data\Filter\FilterMatch;
+use Icinga\Data\Filter\FilterMatchNot;
 use Icinga\Data\Inspectable;
 use Icinga\Data\Inspection;
 use PDO;
@@ -552,13 +554,13 @@ class DbConnection implements Selectable, Extensible, Updatable, Reducible, Insp
             throw new ProgrammingError(
                 'Unable to render array expressions with operators other than equal or not equal'
             );
-        } elseif ($sign === '=' && strpos($value, '*') !== false) {
+        } elseif ($filter instanceof FilterMatch && strpos($value, '*') !== false) {
             if ($value === '*') {
                 return $column . ' IS NOT NULL';
             }
 
             return $column . ' LIKE ' . $this->dbAdapter->quote(preg_replace('~\*~', '%', $value));
-        } elseif ($sign === '!=' && strpos($value, '*') !== false) {
+        } elseif ($filter instanceof FilterMatchNot && strpos($value, '*') !== false) {
             if ($value === '*') {
                 return $column . ' IS NULL';
             }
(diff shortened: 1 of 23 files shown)


library/Icinga/Data/Db/DbQuery.php
```diff
@@ -222,7 +223,16 @@ protected function renderFilter($filter, $level = 0)
...
             }
         } else {
-            $str .= $this->whereToSql($filter->getColumn(), $filter->getSign(), $filter->getExpression());
+            if ($filter instanceof FilterEqual) {
+                $str .= sprintf(
+                    '%s %s %s',
+                    $filter->getColumn(),
+                    $filter->getSign(),
+                    $this->escapeForSql($filter->getExpression())
+                );
```
nilmerg, line 232: This bypasses not only the wildcard rendering. There's also handling for arrays and the cases for `IS NULL` and `IS NOT NULL` which are lost now.

Since the method is deprecated anyway and (to be) removed by #4758, patch DbConnection::renderFilterExpression instead. There you can differentiate better as well.

nilmerg (changes requested):

library/Icinga/Data/Db/DbQuery.php
raviks789: Changes made in DbConnection::renderFilterExpression, and the above changes are removed from DbQuery::renderFilter().

modules/monitoring/library/Monitoring/DataView/DataView.php
```diff
@@ -442,7 +443,7 @@ public function applyFilter(Filter $filter)
      */
     public function validateFilterColumns(Filter $filter)
     {
-        if ($filter instanceof FilterMatch) {
+        if ($filter instanceof FilterMatch || $filter instanceof FilterEqual) {
```
nilmerg, line 446: Just noticed, this should only check for `FilterExpression`. That's been wrong all these years already.

library/Icinga/Data/Db/DbConnection.php
```diff
@@ -552,7 +554,7 @@ protected function renderFilterExpression(Filter $filter)
             throw new ProgrammingError(
                 'Unable to render array expressions with operators other than equal or not equal'
             );
-        } elseif ($sign === '=' && strpos($value, '*') !== false) {
+        } elseif ($filter instanceof FilterMatch && strpos($value, '*') !== false) {
```
nilmerg, line 557: There's also the check for `!=` below which requires an update.

nilmerg (changes requested):

nilmerg (approved):

nilmerg (approved):
Had to adjust some more things as the issue affected also all reference queries (such as comment and downtime queries)

nilmerg (approved):

PR #4773, opened 2022-05-05
modal.js: Fix modal hide error

yhabteab:
Fixes a bug that when e.g. closing a color picker by clicking the escape key,
also the `modal.hide()` action is performed, even though there was no real modal
opened before, but it seems to react automatically on the escape key click.

### DIFF
--- a/public/js/icinga/behavior/modal.js
+++ b/public/js/icinga/behavior/modal.js
@@ -171,7 +171,10 @@
         var _this = event.data.self;
 
         if (! event.isDefaultPrevented() && event.key === 'Escape') {
-            _this.hide(_this.$layout.children('#modal'));
+            let $modal = _this.$layout.children('#modal');
+            if ($modal.length) {
+                _this.hide($modal);
+            }
         }
     };
 


nilmerg:
This is rather a problem in the event handler `onKeyDown` which shouldn't attempt to close the modal if there's none.

public/js/icinga/behavior/modal.js
```diff
@@ -170,7 +170,7 @@
     Modal.prototype.onKeyDown = function(event) {
         var _this = event.data.self;
 
-        if (! event.isDefaultPrevented() && event.key === 'Escape') {
+        if (_this.modalOpener && ! event.isDefaultPrevented() && event.key === 'Escape') {
```
nilmerg, line 173: Nah, the modal opener isn't the right indicator. It's searching for the modal below already and passes an empty set to `hide` if there's none. Just check if the set is empty and don't cal hide then.

nilmerg (changes requested):

nilmerg (approved):

PR #4774, opened 2022-05-05
Provide less vars loader hook

yhabteab:


modules/monitoring/run.php
```diff
@@ -6,3 +6,4 @@
 $this->provideHook('ApplicationState');
 $this->provideHook('Health');
 $this->provideHook('X509/Sni');
+$this->provideHook('Themebuilder/LessVarsLoader');
```
nilmerg, line 9: It's no monitoring hook, but one of the main application. Should be registered during bootstrapping I guess.

library/Icinga/Application/Web.php
```diff
@@ -461,6 +463,18 @@ protected function setupFatalErrorHandling()
...
     }
 
+    /**
+     * Register all hooks provided by the main application
+     *
+     * @return $this
+     */
+    protected function registerApplicationHooks()
+    {
+        Hook::register('Themebuilder\\LessVarsLoader', LessVarsLoader::class, LessVarsLoader::class);
```
Al2Klimov, line 473: This method is deprecated since fd5f0e3c7182d942880b22bfdf556043bab48dda.

Al2Klimov (changes requested):

PR #4777, opened 2022-05-10
Fix: Scheme and hostname prepended to col2 url in the address bar

### LINKED ISSUE #4754: Scheme and hostname prepended to col2 url in the address bar
## Describe the bug
If a filterable Icinga DB Web view is active in the right column and the user adds a filter, `http://localhost` is prepended to the url in the address bar. (`http://localhost/icingaweb2/dashboard#!http://localhost/icingaweb2/icingadb/hosts?host.name_ci=*dummy*`)

## To Reproduce

1. Add the hosts view of Icinga DB Web to the dashboard
2. Go to the dashboard settings
3. Open the hosts view by clicking on the link of the dashlet
4. Add a filter in the hosts view

## Expected behavior
There should be no scheme or hostname prepended. Just like how it is if filtering the same view while it is in the left column.

sukhwinder33445:
Direct access (a.href) automatically prepends the scheme and hostname.

fixes #4754

### DIFF
--- a/public/js/icinga/loader.js
+++ b/public/js/icinga/loader.js
@@ -828,9 +828,7 @@
             if (!! contentSeparator) {
                 var locationQuery = req.getResponseHeader('X-Icinga-Location-Query');
                 if (locationQuery !== null) {
-                    var a = this.icinga.utils.getUrlHelper().cloneNode(true);
-                    a.search = locationQuery ? '?' + locationQuery : '';
-
+                    let url = currentUrl.path + (locationQuery ? '?' + locationQuery : '');
                     if (req.autosubmit || autoSubmit) {
                         // Also update a form's action if it doesn't differ from the container's url
                         var $form = $(referrer.forceFocus).closest('form');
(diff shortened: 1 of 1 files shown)


nilmerg (approved):

public/js/icinga/loader.js
```diff
@@ -828,9 +828,7 @@
             if (!! contentSeparator) {
                 var locationQuery = req.getResponseHeader('X-Icinga-Location-Query');
                 if (locationQuery !== null) {
-                    var a = this.icinga.utils.getUrlHelper().cloneNode(true);
-                    a.search = locationQuery ? '?' + locationQuery : '';
-
+                    let url = currentUrl.path + (locationQuery ? '?' + locationQuery : '');
```
nilmerg, line 831: Woops, wait a moment, `currentUrl` is defined conditionally. While the block below can make use of it pretty fine (it's guarded by the same condition), this isn't. Parse the url yourself, if `currentUrl` isn't defined yet.

nilmerg (changes requested):

public/js/icinga/loader.js
sukhwinder33445: ohh yes. But isn't it better if I initialize the `currentUrl` just before (out of) the upper if condition? 🤔

nilmerg: Then you'd parse it without needing it in case.

sukhwinder33445: We need it for the `url` anyway.

nilmerg (approved):

PR #4827, opened 2022-06-15
Fix about page mobile layout

flourish86:


public/css/icinga/about.less
```diff
@@ -97,3 +105,21 @@
...
+
+  .name-value-table {
+    overflow-wrap: break-word;
+
+    td.col-commit-id {
+      white-space: normal;
+      min-width: 13em;
+      width: 100%;
+      // overflow-wrap hack, illustrated here:https://stackoverflow.com/questions/1258416/word-wrap-in-an-html-table
+      max-width: 1px;
```
nilmerg, line 122: You're linking a 12 years old stackoverflow entry? Why not just `work-break:break-all`? (Doesn't require overflow-wrap as well) It's a commit id after all, which has no spaces to begin with where overflap-wrap would be of advantage it seems.

nilmerg (changes requested):

public/css/icinga/about.less
flourish86: ✅

nilmerg (approved):

PR #4831, opened 2022-06-23
Fix that navbar forgets active state after refresh

sukhwinder33445:


public/js/icinga/behavior/navigation.js
```diff
@@ -222,13 +222,20 @@
...
                     this.setActiveAndSelected($(el));
                 }
             });
         } else if ($el.length) {
-            if ($el[0].offsetWidth || $el[0].offsetHeight || $el.closest('.config-menu').length) {
+            let isVisibleMenuItem = $el[0].offsetHeight || $el[0].offsetWidth;
+            if (! isVisibleMenuItem) {
+                let parent = $el[0].closest('.nav-level-2');
+                // check if the outer menu is visible
+                isVisibleMenuItem = parent ? (parent.parentNode.offsetHeight || parent.parentNode.offsetWidth) : null;
```
nilmerg, line 235: With `.nav-level-1 > .nav-item` you get immediate access to the top-level nav item. And since a sub-level nav-item can't exist without a top-level nav-item, you don't have to check the return value. Just access `parent.offsetHeight` and `parent.offsetWidth`.

You could also add `, .config-menu` to the selector. That way you can simplify the check below to just `if (isVisibleMenuItem) {`

nilmerg (approved):

PR #4835, opened 2022-06-27
Fix light mode variable references resolution issue

yhabteab:
Currently, if a variable is pointing to another variable, which in turn is pointing to another variable, e.g. `@foo: green; @foo-bar: @foo; @foo-bar-foo: @foo-bar;` and when the variable `@foo-bar-foo` is used somewhere, the resolution of the variable hierarchy is not performed as defined. The resolved version of this variable would look like `var(--foo-bar-foo, green)`, which is actually wrong. This is because all variables in between will simply omitted and results in the following behavior. Suppose that the variable `@foo: green` is a base variable of Icinga Web 2 and has an appropriate light mode rule defined for it and changes the value to `--foo: white;`. But since the variable hierarchy is not taken into account, the parser will never notice that the value of the base variable has been changed meanwhile and will always render the same value in light and dark mode.

One can solve this by defining a separate light mode rule for the used variable `@foo-bar-foo`, but this would be cumbersome and one would have to look everywhere in which module the base variables from Icinga Web 2 are overwritten. Alternatively, the variable override hierarchy could be resolved exactly as it is defined in the respective less files, and you will always get the desired values in dark and light mode. This PR implements the second solution and the variables are rendered as follows, for example.

```
@color: #ddd;
@mycolor: @color;
@foo-color: @mycolor; 
@foo-bar-color: @foo-color;
@foo-bar-foo-color: @foo-bar-color;
@fade-color: fade(@mycolor, 50);
@fade-color-out: fadeout(@mycolor, 50%);
@fade-color-in: fadein(@fade-color-out, 50%);

.container { 
    color: @foo-bar-foo-color;
    background: @fade-color;
    background: @fade-color-out;
    background: @fade-color-in;
    background-color: fade(@foo-bar-color, 50%);
}
```

### Before

```
.container {
  color: var(--foo-bar-foo-color, #dddddd);
  background: var(--fade-color, rgba(221, 221, 221, 0.5));
  background: var(--fade-color-out, rgba(221, 221, 221, 0.5));
  background: var(--fade-color-in, #dddddd);
  background-color: rgba(221, 221, 221, 0.5);
}
```

### After

```
.container {
  color: var(--foo-bar-foo-color, var(--foo-bar-color, var(--foo-color, var(--mycolor, var(--color, #dddddd)))));
  background: var(--fade-color, rgba(221, 221, 221, 0.5));
  background: var(--fade-color-out, rgba(221, 221, 221, 0.5));
  background: var(--fade-color-in, #dddddd);
  background-color: rgba(221, 221, 221, 0.5);
}
```

fixes https://github.com/Icinga/icingaweb2-module-vspheredb/issues/355

### DIFF
--- a/library/Icinga/Less/Call.php
+++ b/library/Icinga/Less/Call.php
@@ -34,7 +34,7 @@ class Call extends Less_Tree_Call
             }
 
             if ($name) {
-                foreach (array_reverse($env->frames) as $frame) {
+                foreach ($env->frames as $frame) {
                     if (($v = $frame->variable($name))) {
                         // Variables from the frame stack are always of type LESS Tree Rule
                         $vr = $v->value;
--- a/library/Icinga/Less/Call.php
+++ b/library/Icinga/Less/Call.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace Icinga\Less;
+
+use Less_Tree_Call;
+use Less_Tree_Color;
+use Less_Tree_Value;
+use Less_Tree_Variable;
+
+class Call extends Less_Tree_Call
+{
+    public static function fromCall(Less_Tree_Call $call)
+    {
+        return new static($call->name, $call->args, $call->index, $call->currentFileInfo);
+    }
+
+    public function compile($env = null)
+    {
+        if (! $env) {
+            // Not sure how to trigger this, but if there is no $env, there is nothing we can do
+            return parent::compile($env);
+        }
+
+        foreach ($this->args as $arg) {
+            $name = null;
+            if ($arg->value[0] instanceof Less_Tree_Variable) {
+                // This is the case when defining a variable with a callable LESS rules such as fade, fadeout..
+                // Example: `@foo: #fff; @foo-bar: fade(@foo, 10);`
+                $name = $arg->value[0]->name;
+            } elseif ($arg->value[0] instanceof ColorPropOrVariable) {
+                // This is the case when defining a CSS rule using the LESS functions and passing
+                // a variable as an argument to them. Example: `... { color: fade(@foo, 10%); }`
+                $name = $arg->value[0]->getVariable()->name;
+            }
+
+            if ($name) {
+                foreach (array_reverse($env->frames) as $frame) {
+                    if (($v = $frame->variable($name))) {
+                        // Variables from the frame stack are always of type LESS Tree Rule
+                        $vr = $v->value;
+                        if ($vr instanceof Less_Tree_Value) {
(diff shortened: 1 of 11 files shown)


test/php/library/Icinga/Util/LessParserTest.php
```diff
@@ -60,6 +60,38 @@ public function testSimpleVariables()
         );
     }
 
+    public function testSimpleNestedVariables()
```
nilmerg, line 63: `SimpleNested` makes no sense, it's `testNestedVariables` I'd say

test/php/library/Icinga/Util/LessParserTest.php
```diff
@@ -60,6 +60,38 @@ public function testSimpleVariables()
...
+}
+
+CSS
+            ,
+            $this->compileLess(<<<LESS
+@black: black;
+@my-color: @black;
+@my-black-color: @my-color;
+// This triggers invalid argument exception without our custom call class
+@fade-color: fade(@my-color, 50%);
```
nilmerg, line 81: Please create a separate test case for this. And don't just test for the non-error, but check the result as well.

nilmerg (changes requested):

library/Icinga/Less/Call.php
```diff
@@ -0,0 +1,57 @@
...
+                // Example: `@foo: #fff; @foo-bar: fade(@foo, 10);`
+                $name = $arg->value[0]->name;
+            } elseif ($arg->value[0] instanceof ColorPropOrVariable) {
+                // This is the case when defining a CSS rule using the LESS functions and passing
+                // a variable as an argument to them. Example: `... { color: fade(@foo, 10%); }`
+                $name = $arg->value[0]->getVariable()->name;
+            }
+
+            if ($name) {
+                foreach (array_reverse($env->frames) as $frame) {
```
sukhwinder33445, line 27: $env can be null according to the parameter of the method.

lippserd (approved):

library/Icinga/Less/Call.php
```diff
@@ -0,0 +1,73 @@
...
+    public function compile($env = null)
+    {
+        if (! $env) {
+            // Not sure how to trigger this, but if there is no $env, there is nothing we can do
+            return parent::compile($env);
+        }
+
+        foreach ($this->args as $arg) {
+            $name = null;
+            if ($arg->value[0] instanceof Less_Tree_Variable) {
```
nilmerg, line 28: Please add the `! is_array then continue` here

nilmerg (dismissed):

nilmerg (approved):

PR #4848, opened 2022-06-30
Docs: Fix Upgrading

lippserd:
Our icinga.com/docs tooling does not render the previous state properly.


nilmerg:
There are conflicts. You seem to have used an old version for this.

nilmerg (approved):

Issue #4852, opened 2022-07-01
The Event Grid is not visible and throws an SQLSTATE[42883] error


agougo:
## Describe the bug
By clicking on the Menu -> History -> Event Grid an error is shown instead of the event Grid.

SQLSTATE[42883]: Undefined function: 7 ERROR: function from_unixtime(timestamp without time zone) does not exist
LINE 1: SELECT DATE(FROM_UNIXTIME(sth.timestamp)) AS day, SUM(CASE W...
^
HINT: No function matches the given name and argument types. You might need to add explicit type casts., query was: SELECT DATE(FROM_UNIXTIME(sth.timestamp)) AS day, SUM(CASE WHEN sth.state = 2 AND sth.type = 'hard_state' THEN 1 ELSE 0 END) AS cnt_critical_hard FROM (SELECT sh.statehistory_id AS id, ('service') AS object_type, h.display_name AS host_display_name, so.name1 AS host_name, so.name2 AS service_description, s.display_name AS service_display_name, so.name1 AS service_host_name, (CASE WHEN sh.state_type = 1 THEN 'hard_state' ELSE 'soft_state' END) AS type, sh.state_time AS timestamp, sh.object_id, sh.state, (CASE WHEN sh.state_type = 1 THEN sh.output ELSE '[ ' || sh.current_check_attempt || '/' || sh.max_check_attempts || ' ] ' || sh.output END) AS output FROM icinga_statehistory AS sh
INNER JOIN icinga_objects AS so ON so.object_id = sh.object_id AND so.is_active = 1 AND so.objecttype_id = 2
INNER JOIN icinga_services AS s ON s.service_object_id = so.object_id
INNER JOIN icinga_hosts AS h ON h.host_object_id = s.host_object_id WHERE (sh.state_time >= '2022-04-01 12:15:32')) AS sth GROUP BY DATE(FROM_UNIXTIME(sth.timestamp)) ORDER BY DATE(FROM_UNIXTIME(sth.timestamp)) DESC

## To Reproduce
Click on the Menu -> History -> Event Grid

## Expected behavior
The expectation is to show this:

![image](https://user-images.githubusercontent.com/8976509/176893278-2c097e82-72fb-43be-b3a9-1a6cd463c6d6.png)

## Screenshots
Instead I see this:

![Capture](https://user-images.githubusercontent.com/8976509/176893911-9cb5c183-5e67-4243-8866-02c6ceadb88f.PNG)

## Your Environment
[...]

PR #4855, opened 2022-07-01
Fix Event Grid SQLSTATE error.

sebastic:
As reported in #4852, the Event Grid is broken in 2.11.0 because the `timestamp` column is converted from `bigint` to `timestamp`, but it already is. The `from_unixtime` function call is no longer required.


sebastic:
I don't have time to review CLA terms, you can use my changes in this PR as you see fit, I disclaim any copyright.

PR #4864, opened 2022-07-05
Setup wizard: Show module dependencies on requirements page

### LINKED ISSUE #4850: Setup wizard fails to enable modules with dependencies
## Describe the bug
Choosing a module in the setup wizard that depends on another module cannot be enabled by the wizard if the dependency is not also chosen. Though, this is only at the finish of the wizard obvious to the user.

A module might also not be enabled because the order in which modules are enabled by the wizard collides with the implicit dependency check. (e.g. cube before icingadb)

## To Reproduce

1. Visit the setup wizard
2. Choose to enable the director
3. Don't choose the incubator
4. Finish the wizard

## Expected behavior
The requirements page should inform the user about missing/outdated modules.

The wizard should also ensure to enable module dependencies first before the dependent module, or alternatively, disable the implicit dependency checks. (The requirements page should have already ensured that all dependencies are available)

## Your Environment

* Icinga Web 2 version and modules (System - About): 2.11.0

sukhwinder33445:
fixes #4850


modules/setup/application/forms/ModulePage.php
```diff
@@ -105,4 +106,18 @@ public function getModuleWizards()
 
         return $wizards;
     }
+
+    public function getDependecyRequirements()
```
nilmerg, line 110: *Depende**n**cy*

modules/setup/application/views/scripts/form/setup-requirements.phtml
```diff
@@ -13,6 +13,10 @@ if (! $form->getWizard()->getRequirements()->fulfilled()) {
 <h1><?= ucwords($moduleName) . ' ' . $this->translate('Module'); ?></h1>
 <?= $wizard->getRequirements(); ?>
 <?php endforeach ?>
+<?php foreach ($form->getWizard()->getPage('setup_modules')->getDependecyRequirements() as $moduleName => $wizard): ?>
+    <h1><?= ucwords($moduleName) . ' ' . $this->translate('Module Dependencies'); ?></h1>
+    <?= $wizard->getRequirements(); ?>
+<?php endforeach ?>
```
nilmerg, line 19: I wonder why that's necessary, can't you just make `getModuleWizards()` from above to automatically return the dependency wizards?

modules/setup/library/Setup/ModuleDependencyWizard.php
```diff
@@ -0,0 +1,143 @@
...
+use Icinga\Module\Setup\Requirement\ModuleMissingRequirement;
+use Icinga\Module\Setup\Requirement\SetRequirement;
+
+class ModuleDependencyWizard implements SetupWizard
+{
+    protected $module;
+
+    protected $checkedModules;
+
+    public function __construct(Module $module, $checkedModules)
```
nilmerg, line 16: It doesn't hurt to require `$checkedModule` as array type or does it?

modules/setup/library/Setup/ModuleDependencyWizard.php
```diff
@@ -0,0 +1,143 @@
...
+    {
+        $this->module = $module;
+        $this->checkedModules = $checkedModules;
+    }
+
+    public function getSetup()
+    {
+        // TODO: Implement getSetup() method.
```
nilmerg, line 24: If you don't plan to do this, remove such comments.

nilmerg (changes requested):
Plus what we discussed :)

modules/setup/library/Setup/Requirement/WebModuleRequirement.php
```diff
@@ -26,6 +36,6 @@ protected function evaluate()
         }
 
         $this->setStateText(sprintf(mt('setup', '%s version: %s'), $this->getAlias(), $moduleVersion));
-        return version_compare($moduleVersion, $version, $op);
+        return ($version === true || $version === null) ? true : version_compare($moduleVersion, $version, $op);
```
nilmerg, line 39: Ehm, `true || null || compare` ?

modules/setup/application/forms/ModulePage.php
```diff
@@ -100,6 +101,8 @@ public function getModuleWizards()
         foreach ($checked as $name => $module) {
             if ($module->providesSetupWizard()) {
                 $wizards[$name] = $module->getSetupWizard();
+            } elseif (! empty($module->getRequiredModules())) {
```
nilmerg, line 104: There may also be libraries required. If there's one missing a module can't be enabled as well.

Libraries don't need to be chosen, but they have to be installed. If we're now automatically checking module dependencies in the wizard, we should do the same for libraries.

modules/setup/library/Setup/ModuleDependency.php
```diff
@@ -0,0 +1,97 @@
...
+use Icinga\Application\Modules\Module;
+use Icinga\Module\Setup\Requirement\ModuleMissingRequirement;
+use Icinga\Module\Setup\Requirement\SetRequirement;
+use Icinga\Module\Setup\Requirement\WebModuleRequirement;
+
+class ModuleDependency
+{
+    /**
+     * @var Module The given Module
+     */
```
nilmerg, line 14: Either `/** @var ... */` or 

```
/**
 * The given module
 *
 * @var Module
 */
```

please

modules/setup/library/Setup/ModuleDependency.php
```diff
@@ -0,0 +1,97 @@
...
+    protected $module;
+
+    /**
+     * @var array The chosen modules
+     */
+    protected $checkedModules;
+
+    public function __construct(Module $module, array $checkedModules)
```
nilmerg, line 22: No doc here?

modules/setup/library/Setup/ModuleDependency.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        $this->module = $module;
+        $this->checkedModules = $checkedModules;
+    }
+
+    /** Get the module dependency requirements
```
nilmerg, line 28: A line break please?

modules/setup/library/Setup/ModuleDependency.php
```diff
@@ -0,0 +1,97 @@
+<?php
+
+namespace Icinga\Module\Setup;
+
+use Icinga\Application\Modules\Module;
+use Icinga\Module\Setup\Requirement\ModuleMissingRequirement;
+use Icinga\Module\Setup\Requirement\SetRequirement;
+use Icinga\Module\Setup\Requirement\WebModuleRequirement;
+
+class ModuleDependency
```
nilmerg, line 10: How does this work? It's not a page nor a wizard.

nilmerg (changes requested):

modules/setup/library/Setup/ModuleDependency.php
sukhwinder33445: `setup-requirements.phtml` (line 12) gets the wizard and calls `getRequirements()`.
The class `ModuleDependency` contains only this method, i.e. if no `SetupWizard` is provided for the current module (e.g. Cube), this class returns module and lib dependencies.

nilmerg:
Note to myself:

> Module "director" could not be enabled. An error occured:
> WARNING: Module "director" can't be enabled. Module has unmet dependencies
> Module "incubator" has been successfully enabled.

Issue #4862, opened 2022-07-04
Allow export for User Audit


bobapple:
## Is your feature request related to a problem? Please describe.
Currently the `Configuration / Access Control / Audit` menu does not allow the export to PDF.

## Describe the solution you'd like
Allow PDF export for the said menu 🙃 

## Additional context
The feature in question will allow enterprise users of Icinga Web to export and store the permission configuration of single users. Those documents can later be used during security audits to to prove permissions are set correctly.

ref/IP/39011

PR #4875, opened 2022-07-12
Add pdf export for user-audit refs #4862

moreamazingnick:


application/controllers/RoleController.php
```diff
@@ -143,6 +144,7 @@ public function removeAction()
 
     public function auditAction()
     {
+        $this->getTabs()->extend(new OutputFormat(["csv","json"]));
```
nilmerg, line 147: Please add a space after commas in arrays.

application/controllers/RoleController.php
```diff
@@ -244,11 +246,14 @@ public function auditAction()
             ]
         ));
 
-        $this->addControl($header);
```
nilmerg, line 247: It's not an option to remove the header from the controls always. It must not scroll away.

You'd need to do that conditionally instead, though that's not really an option as this can only be known at this stage by accessing the `format` URL parameter, which I don't like.

I'd suggest to set what's inside `$header` also as page title. (`$this->setTitle()`) This way it shows naturally up on the export (on every page) and it even has a benefit for normal views in the browser.

nilmerg (changes requested):

application/controllers/RoleController.php
moreamazingnick: done

application/controllers/RoleController.php
moreamazingnick: sorry didn't see the "scroll away" part, done

application/controllers/RoleController.php
nilmerg: I still see no space there.

```suggestion
        $this->getTabs()->extend(new OutputFormat(["csv", "json"]));
```

application/controllers/RoleController.php
```diff
@@ -244,6 +246,9 @@ public function auditAction()
             ]
         ));
 
+        $this->view->title = $title;
```
nilmerg, line 249: Please use `setTitle` as suggested.

```suggestion
        $this->setTitle($title);
```

application/controllers/RoleController.php
```diff
@@ -244,6 +246,9 @@ public function auditAction()
...
+
+        $header = Html::tag('h2', $title);
         $this->addControl($header);
```
nilmerg, line 252: This can be combined into a single line.

nilmerg (changes requested):

nilmerg (approved):

PR #4882, opened 2022-07-20
loader.js: Don't update history for redirects

### LINKED ISSUE #4879: Redirects are pushed to history
## Describe the bug
Redirects are pushed to history.

## To Reproduce
1) Click on a link in the detail area, that redirects to **another** url than what's already in `#col1`
2) Navigate back once after the redirect finished
3) Note that the previous detail area is now the only column (`#col1`)

## Expected behavior
Navigating back once after such a redirect must show the previous column state. (`#col1` + `#col2`)

## Your Environment

* Icinga Web 2 version and modules (System - About): master

## Additional context
It doesn't happen if the current history state is already the same. So if you click on a link in `#col1` that redirects, the history state being pushed is ignored by the browser. That's also why this issue was never a problem, as there are not many detail areas with links that redirect like this.

nilmerg:
A redirect now also does NOT:

* Remove `.impact` from the container
* Try to set a menu item active..

fixes #4879

### DIFF
--- a/public/js/icinga/loader.js
+++ b/public/js/icinga/loader.js
@@ -926,56 +926,9 @@
                 req = reqOrError;
             }
 
-            // Remove 'impact' class if there was such
-            if (req.$target.hasClass('impact')) {
-                req.$target.removeClass('impact');
-            } else {
-                var $impact = req.$target.find('.impact').first();
-                if ($impact.length) {
-                    $impact.removeClass('impact');
-                }
-            }
-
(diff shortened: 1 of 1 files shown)


sukhwinder33445:
I have tested this PR with bp nodes as services in `DetailViewExtension`. The redirect works fine and is not added to history. I also checked the general redirects like form submission and switching between icingadb and monitoring.

PR #4884, opened 2022-07-21
Enhance collapsible js

nilmerg:
Cherry picked from #4702


nilmerg:
Should be backported to 2.9 and 2.10.

PR #4891, opened 2022-08-04
Unveil hidden downtime in checkable detail view

Al2Klimov:
Case: there are no comments, one downtime and you may not schedule additional ones.

Before: downtime is not shown in checkable detail view.

After: downtime is shown in checkable detail view.

Found while working on ref/IP/42380 .


lippserd (approved):

lippserd (changes requested):
Please adjust the commit message so that it reflects what you've actually fixed 😉

PR #4893, opened 2022-08-09
Docs: Add installation instructions for RHEL 9

htriem:
Simple instructions for properly installing Icingaweb2 on RHEL 9. @nilmerg what's the proper way to go about backporting this to the latest support branch?


Saeid-Abadi:
Since the rhel7 ist not supported, the doc should be modified => Remove rhel7 section.

PR #4895, opened 2022-08-09
Improve Current Incidents Dashboard

jabdoa2:
Only show service incidents for services and hosts which are not in a downtime. Otherwise, it becomes very noisy and people tend to ignore it after some time.


bobapple:
@cla-bot check

PR #4897, opened 2022-08-18
Make "light" the default mode on themes

Mikesch-mp:
Make "light" the default mode on themes

### DIFF
--- a/library/Icinga/Web/StyleSheet.php
+++ b/library/Icinga/Web/StyleSheet.php
@@ -26,7 +26,7 @@ class StyleSheet
      *
      * @var string
      */
-    const DEFAULT_MODE = 'none';
+    const DEFAULT_MODE = 'light';
 
     /**
      * The themes that are compatible with the default theme


nilmerg:
I hoped this discussion is over. If you would call it that. Opening this PR without seeking discussion first leads to this: Closing.

Mikesch-mp:
this discussion? 
https://community.icinga.com/t/theme-mode-in-config-ini/9761/5

> I personally see this request only as a result by not liking our default dark theme. The module providing the old legacy theme does not fully solve this as it’s still dark by default, only its light mode is the old legacy design. So even if a feature request is justified somewhere, it’s on the theme’s issue tracker.

i dont find any other discussions about what the default should be.

nilmerg:
A discussion on Github of course. You weren't even part of that discussion :man_shrugging: 

The `default` mode I was talking about there is what the theme provides *by default*. This is, by design, usually the dark mode, though that's not mandatory, just (highly) recommended. No one hinders you on creating a theme that's only light and one that's only dark, then the admin can set the light theme as default in the general configuration. That's the old way. What's lost then is the automatic change depending on system settings though.

We have no plans to change the *default theme mode* in the framework.

Mikesch-mp:
Actualy i have around ~30 customers that only use the default theme and have to explain nearly every new user how to set it back to light. I totaly can understand the frustration of the admins, and only try to give them way to get the default back.

And because you mention it, how on earth can we have a theme with light only? i didnt found any hint (beside it should be possible) nor any documentation. Because then it would be easy to clone the default Icinga theme and have it with light mode only.

Thanks in advance for any hint about the light only themes

nilmerg:
https://github.com/Icinga/icingaweb2/blob/v2.9.7/public/css/icinga/legacy-theme.less

Issue #4888, opened 2022-07-28
Info about deactivated notifications disappeared


tmatthaeus:
## Describe the bug
If you disable the Notifications item in Icinga Web2 under Settings -> System -> State -> Icinga, then at least since version r2.13.4-1 you don't get any info about notifications being centrally disabled. In version r2.12.1-1
 it definitely still worked. (See screenshots for details).

## To Reproduce

1. Open https://ICINGASERVER/health#!/monitoring/health/info in your webbrowser
2. disable notifications

## Expected behavior
There should be a clear information that the notification is disabled

## Screenshots

New version without disabled notification:
<img width="1413" alt="icinga-r2 13 4-1" src="https://user-images.githubusercontent.com/1618310/181546241-ba6c34b4-ac83-4635-ae43-ce1bcb11e117.png">

Old version with disabled notification and the behaviour which I am missing:
<img width="205" alt="icinga-r2 12 1-1" src="https://user-images.githubusercontent.com/1618310/181546282-ffa78eed-7e18-420c-9c4a-ba288b4ccf64.png">

## Your Environment

* Icinga Web 2 version and modules (System - About):  2.11.1
* Web browser used: Chrome
* Icinga 2 version used: r2.13.4-1
* PHP version used (`php --version`): 7.3.31-1~deb10u1
* Server operating system and version: Debian 10

nilmerg:
Yeah, that indicator got lost with #4651. The easiest and most-fitting solution now would probably be to extend `Icinga\Module\Monitoring\ProvidedHook\Health` to also provide statuses the `Icinga\Module\Monitoring\Web\Navigation\Renderer\BackendAvailabilityNavigationItemRenderer` has.

PR #4898, opened 2022-08-19
Show warning health state badges

raviks789:
The warning state badges should be shown when notifications or active host/service checks are disabled.

#4888


nilmerg (changes requested):
This overrides the critical state

nilmerg:
Also, please drop the `BackendAvailabilityRenderer` thing.

nilmerg (approved):

PR #4901, opened 2022-09-01
Fix LDAP pagination for PHP7.3+

### LINKED ISSUE #4900: LDAP pagination does not work with PHP 7.3+
LDAP pagination does not work with PHP 7.3+ because we do neither [read](https://github.com/Icinga/icingaweb2/blob/v2.11.1/library/Icinga/Protocol/Ldap/LdapConnection.php#L994) the cookie from the response nor [update](https://github.com/Icinga/icingaweb2/blob/v2.11.1/library/Icinga/Protocol/Ldap/LdapConnection.php#L1298) it in the request.

lippserd:
Previously, with PHP 7.3+, the cookie value was neither read from the response nor updated on the subsequent requests.

Please note that this commit changes the parameters of the public function `LdapConnection::ldapSearch()`, but at least in our ecosystem, `LdapConnection` is the only consumer of this function. I choose not to try to keep this backward compatible and just break the API.

With PHP 7.3+, the server-side sorting control is now passed directly to the requests as well.

fixes #4900 
fixes Icinga/icingaweb2-module-director#2574

### DIFF
--- a/library/Icinga/Protocol/Ldap/LdapConnection.php
+++ b/library/Icinga/Protocol/Ldap/LdapConnection.php
@@ -882,6 +882,7 @@ class LdapConnection implements Selectable, Inspectable
             }
         }
 
+        $controls = [];
         $legacyControlHandling = version_compare(PHP_VERSION, '7.3.0') < 0;
 
         $count = 0;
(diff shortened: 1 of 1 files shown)


Thomas-Gelf:
Thank you, has successfully been tested on RHEL 7.9 with PHP 7.3

nilmerg (approved):
Didn't test and only took a quick look at it

PR #4927, opened 2022-10-11
Tests: Introduce base cli bootstrapping to be used for module unit tests

yhabteab:
All our modules can then simply do the following to run custom unittests:

```php
$icingaweb2 = getenv('ICINGAWEB_BASEDIR') ?: '/usr/share/icingaweb2';

require_once $icingaweb2 . '/library/Icinga/Test/Bootstrap.php';
Bootstrap::bootstrap('x509', $basedir, $basedir . '/tests');
```

### DIFF
--- a/library/Icinga/Test/Bootstrap.php
+++ b/library/Icinga/Test/Bootstrap.php
@@ -1,0 +1,30 @@
+<?php
+
+/* Icinga Web 2 | (c) 2022 Icinga GmbH | GPLv2+ */
+
+namespace Icinga\Test;
+
+use Icinga\Application\Cli;
+
+class Bootstrap
+{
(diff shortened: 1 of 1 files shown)


nilmerg:
Using the icingacli does not suffice? (`$ icingacli test php unit --include 'Tests\\\\Icinga\\\\Modules\\\\x509'`)

yhabteab:
This isn't just to run module unit tests with all other Icinga Web 2 tests. With this PR, all modules are able to run their specific unit tests on the command line, but also as GitHub actions. See the above mentioned x509 PR.

nilmerg:
I've made [this](https://github.com/Icinga/icingadb-web/blob/master/.github/workflows/php.yml#L44-L81) possible with #4986 now. This shouldn't be necessary anymore.

PR #4936, opened 2022-10-21
CommentParser: Wrap outputs after `120` chars and not `72`

yhabteab:
Well, what should I say, the PHP code sniffer allows us up to `120` characters line length and when you code/format based on these rules, e.g the description of a cli command, it will mess everything up when you run `icingacli module --help`. So if we are allowed to have up to 120 characters on a line in the code, why not in the output.


nilmerg:
If we're touching this, I'd rather let it wrap at the available column space.

nilmerg (approved):

PR #4946, opened 2022-11-21
CSS: Add `crontab` css rules

yhabteab:


nilmerg:
Is this still required? Then please move it to `compat.less` in ipl-web, otherwise close it.

Issue #4970, opened 2023-01-05
Add/Edit dashlet not possible 


staypekos:
## Describe the bug
The add form ends with failure "The url must not be external." no matter what is used as url. Same as update form even without any change.

The problem is probably in [InternalUrlValidator#21](https://github.com/Icinga/icingaweb2/blob/master/library/Icinga/Web/Form/Validator/InternalUrlValidator.php#L21).

As I remember the relative url (e.g. monitoring/tactical; without initial /) worked to add dashlet (was also possible to finish with full url, but the dashlet does not work after that). AFAIK relative url does not have the scheme so it's not good idea to compare it with app scheme.
Tested scenarios (URL -> result):
monitoring/tactical -> fail on second part (scheme is NULL)
/monitoring/tactical -> fail on first part (relativeUrl is empty string)
https://monitoring/tactical -> "works" - add new dashlet which is not working, because getRelativeUrl() returns only "tactical"
https://icinga.company.my/monitoring/tactical -> still fail on first part (relativeUrl is empty string) (maybe here is the mistake for much longer time?)

## To Reproduce
1. Go to Dashboard -> Add Dashlet
2. Fill required fields
3. Push "Add To Dashboard" button

or
1. Go to Dashboard -> Settings
2. Select dashlet to edit
3. Push "Update Dashlet" button

## Expected behavior
The form accepts some relative or absolute url.

## Screenshots
![image](https://user-images.githubusercontent.com/26731137/210879084-e7c97678-5d55-4be2-9648-dd0f96dd850f.png)

## Your Environment
* Icinga Web 2 version and modules (System - About): 2.11.3
* Web browser used: Google Chrome 108
* Icinga 2 version used (`icinga2 --version`): r2.13.6-1
* PHP version used (`php --version`): 7.4.33
* Server operating system and version: Debian 11

ValeDaRold:
Hi!
We encountered the same problem also with Icingaweb2 version 2.10.4.

tbauriedel:
ref/NC/773400

PR #4972, opened 2023-01-09
Check if url is external in validation function

ValeDaRold:
This PR fixes issue: #4970

### DIFF
--- a/library/Icinga/Web/Form/Validator/InternalUrlValidator.php
+++ b/library/Icinga/Web/Form/Validator/InternalUrlValidator.php
@@ -18,7 +18,7 @@ class InternalUrlValidator extends Zend_Validate_Abstract
     public function isValid($value)
     {
         $url = Url::fromPath($value);
-        if ($url->getRelativeUrl() === '' || $url->getScheme() !== Icinga::app()->getRequest()->getScheme()) {
+        if ($url->getRelativeUrl() === '' || $url->isExternal()) {
             $this->_error('IS_EXTERNAL');
 
             return false;


nilmerg:
This has originally been adjusted to allow sub domains: c2f74d49cbb49c3b201ff528491efc39367c8080

While it fixes the current issue, it breaks the old one. This requires a more sophisticated fix.

ValeDaRold:
Hi @nilmerg!
Thank you for your quick answer.
I got your point and this solution can't work for subdomains. What about updating directly the Url.php class in addition to the change already committed in this PR?
At line 181 we can change the code as follow for accepting subdomains and fixing the problem reported in the issue:
```
        $urlParts = parse_url($url);
        $serverName = str_replace('.', '\.',$request->getServer('SERVER_NAME'));
        if (isset($urlParts['scheme']) && (
            $urlParts['scheme'] !== $request->getScheme()
            || (isset($urlParts['host']) && preg_match("/^([\w\-\.]+\.)?$serverName$/", $urlParts['host']) !== 1)
            || (isset($urlParts['port']) && $urlParts['port'] != $request->getServer('SERVER_PORT')))
        ) {
            $urlObject->setIsExternal();
        }
```

nilmerg:
I don't like changing the meaning of *external* on URL objects. Maybe a new method `getDomainName` which returns `'c.com'` for `'b.c.com'` and `'a.b.c.com'`?

ValeDaRold:
Hi @nilmerg!
As agreed with @lippserd, we wait for your best fix (which should be released within the week).
We are available for help, if necessary. :)

nilmerg:
We came to the conclusion that this change is sufficient. Thank you!

nilmerg (approved):

Issue #3507, opened 2018-07-06
Sort hosts by DNS hierarchy


chucker:
I’m considering migrating our icinga setup such that monitored hosts have full DNS names, such as `mymachine.location.foocorp.com` (e.g., `mail1.hamburg.foocorp.com` and `web1.munich.foocorp.com`). Currently, hosts are instead named in the reverse hierarchy, e.g hamburg-mail1 and munich-web1.

DNS style this feels more ‘correct’ and more scalable to me.

However, it’s not a good fit for icingaweb2, because its sorting doesn’t take this hierarchy into account. It presumably simply sorts alphabetically, like so:

mail1.hamburg.foocorp.com
mail1.munich.foocorp.com
mail2.hamburg.foocorp.com
web1.hamburg.foocorp.com

…when in this case, this kind of sort would be desirable:

mail1.hamburg.foocorp.com
mail2.hamburg.foocorp.com
web1.hamburg.foocorp.com
mail1.munich.foocorp.com

That is, from the top-most domain level, sort each level:

    com
      foocorp
        hamburg
          mail1
        munich

Alternatively, it would be nice to _group_ hosts into sections by hostgroup in the service grid. (And perhaps horizontally, group services by service group!)

cstegm:
This could work: 
Create a custom var for wich contains only the domain. Use the "hidden" feature of Icingaweb to create a list with this Column and than sort by this column. (https://www.icinga.com/docs/icingaweb2/latest/modules/monitoring/doc/11-Add-Columns-List-Views/)

ip1981:
I'd use grouping, e. g. HostGroup "Munich".

chucker:
>This sounds like a very specific use case.

I'm a bit surprised to hear that; I thought using a fully-qualified hostname would be a common convention to refer to hosts in a more complex infrastructure.

>It can only be applied when all host objects follow the same naming schema

True; it does require discipline on the admin's part.

>and it likely needs a different text styling. Users read from left to right and might be confused by a different sorting schema.

Well, arguably, that's not Icinga's fault, but sort of a weird artifact of how DNS was created. But yes, this is tricky.

Maybe the way [some browsers colorize parts of the hostname differently](https://i.kinja-img.com/gawker-media/image/upload/18ixbth10dx0tjpg.jpg) could serve as inspiration.

[...]

PR #4994, opened 2023-02-06
Feature/dns sorting 3507

kroseneg:
This adds an option to the service grid to sort and group the hosts by the DNS hierarchy of the host name.
See #3507 .


kroseneg:
Hi

I've just signed the CLA.
Please re-check my pull request, thank you.


Korbinian

decosilo:
@cla-bot check

bobapple:
@cla-bot check

nilmerg:
Hi,

I don't think we will merge this for the following reasons:

* It has a massive disadvantage because it only has access to a partial result returned by the database. Only a subset of so called *pages* will be ordered, though there may be more data on previous/next pages which would influence the shown result.
* Such a functionality won't be added here anymore. Icinga DB Web already supports sorting by multiple columns. The suggestion made [here](https://github.com/Icinga/icingaweb2/issues/3507#issuecomment-404177815) is already possible by using such a sort parameter: `?sort=host.vars.sub1,host.vars.sub0,host.vars.domain`
  * A widget in the UI allowing to assemble such a rule more easily is already planned

Though, thanks for your contribution and effort anyway. It just would have been a good idea to ask whether such an implementation is worth taking on :wink:

PR #5003, opened 2023-02-14
PHP 8.2 Support

TAINCER:
## depends on
 - #5007


library/Icinga/Less/Visitor.php
```diff
@@ -193,7 +193,7 @@ public function visitColor($c)
     {
         if ($this->definingVariable !== false) {
             // Make sure that all less tree colors do have a proper name
-            $c->name = $this->variableOrigin->name;
+            // $c->name = $this->variableOrigin->name;
```
andypost, line 196: it could be `$c->name = $this->variableOrigin?->name;` as `variableOrigin` could be null

PR #5006, opened 2023-02-15
Upgrade vendor libs for PHP 8.2 Support

TAINCER:
- [Zend](https://github.com/Shardj/zf1-future) change to another fork with version [1.22.0](https://github.com/Shardj/zf1-future/tree/release-1.22.0)

## depends on
 - #4920


nilmerg:
* Did you check if any of our changes are still required in the new zend fork?
* Is it an option to include both now in icinga-php-thirdparty and drop them here?

TAINCER:
> * Did you check if any of our changes are still required in the new zend fork?
I tested it in the Browser. I'll check if the bugs are also fixed in the fork.

> * Is it an option to include both now in icinga-php-thirdparty and drop them here?
Yes, all except Zend. I've opened PRs for that already: 
 - https://github.com/Icinga/icingaweb2/pull/5007
 - https://github.com/Icinga/icinga-php-thirdparty/pull/10

nilmerg:
Why not Zend?

TAINCER:
> Why not Zend?

I don't know if it's going to be merged with the old repository we have for that or create a new one. @lippserd knows more about that.

Issue #4789, opened 2022-05-20
Sort by name of roles does not work properly


dgoetz:
## Describe the bug
Sort by name of roles does not work properly, sorting by other columns works fine.

## To Reproduce
1. Create some roles
2. Look at the default list and it will not be in order by name as the view suggests.
3. Switch set sort to another columns and it will be properly sorted
4. Switch back to sort by name and it will be again not ordered but differ then the initial order

## Expected behavior
Sort by name works, so its easier to find roles without search.

## Screenshots
Not allowed to take a screenshot from this environment because of the LDAP groups.

## Your Environment
Include as many relevant details about the environment you experienced the problem in

* Icinga Web 2 version and modules (System - About): 2.9.6 / 2.10.2
* Web browser used: Edge, Firefox, Chrome
* PHP version used (`php --version`): 7.4.19
* Server operating system and version: RHEL 8

nilmerg:
This is due to roles being stored on disk. The current implementation for this is deprecated though. A near future update will change this to database storage and this issue disappears naturally. So I'm closing here.

Nathaniel-Donahue:
Hi,
in which version will this fix be available? Or is it already implemented?

regards,
Nathaniel

nilmerg:
There will be no *fix*. The storage type change is in-progress and should be finished with v2.12 or v2.13.

nilmerg:
This broke with v2.9.0 (#4349), I'm sorry.

PR #5008, opened 2023-02-22
Fix issue #4789. Sort roles in by name

w1ll-i-code:
Add a column 'name' with the name of the role to the role array so that it is possible to sort the roles by name.

### DIFF
--- a/application/controllers/GroupController.php
+++ b/application/controllers/GroupController.php
@@ -378,7 +378,7 @@ class GroupController extends AuthBackendController
                     'title'         => $this->translate(
                         'Configure roles to permit or restrict users and groups accessing Icinga Web 2'
                     ),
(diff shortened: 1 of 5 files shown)


w1ll-i-code:
Hi, is there something missing for this PR? please let me know so I can fix it in case.

nilmerg:
For a start, a signed CLA ;)

nilmerg:
Any update?

lippserd:
@cla-bot check

lippserd:
CLA was signed just recently for an Icinga 2 PR.

nilmerg:
Your proposed changes are not quite correct. But they pointed me in the correct direction to actually understand the underlying problem. So I took the chance to fix this on my own, as I broke it as well originally. :blush: (#5285)

PR #5014, opened 2023-03-21
Add CSP headers

TAINCER:


library/Icinga/Util/CspHeader.php
```diff
@@ -0,0 +1,57 @@
...
+            return;
+        }
+
+        $sessionNonce = $session->get('csp_nonce');
+        if ($sessionNonce !== null) {
+            $this->cspNonce = $sessionNonce;
+            return;
+        }
+
+        $this->cspNonce = bin2hex(random_bytes(16));
```
lippserd, line 38: Consider using `base64_encode()` instead of `bin2hex()` and tell me why 😉.

library/Icinga/Web/Response.php
```diff
@@ -357,6 +368,19 @@ protected function prepare()
             }
         }
 
+        $useCsp = (bool) Config::app()->get('security', 'use_csp', false);
+
+        if ($useCsp) {
+            $this->setHeader('Strict-Transport-Security', 'max-age=31536000; includeSubdomains; preload', true);
```
lippserd, line 374: Why all the other headers?

library/Icinga/Util/CspHeader.php
```diff
@@ -0,0 +1,57 @@
...
+        $session = Session::getSession();
+
+        if ($this->cspNonce !== null) {
+            return;
+        }
+
+        $sessionNonce = $session->get('csp_nonce');
+        if ($sessionNonce !== null) {
+            $this->cspNonce = $sessionNonce;
+            return;
```
lippserd, line 35: Add newline before return please.

lippserd (changes requested):
In addition to the requested changes please:
* Remove all whitespace changes that are not related to the code you introduce.
* Regenerate the nonce if it's not an AJAX request.

library/Icinga/Util/CspHeader.php
lippserd: > The base64 string would be smaller due to the larger character set (base-64 vs base-16), which would make the HTML output smaller. I don't see an apparent advantage in using base64 here other than that.

Yep, that's the reason why `base64_encode()` should be preferred.

lippserd:
> > * Regenerate the nonce if it's not an AJAX request.
> 
> I did this at first, but when column page contents are being loaded, they are not considered a XML Request `$this->getRequest()->isXmlHttpRequest();` (called from Response.php). That's the main reason why I ended up persisting the nonce throughout the session.

How is that possible? Do you have proof? Everything would fall apart, if they weren't AJAX requests.

library/Icinga/Util/CspHeader.php
```diff
@@ -0,0 +1,57 @@
...
+        if (self::$instance === null) {
+            self::$instance = new self();
+        }
+
+        return self::$instance;
+    }
+
+    private function ensureNonce()
+    {
+        $session = Session::getSession();
```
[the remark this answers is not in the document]
TAINCER: I'm having a problem with the window session.

With the [current implementation](https://github.com/Icinga/icingaweb2/pull/5014/files#diff-4583516c6dde0fd718bae49945f9a2c9787162720125d9801bbd52b0d270520bR26):
```php
$windowSession = Window::getInstance()->getSessionNamespace('csp');
```

The instance is coupled to the `X-Icinga-WindowId` header. The initial request doesn't has this header set, only all following XHR requests. Thus the nonce is different in the initial request header and all following requests. Am I using the API wrong?

library/Icinga/Web/Response.php
TAINCER: Could this be clarified in the docs? Possibly linking to it in the description when setting up CSP? Since this seems to be best practice.

TAINCER:
I just rechecked. Now it always returns true with XHR Requests, my bad. I think I first tested this when some Javascript was still blocked by CSP.

library/Icinga/Util/CspHeader.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+namespace Icinga\Util;
+
+use Icinga\Web\Session;
+use Icinga\Web\Window;
+
+class CspHeader
```
lippserd, line 8: I would rename that class to just `Csp` and would like to use it as follows:

* `Csp::getStyleNonce()` returns the nonce for use in inline styles. We don't have `getScriptNonce()` yet but we might so it makes sense to think about separation right from the beginning.
* `Csp::createNonces()` creates the style nonce. May also create the script nonce in the future.
* `Csp::addHeader(Response)` adds the csp header to the response and also adds the window id header as we're working with `Window` internally.

library/Icinga/Web/Response.php
```diff
@@ -357,6 +357,21 @@ protected function prepare()
...
 
+        if ($this->getOverrideWindowId()) {
+            $this->setHeader('X-Icinga-WindowId', Window::getInstance()->getId(), true);
+        }
+
+        $useCsp = (bool) Config::app()->get('security', 'use_csp', false);
+        $layout = Zend_Layout::getMvcInstance();
+        $hasLayout = $layout !== null && $layout->isEnabled();
+
+        if ($this->getRequest()->isXmlHttpRequest() && $hasLayout && $useCsp) {
```
lippserd, line 368:
```
if (! $this->getRequest()->isXmlHttpRequest()) {
    $layout = Zend_Layout::getMvcInstance();
    // A comment that explains this if sctructure
    if ($layout !== null && $layout->isEnabled()) {
        if (Config::app()->get('security', 'use_csp', false)) {
            Csp::createNonces();
            Csp::addHeader($this);
        }
    }
}

**Review by lippserd — CHANGES_REQUESTED** [review]

File: application/forms/Config/General/ApplicationConfigForm.php [hunk 7]
```diff
@@ -88,6 +88,19 @@ public function createElements(array $formData)
...
         );
 
+        $this->addElement(
+            'checkbox',
+            'security_use_csp',
+            [
+                'label'         => $this->translate('Enable strict content security policy'),
+                'description'   => $this->translate(
+                    'Set whether to to use strict content security policy (CSP).'
+                    . ' This setting prevents cross site scripting.'
```
lippserd, line 98: Unfortunately nothing really prevents `XSS` so I would write: `This setting helps to protect from cross-site scripting (XSS).`

library/Icinga/Util/Csp.php
```diff
@@ -0,0 +1,97 @@
...
+namespace Icinga\Util;
+
+use Icinga\Web\Response;
+use Icinga\Web\Session;
+use Icinga\Web\Window;
+
+class Csp
+{
+    /** @var self */
+    private static $instance = null;
```
lippserd, line 12: This targets all `self` references which in fact prevents that this class is ever subclassed. Please use `static` everywhere instead. `self` as return type declaration is fine though.

library/Icinga/Util/Csp.php
```diff
@@ -0,0 +1,97 @@
...
+
+    private function loadNoncesFromSession()
+    {
+        $instance = self::getInstance();
+        $windowSession = Window::getInstance()->getSessionNamespace('csp');
+
+        $sessionStyleNonce = $windowSession->get('csp_style_nonce');
+        $sessionScriptNonce = $windowSession->get('csp_script_nonce');
+
+        if (! empty($sessionStyleNonce)) {
```
lippserd, line 28: Please remove the empty check. This may hide bugs. `loadNoncesFromSession()` **must** apply the session state. If the nonces are null, they have to be set to null.

library/Icinga/Util/Csp.php
```diff
@@ -0,0 +1,97 @@
...
+            $instance->styleNonce = $sessionStyleNonce;
+        }
+
+        if (! empty($sessionScriptNonce)) {
```
lippserd, line 32: We don't have use cases for this right now, right? I would only introduce this once we really need it. The separate method name for style is just a preparation.

library/Icinga/Util/Csp.php
```diff
@@ -0,0 +1,97 @@
...
+    private static function getInstance(): self
+    {
+        if (self::$instance === null) {
+            self::$instance = new self();
+        }
+
+        return self::$instance;
+    }
+
+    public static function getStyleNonce(): ?string
```
lippserd, line 46: This must throw an exception. If I request a nonce, there has to be a nonce. Else we have bugs in our code.

library/Icinga/Util/Csp.php
```diff
@@ -0,0 +1,97 @@
...
+
+        if (! empty($sessionScriptNonce)) {
+            $instance->scriptNonce = $sessionScriptNonce;
+        }
+    }
+
+    private static function getInstance(): self
+    {
+        if (self::$instance === null) {
+            self::$instance = new self();
```
lippserd, line 40: I would load the nonces from the session here.

lippserd (changes requested):

library/Icinga/Util/Csp.php
TAINCER: It's currently not being used. If the method should remain, it's going to always return `null`. Or should the method also be commented out for now?

lippserd: It's okay to remove everything related to script nonce. We can easily add it, if we need it.

PR #5016, opened 2023-03-23
CSS: Set correct height input[type=“datetime-local”]

flourish86:
The height for `input[type=“datetime-local”]` ist fixed.

Screenshot 2023-03-23 at 16 07 20

But I can’t reproduce the problem for `select`s

Screenshot 2023-03-23 at 16 07 07

### DIFF
--- a/public/css/icinga/forms.less
+++ b/public/css/icinga/forms.less
@@ -228,7 +228,8 @@ form.icinga-form {
   }
 }
 
-form.icinga-form select:not([multiple]) {
+form.icinga-form select:not([multiple]),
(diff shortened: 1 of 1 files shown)


lippserd:
> But I can’t reproduce the problem for `select`s

Maybe with select multiple?

lippserd (commented):
Is `select multiple` affected maybe?

lippserd:
> Is `select multiple` affected maybe?

It's not.

lippserd (approved):

nilmerg (approved):

PR #5031, opened 2023-05-25
Javascript: Add `copy-to-clipboard` behavior

raviks789:
Utilizes `CopyToClipboard.js` from ipl-web

Blocked by Icinga/ipl-web#151

ref: https://github.com/Icinga/icingadb-web/issues/677

### DIFF
--- a/public/js/icinga/behavior/copy-to-clipboard.js
+++ b/public/js/icinga/behavior/copy-to-clipboard.js
@@ -1,0 +1,33 @@
+(function (Icinga) {
+
+    "use strict";
+
+    class CopyToClipboard extends Icinga.EventListener {
+        constructor(icinga)
+        {
+            super(icinga);
+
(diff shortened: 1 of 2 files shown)


public/js/icinga/behavior/copy-to-clipboard.js
```diff
@@ -0,0 +1,33 @@
+(function (Icinga) {
+
+    "use strict";
+
+    class CopyToClipboardEventListener extends Icinga.EventListener {
```
nilmerg, line 5: Please drop the `EventListener` suffix.

nilmerg (changes requested):
Please move this over to Icinga DB Web. We won't publish a new Web release just for this.

raviks789:
These changes will be moved to Icinga DB Web.

PR #5037, opened 2023-06-12
Flyout menu: Render menu `bottom-up` if it exceeds the viewport height

### LINKED ISSUE #4594: Navigation menu hover arrow misplassed
## Screenshots
Screenshot 2021-11-09 at 13 48 15

sukhwinder33445:
fixes #4594

### DIFF
--- a/public/js/icinga/behavior/navigation.js
+++ b/public/js/icinga/behavior/navigation.js
@@ -319,18 +319,22 @@
             $target.siblings().not($target).removeClass('hover');
             $target.addClass('hover');
 
+            var targetHeight = $target.offset().top + $target.outerHeight();
             $flyout.css({
                 bottom: 'auto',
-                top: $target.offset().top + $target.outerHeight()
+                top: targetHeight
             });
 
             var rect = $flyout[0].getBoundingClientRect();
 
-            if (rect.y + rect.height > window.innerHeight) {
+            if (rect.bottom > window.innerHeight) {
+                $flyout.addClass('bottom-up');
                 $flyout.css({
-                    bottom: 0,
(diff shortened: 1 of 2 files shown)


flourish86 (changes requested):
Basically it works.

I found one problem, though, which can best be seen on this screenshot:
in extreme conditions this could fire back and now exceed the top
Screenshot 2024-08-08 at 16 44 59

Since the items outside of the screen aren't accessible anymore in this case, this should be prevented by restricting the height and make the flyout scrollable.

I made a quick mockup for this.

Frame 1

While the screenshot shows the collapsed version, this problem also occurs in the expanded version of the sidebar.

Screenshot 2024-08-08 at 16 55 56

sukhwinder33445:
Hi @flourish86,

I have discussed the problem of overflow with Johannes. It shouldn't happen in a normal screen size, and it's a different issue anyway. So for this PR we will leave it as it is.

public/js/icinga/behavior/navigation.js
```diff
@@ -319,18 +319,22 @@
...
             $flyout.css({
                 bottom: 'auto',
-                top: $target.offset().top + $target.outerHeight()
+                top: targetHeight
             });
 
             var rect = $flyout[0].getBoundingClientRect();
 
-            if (rect.y + rect.height > window.innerHeight) {
+            if (rect.bottom > window.innerHeight && rect.height < targetHeight) {
```
nilmerg, line 330: Why did you change this? Doesn't reliably work anymore for me.

nilmerg (changes requested):

nilmerg (approved):

PR #5046, opened 2023-06-21
Provide migration hook

### LINKED ISSUE #5043: Provide a way to easily perform database migrations

Right now, users have to perform database migration for each individual module manually. Some aren't even aware that database migration is required when they upgrade, but notice it when faced with a nasty stack trace that prevents them from using the module any further. As discussed offline with @lippserd, this should facilitate through a hook implementation. I will describe below how this hook should operate based on my needs, without deep diving into the implementation detail. Nothing is decided yet, if someone has improvements, ideas, then feel free to supplement this.

The purpose of this hook is quite straight forward. The user should be able to perform all required database migrations of the respective modules by just pressing a button. While I personally would like to have to confirm the pending DB migrations of all modules only once and Icinga Web 2 would then apply all of them automatically, though, doesn't necessarily have to be this way. Whether a database migration is required and its importance should be left to that module implementing this hook.

- Icinga Web 2 should be able to scan for pending database migrations of all modules and then inform the user (perhaps by some state badge over the `cog` icon).
- There should also be a menubar as subset of the `cog` icon listing all potential modules that require a DB migration with some brief explanations what the migration is needed for (maybe even providing an option to confirm all the pending migrations right there, or link to the appropriate module).
- Depending on the pending DB migrations (i.e. whether it is optional or required), Icinga Web 2 should force the user to migrate, when accessing the module views despite the rendered popups.

In order for this to work seamlessly, all modules will need to maintain `schema_version` table similar to Icinga 2 and Icinga DB. Besides, Icinga Web 2 should also implement the required widgets so that all modules can use them in a consistent way.

yhabteab:
For testing this hook you can use these PRs from [Reporting](https://github.com/Icinga/icingaweb2-module-reporting/pull/188) and [X509](https://github.com/Icinga/icingaweb2-module-x509/pull/195). But make sure you have created some report schedules with the master branch beforehand.

You can also use this example hook to test PHP code migrations:

```diff
diff --git a/library/Icinga/Application/ApplicationBootstrap.php b/library/Icinga/Application/ApplicationBootstrap.php
index 4e7ddd954..b909aa20e 100644
--- a/library/Icinga/Application/ApplicationBootstrap.php
+++ b/library/Icinga/Application/ApplicationBootstrap.php
@@ -7,6 +7,7 @@ use DirectoryIterator;
 use ErrorException;
 use Exception;
 use Icinga\Application\ProvidedHook\DBMigration;
+use Icinga\Application\ProvidedHook\DummyMigrator;
 use ipl\I18n\GettextTranslator;
 use ipl\I18n\StaticTranslator;
 use LogicException;
@@ -770,6 +771,7 @@ abstract class ApplicationBootstrap
     protected function registerApplicationHooks(): self
     {
         Hook::register('migration', DBMigration::class, DBMigration::class);
+        Hook::register('migration', DummyMigrator::class, DummyMigrator::class);
 
         return $this;
     }
```

```php
<?php

namespace Icinga\Application\ProvidedHook;

use Icinga\Application\Hook\PHPMigrationHook;

class DummyMigrator extends PHPMigrationHook
{
    protected const EXECUTE_AFTER_SCHEMA_VERSION = '2.12.0';

    public function getDescription(): string
    {
        return $this->translate('This is Icinga Web 2');
    }

    public function getName(): string
    {
        return $this->translate('Icinga Web 2');
    }

    public function apply(): bool
    {
        return $this->run(static::VERSION, function () {
            var_dump("Dummy");
        });
    }
}
```

resolves #5043

### DIFF
--- a/library/Icinga/Application/MigrationManager.php
+++ b/library/Icinga/Application/MigrationManager.php
@@ -327,9 +327,16 @@ final class MigrationManager implements Countable
             $conn = $this->elevateDatabaseConnection($conn, $elevateConfig);
         }
 
+        $wizardProperties = (new ReflectionClass(WebWizard::class))
+            ->getDefaultProperties();
+        /** @var array<int, string> $tables */
+        $tables = $wizardProperties['databaseTables'];
+
         $dbTool = $this->createDbTool($conn);
         $dbTool->connectToDb();
-        if (! $dbTool->checkPrivileges($this->getRequiredDatabasePrivileges())) {
+        if (! $dbTool->checkPrivileges($this->getRequiredDatabasePrivileges())
+            && ! $dbTool->checkPrivileges($this->getRequiredDatabasePrivileges(), $tables)
+        ) {
             return false;
         }
 


flourish86 (commented):
### Collapse Toggle Icons
Screenshot 2023-07-25 at 10 47 27

Please invert the arrow direction. We use to display actions not states. So, `caret-down` when collapsed and vice versa.

### List Item Apple Button
Screenshot 2023-07-25 at 10 49 41

### Apply All Button
"Apply All" is too generic. Please provide a clearer label, like "Migrate all modules".
Screenshot 2023-07-25 at 10 50 09

### Visual aligment
Please align on baseline
Screenshot 2023-07-25 at 10 52 47

### Collapse state won't be conserved on auto-refresh

### Collapse button position
While it makes sense to have the collapse button stay in place for interaction reasons (button doesn't "jump"), clarity suffers from a structural aspect, when the migration list is expanded. Please move the button to the bottom of the list.

### Collapse button width
Screenshot 2023-07-25 at 11 07 49

The button doesn't need span the full width. It's doesn't make a difference at first, but when focussed it looks awkward. Probably `display: inline-block` will easily fix this.

yhabteab:
> ### List Item "Migrate" Button
> Screenshot 2023-07-25 at 10 49 41
Please apply the style from the mockups

The mockups uses the style of an `action/button` link but this is a regular submit button and ist styled like another submit buttons of Icinga Web 2. So, why should we change it here?

> While it makes sense to have the collapse button stay in place for interaction reasons (button doesn't "jump"), clarity suffers from a structural aspect, when the migration list is expanded. Please move the button to the bottom of the list.

A `details/summary` tag is standardised to always render the details after the collapse/expand marker. But if you insist on changing this behavior, you can try, cause I'm not and wasn't able to do this.

yhabteab:
For reference here're the mockups!
Pending Migrations

Migrations List

yhabteab:
> > ### List Item "Migrate" Button
> > Screenshot 2023-07-25 at 10 49 41
> > Please apply the style from the mockups
> 
> The mockups uses the style of an `action/button` link but this is a regular submit button and ist styled like another submit buttons of Icinga Web 2. So, why should we change it here?

Done!

flourish86 (dismissed):
Screenshot 2023-07-26 at 16 23 14

Not sure about the reason, but your branch messes with the Icinga DB list items



Screenshot 2023-07-26 at 16 26 47

1) I find it acceptable to label to the list item Button "Apply" since we're talking about a list of (pending) migrations, though it differs from the mockups.

Following this reasoning it doesn't make sense to label the "global" button this way, though. "Apply All Modules" just doesn't make sense, since you aren't list modules.
"Apply All Module Migrations" would be correct. I would rather go for "Apply all", though.

2) Please remove the button background according to the mockups. It's less about default styles than visual clarity in this case and the button shape makes the element too distracting from the headline and the content.

flourish86:
Here are the mockups that feature the discussed changes from our offline meeting.

### The about page

* Use only one button to get to the Migrations list view
* Remove the buttons for each list item

Frame 37

If you want to go a little fancier, this is another variation

Frame 38

### Migrations View

* Move the "Migrations" view from "About" to Menu > System

Frame 44

#### Admins

Frame 7

* Currently not migratable Items should have a disabled button
* Let's Use the icing web2 standard button design we usually have for Lists. It's not consistent, but again, we should take care of this in another issue.
* The disabled buttons should have a tool tip (see mockup)

I don't see hiding the "Apply" buttons as a valid option, we should at least have some hint for non admin users.
* The global "Apply" Button could be disabled have a label in this case, that explains why (Label see mockup, if this is too specific a different version could be "You don’t have the required permissions to apply migrations")

Frame 39

yhabteab:
> I don't see hiding the "Apply" buttons as a valid option, we should at least have some hint for non admin users.

This has been decided in our last discussion. We never render a button if the user doesn't have the required permission to perform an action, neither do we show any hints why!

flourish86:
> > I don't see hiding the "Apply" buttons as a valid option, we should at least have some hint for non admin users.
> 
> This has been decided in our last discussion. We never render a button if the user doesn't have the required permission to perform an action, neither do we show any hints why!
> 
> cc @nilmerg @lippserd

The let's amend the decision. 😉

nilmerg (changes requested):
Firstly, the 2.12 upgrade is mandatory but not documented. Please add this to doc/80-Upgrading.md

nilmerg (changes requested):

library/Icinga/Application/Hook/DBMigrationHook.php
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    public function apply(): bool
+    {
+        if (! $this->preApply()) {
+            return false;
+        }
+
+        foreach ($this->getMigrations() as $migration) {
+            if (! $this->applyOne($migration->getVersion())) {
```
nilmerg, line 57: This also runs `preApply` again. If that's unavoidable, `preApply` should at least cache its result.

nilmerg (changes requested):
Is there a specific reason why most of the heavy work (IO and such) is done during hook/manager construction? I'm not a huge fan of constructors doing such.

library/Icinga/Application/Hook/MigrationHook.php
```diff
@@ -0,0 +1,298 @@
...
+/**
+ * Allows you to automatically perform database migrations and execute arbitrary PHP code for a given version.
+ */
+abstract class MigrationHook implements Countable
+{
+    use HookUtils;
+    use Translation;
+    use Database {
+        getDb as getParentDB;
+    }
```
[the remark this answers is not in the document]
yhabteab: As discussed offline with `getParentDB()`, module hooks can access the sql connection of Icinga Web 2.

library/Icinga/Application/Hook/DBMigrationHook.php
yhabteab: It must be checked whenever you try to run it, since it is associated with other hooks, i.e. if `preApply()` returns false now, it might return true after applying some Icinga Web 2 hooks.

library/Icinga/Application/Hook/DBMigrationHook.php
```diff
@@ -0,0 +1,164 @@
...
+namespace Icinga\Application\Hook;
+
+use DirectoryIterator;
+use Icinga\Application\Hook\Common\DBMigration;
+use Icinga\Application\Icinga;
+use Icinga\Application\Modules\Module;
+use SplFileInfo;
+
+/**
+ * @inheritDoc
```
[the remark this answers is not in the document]
yhabteab: Not using it doesn't solve the problem either.

library/Icinga/Web/Widget/Tab.php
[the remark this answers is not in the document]
yhabteab: cc @flourish86

nilmerg (changes requested):
Instead of `DB` or `Db`, `Database`(`Migration`)?

That's it for now. Still need to test it and look at how it actually works.

library/Icinga/Web/Navigation/ConfigMenu.php
```diff
@@ -222,6 +228,20 @@ protected function createHealthBadge()
...
 
+    protected function createMigrationBadge(): ?StateBadge
+    {
+        $mm = MigrationManager::instance();
+        $count = $mm->count();
+
+        $stateBadge = null;
+        if ($count > 0) {
+            $stateBadge = new StateBadge($count, BadgeNavigationItemRenderer::STATE_PENDING);
+            $stateBadge->addAttributes(['class' => 'disabled']);
```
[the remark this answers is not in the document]
yhabteab: I don't know, just wanted to be consistent with the health badge.

library/Icinga/Application/Hook/Common/DbMigration.php
```diff
@@ -0,0 +1,150 @@
...
+            if (preg_match('/\s*delimiter\s*(\S+)\s*$/im', $statements, $matches)) {
+                $statements = preg_replace('/\s*delimiter\s*(\S+)\s*$/im', '', $statements);
+                $statements = preg_replace('/' . preg_quote($matches[1], '/') . '$/m', ';', $statements);
+            }
+
+            $this->query = $statements;
+        }
+
+        if (empty($this->query)) {
+            throw new LogicException('Nothing to migrate');
```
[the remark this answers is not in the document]
yhabteab: If you have empty upgrade script files, then yes.

library/Icinga/Application/Hook/Common/HookUtils.php
```diff
@@ -0,0 +1,164 @@
...
+     * @param Connection $conn
+     * @param Query $schemaQuery
+     * @param string $version
+     * @param string $reason
+     *
+     * @return $this
+     */
+    protected function insertFailedEntry(Connection $conn, Query $schemaQuery, string $version, string $reason): self
+    {
+        if (! static::getColumnType($conn, $schemaQuery->getModel()->getTableName(), 'success')) {
```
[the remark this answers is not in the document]
yhabteab: Why? I don't want to check for driver specific error codes or messages.

library/Icinga/Web/Widget/ItemList/MigrationList.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * Create a new migration list
+     *
+     * @param Generator<string, MigrationHook> $data
+     *
+     * @param callable|null $formCallback
+     */
+    public function __construct(Generator $data, callable $formCallback = null)
+    {
+        parent::__construct($data->valid() ? $data : []);
```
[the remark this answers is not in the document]
yhabteab: Sometimes it crashes with can't iterate already closed generator error.

schema/pgsql.schema.sql
```diff
@@ -118,13 +118,18 @@ ALTER TABLE ONLY "icingaweb_rememberme"
...
+CREATE TYPE boolenum AS ENUM ('n', 'y');
+
 CREATE TABLE "icingaweb_schema" (
   "id"          serial,
-  "version"     smallint NOT NULL,
-  "timestamp"   int NOT NULL,
+  "version"     varchar(64) NOT NULL,
+  "timestamp"   bigint NOT NULL,
+  "success" boolenum DEFAULT NULL,
+  "reason" text DEFAULT NULL,
```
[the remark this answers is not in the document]
yhabteab: Which looks very strange anyway and I don't even get why the columns have to be quoted.

yhabteab:
> Instead of `DB` or `Db`, `Database`(`Migration`)?

Throughout this PR `Db` is used, and I don't see any benefit in changing this to `Database`, it just stretches the class names seamlessly.

flourish86:
As discussed with @lippserd, here's the updated UI for the Pending Migration List.

## The Basic Migration List

Noticeable Changes:
* Remove badges as they don't convey any information
* Ensure multiline paragraph text is still properly readable. It's max-width 45em in this mockups which correspond to a line-length of 90 – 100 characters.
* Remove bold font-weight for inactive subject elements (bold is reserved for clickable elements in list item)
* Limit the maximum number of initially shown migration files (I chose 3, but maybe 5 would also be fine)
* Migration items should be ordered ascendingly, since the latest is the most important to display and would otherwise be obscured, if there are more than the max number of initially displayed items.

Frame 58

## Error display

* Should there be a file that caused an error, the noticing would be displayed as below the module heading
* The error would be expandable similar to the plugin output in Icinga DB Web object detail
* It should still be display in the list to indicate the order
* The "migrate" button should probably be disabled, since it would fail anyway

Frame 60

## Show all button

* Expanding the list "inline" (or rather in list) wouldn't make sense for a large number of items, since the remaining content would get pushed around and probably get out of reach
* So the "Show All" button should open a new column with all the list items
 
Frame 61

application/controllers/ErrorController.php
```diff
@@ -73,6 +75,18 @@ public function errorAction()
 
                 break;
             default:
+                $mm = MigrationManager::instance();
+                $action = $this->getRequest()->getActionName();
+                $controller = $this->getRequest()->getControllerName();
+                if ($action !== 'hint' && $controller !== 'migrations' && $mm->hasMigrations($moduleName)) {
+                    // For some reason that I don't get, ipl-web's view renderer will not render the HTML content set
```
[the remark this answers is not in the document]
yhabteab: The view renderer from IPL web doesn't render any errors and removing the condition there results in an empty page when child classes from `CompatController` throw an exception.

sukhwinder33445 (dismissed):
`Apply` button shows incorrect count in title.

library/Icinga/Application/Hook/MigrationHook.php
```diff
@@ -0,0 +1,360 @@
...
+            $path = Icinga::app()->getBaseDir();
+        } else {
+            $path = Module::get($this->getModuleName())->getBaseDir();
+        }
+
+        $descriptions = $this->providedDescriptions();
+        $version = $this->getVersion();
+        /** @var SplFileInfo $file */
+        foreach (new DirectoryIterator($path . DIRECTORY_SEPARATOR . $upgradeDir) as $file) {
+            if (preg_match('/^(?:r|v)?([\d+.]+)(?:_([\w+]+))?\.sql$/', $file->getFilename(), $m)) {
```
nilmerg, line 254: Please use `^(?:r|v)?((?:\d+\.){0,2}\d+)(?:_([\w+]+))?\.sql$`, otherwise weird cases like this match: `v2.....4.....5_blub.sql`

nilmerg (changes requested):

nilmerg (changes requested):
~~Oh no, wait. The time frame items still open a modal.~~ (Wrong pr)

library/Icinga/Application/Hook/Common/DbMigration.php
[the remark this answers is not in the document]
yhabteab: No! This is only for error cases (when `file_get_contents()` returns `false` on failure).

library/Icinga/Application/ProvidedHook/DbMigration.php
```diff
@@ -0,0 +1,73 @@
...
+use ipl\Orm\Query;
+use ipl\Stdlib\Filter;
+
+class DbMigration extends MigrationHook
+{
+    use Database;
+
+    public function getName(): string
+    {
+        return $this->translate('Icinga Web 2');
```
nilmerg, line 19: Icinga Web :wink:

nilmerg (changes requested):

library/Icinga/Application/Hook/MigrationHook.php
```diff
@@ -361,8 +362,8 @@ protected function loadLastState(string $version): ?string
         /** @var array<string, string> $states */
         $states = $session->get($this->getModuleName(), []);
         if (! isset($states[$version])) {
-            $schemaQuery = $this->getSchemaQueryFor($version);
-            $schemaQuery->setFilter(Filter::all(Filter::equal('success', 'n')));
+            $schemaQuery = $this->getSchemaQuery()
+                ->setFilter(Filter::all(Filter::equal('success', 'n')));
```
nilmerg, line 366: The version doesn't need to be filtered anymore?

library/Icinga/Application/Hook/Common/DbMigration.php
```diff
@@ -0,0 +1,151 @@
...
+     */
+    public function apply(Connection $conn): self
+    {
+        if (! $this->query) {
+            $statements = @file_get_contents($this->getScriptPath());
+            if ($statements === false) {
+                throw new RuntimeException(sprintf('Cannot load upgrade script %s', $this->getScriptPath()));
+            }
+
+            if (empty($this->query)) {
```
nilmerg, line 133: you haven't tested this, I suppose

library/Icinga/Application/Hook/MigrationHook.php
yhabteab: Previous version of this code:
```php
$schemaQuery = $this->getSchemaQueryFor($version);
$schemaQuery->setFilter(Filter::all(Filter::equal('success', 'n')));
```

The version filter was never used `Query::setFilter()` overrides it.

library/Icinga/Application/Hook/Common/DbMigration.php
yhabteab: > you haven't tested this, I suppose

I don't know what's this supposed to say!
Bildschirmfoto 2023-09-12 um 17 56 06

yhabteab: Bildschirmfoto 2023-09-12 um 17 57 07

yhabteab: 🤦‍♂️

public/css/icinga/pending-migration.less
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  .list-item {
+    .visual.upgrade-failed, span.upgrade-failed, .errors-section > header > i {
+      color: @state-critical;
+    }
+
+    .caption.empty-state {
+      color: @gray-semilight;
+    }
```
nilmerg, line 30: please remove this

library/Icinga/Web/Widget/ItemList/MigrationListItem.php
```diff
@@ -0,0 +1,92 @@
...
+    {
+        $header->addHtml($this->createTitle());
+    }
+
+    protected function assembleCaption(BaseHtmlElement $caption): void
+    {
+        if ($this->item->getDescription()) {
+            $caption->addHtml(Text::create($this->item->getDescription()));
+        } else {
+            $caption->getAttributes()->add('class', 'empty-state');
```
nilmerg, line 70: You *can* use this class here, but please consider adding a `EmptyState` to the caption instead.

library/Icinga/Web/Widget/ItemList/MigrationListItemMinimal.php
```diff
@@ -0,0 +1,151 @@
...
+    protected function assembleCaption(BaseHtmlElement $caption): void
+    {
+        $migrations = $this->item->getMigrations();
+        /** @var DbMigration $migration */
+        $migration = array_shift($migrations);
+        if ($migration->getLastState()) {
+            if ($migration->getDescription()) {
+                $caption->addHtml(Text::create($migration->getDescription()));
+            } else {
+                $caption->getAttributes()->add('class', 'empty-state');
```
nilmerg, line 77: also here, consider adding `EmptyState` to the caption instead.

nilmerg (changes requested):

nilmerg (changes requested):
* please rebase
* please fix the check failures

nilmerg (changes requested):
Please adjust the sql migrations so that the existing schema table rows have a valid timestamp and success value

doc/80-Upgrading.md
```diff
@@ -3,6 +3,17 @@
...
 
+## Upgrading to Icinga Web 2.12.0
+
+**Database Schema**
+
+With the latest Icinga Web versions, you no longer need to manually import sql upgrade scripts. Icinga Web `>= 2.12`
+offers you the possibility to perform such migrations in an easy way. You can find and apply all pending migrations
+of your Icinga Web environment in this view: `icingaweb2/migrations`.
+
+If you do not want to use this new migration automation, please apply the `2.12.0.sql` upgrade script, depending on
```
nilmerg, line 14: *You can still apply the migration manually...* sounds better

doc/80-Upgrading.md
```diff
@@ -3,6 +3,17 @@
 Specific version upgrades are described below. Please note that upgrades are incremental. An upgrade from
 v2.6 to v2.8 requires to follow the instructions for v2.7 too.
 
+## Upgrading to Icinga Web 2.12.0
+
+**Database Schema**
+
+With the latest Icinga Web versions, you no longer need to manually import sql upgrade scripts. Icinga Web `>= 2.12`
+offers you the possibility to perform such migrations in an easy way. You can find and apply all pending migrations
+of your Icinga Web environment in this view: `icingaweb2/migrations`.
```
nilmerg, line 12: *...in the menu at \`System -> Migrations\`*, just like we hint at views everywhere else

nilmerg (changes requested):

nilmerg:
The advanced topics chapter describes how to manually setup the database. There an insufficient grant is suggested which does not allow the automatic migration for users who used that example in the past. The example needs to be adjusted now, so that new users following it won't have the same problem. The users who did follow it already though, need to be aware of the limitation and the migration chapter should mention that and either use a separate example to widen the credentials or the same, as now used in the advanced chapter, if it overrides the existing grant properly.

The wizard also issues `GRANT` statements, and needs to be tested if it sets up credentials sufficient to use this new feature in the future.

library/Icinga/Web/Navigation/ConfigMenu.php
```diff
@@ -222,6 +228,20 @@ protected function createHealthBadge()
         return $stateBadge;
     }
 
+    protected function createMigrationBadge(): ?StateBadge
+    {
+        $mm = MigrationManager::instance();
+        $count = $mm->count();
```
nilmerg, line 234: This may throw. The menu should still be usable in case some extension/hook throws. So please add a try-catch here and just log the error. There should be no badge returned in such a case of course:

```suggestion
        try {
            $mm = MigrationManager::instance();
            $count = $mm->count();
        } catch (Throwable $e) {
            Logger::error('Failed to load pending migrations: %s', $e);
            $count = 0;
        }
```

nilmerg (changes requested):

library/Icinga/Application/ProvidedHook/DbMigration.php
```diff
@@ -0,0 +1,73 @@
...
+                        $this->version = $schema->version;
+
+                        break;
+                    }
+                }
+
+                if (! $this->version) {
+                    $this->version = '2.12.0';
+                }
+            } elseif (static::tableExists($conn, $schemaQuery->getModel()->getTableName())) {
```
nilmerg, line 47: The table check isn't reliable, since the setup wizard didn't define the correct permissions for it till now. I'd suggest to check the collation of e.g. the `icingaweb_user_preference.username` column.

nilmerg (changes requested):

library/Icinga/Application/Hook/Common/DbMigration.php
```diff
@@ -0,0 +1,151 @@
+<?php
+
+/* Icinga Web 2 | (c) 2023 Icinga GmbH | GPLv2+ */
+
+namespace Icinga\Application\Hook\Common;
+
+use ipl\Sql\Connection;
+use RuntimeException;
+
+class DbMigration
```
lippserd, line 10: I would rename that to `DbMigrationScript` or `DbMigrationStep`. Sure, it's in a different namespace, but since you're also implementing the hook as `DbMigration`, I don't think the big differences between those classes are reflected by the namespace.

lippserd (changes requested):

library/Icinga/Application/Hook/Common/DbMigration.php
```diff
@@ -0,0 +1,151 @@
...
+     * @param Connection $conn
+     *
+     * @return $this
+     *
+     * @throws RuntimeException Throws an error in case of any database errors or when there is nothing to migrate
+     */
+    public function apply(Connection $conn): self
+    {
+        if (! $this->query) {
+            $statements = @file_get_contents($this->getScriptPath());
```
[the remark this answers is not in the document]
nilmerg: > but I see it more and more often that side-effect free getters/setters are called inside a class

Yes. Because we do that since ages. The same reason why something else was *fine* just a few days ago. Please let's discuss this before requiring this generally.

library/Icinga/Application/ProvidedHook/DbMigration.php
yhabteab: Collation is MYSQL specific.

schema/mysql-upgrades/2.12.0.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE icingaweb_schema
```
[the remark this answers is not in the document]
yhabteab: I was able to find only documents up to [MYSQL >= `5.7`]( https://dev.mysql.com/doc/refman/5.7/en/drop-table.html).

library/Icinga/Application/Hook/Common/DbMigration.php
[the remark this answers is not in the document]
lippserd: Sure. I was just wondering if this is a personal preference or something we encourage. Because I don't remember any PRs with change requests on this.

If you are referring to the `@return void` topic, it is not because we have been leaving it out for ages, but because it is superfluous.

schema/mysql-upgrades/2.12.0.sql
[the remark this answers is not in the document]
lippserd: I think the exists extensions for `DROP CONSTRAINT/INDEX` and `ADD COLUMN` don't work. See [here](https://downloads.mysql.com/docs/refman-5.6-en.pdf). Idempotent migrations would only be possible with something like [this](https://github.com/Icinga/icingaweb2/blob/feature/optimize-ido-mysql/modules/monitoring/schema/migration-procedures.sql).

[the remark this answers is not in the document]
yhabteab: > Please note that "IF EXISTS" is only available in MySQL 5.1.0 or later.

I know this is not the official documentation, but according to this [website](https://www.w3docs.com/snippets/php/how-can-i-employ-if-exists-for-creating-or-dropping-an-index-in-mysql.html ) the `IF EXISTS` is supported since `5.1`.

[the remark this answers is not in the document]
lippserd: Please test this in a MySQL 5.5 docker container and share the results.

library/Icinga/Application/Hook/MigrationHook.php
```diff
@@ -0,0 +1,387 @@
...
+            $path = Icinga::app()->getBaseDir();
+        } else {
+            $path = Module::get($this->getModuleName())->getBaseDir();
+        }
+
+        $descriptions = $this->providedDescriptions();
+        $version = $this->getVersion();
+        /** @var SplFileInfo $file */
+        foreach (new DirectoryIterator($path . DIRECTORY_SEPARATOR . $upgradeDir) as $file) {
+            if (preg_match('/^(?:r|v)?((?:\d+\.){0,2}\d+)(?:_([\w+]+))?\.sql$/', $file->getFilename(), $m)) {
```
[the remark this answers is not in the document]
nilmerg: `^(?:r|v)([^_]+)(?:_(\w+))?\.sql$` would do the job

library/Icinga/Application/MigrationManager.php
```diff
@@ -0,0 +1,398 @@
...
+     * @param array $elevateConfig
+     */
+    public function fixIcingaWebMysqlGrants(Sql\Connection $db, array $elevateConfig): void
+    {
+        $wizardProperties = (new ReflectionClass(WebWizard::class))
+            ->getDefaultProperties();
+        $privileges = $wizardProperties['databaseUsagePrivileges'];
+        $tables = $wizardProperties['databaseTables'];
+
+        $privileges = $this->getRequiredDatabasePrivileges();
```
nilmerg, line 256: please remove this, or, don't commit it

library/Icinga/Application/ProvidedHook/DbMigration.php
```diff
@@ -0,0 +1,81 @@
...
+use Icinga\Application\Hook\DbMigrationHook;
+use Icinga\Common\Database;
+use Icinga\Model\Schema;
+use ipl\Orm\Query;
+use ipl\Sql\Connection;
+
+class DbMigration extends DbMigrationHook
+{
+    use Database {
+        getDb as public getPublicDb;
```
nilmerg, line 16: My initial goal was to just change the visibility. (Without needing to override the method)

Now the visibility of public doesn't make sense anymore, also the name of the alias does not.

Use `as private getWebDb;` instead. The alias shouldn't be public. It's for internal use after all.

library/Icinga/Application/ProvidedHook/DbMigration.php
nilmerg: Not a problem. Then perform a different check depending on the database vendor. The collation check is only required for MySQL anyway, since there the issue with the grants exist. For postgres this is probably not the case, so there you can check for the schema tables existence.

schema/mysql-upgrades/2.12.0.sql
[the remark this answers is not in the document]
yhabteab: Even MYSQL 5.7 doesn't fully support the `IF EXISTS' extension, so I'm going to drop it from here and x509/reporting modules.

schema/pgsql-upgrades/2.12.0.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TYPE boolenum AS ENUM ('n', 'y');
```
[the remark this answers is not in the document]
yhabteab: No. Neither is the mysql upgrade script!

public/css/icinga/pending-migration.less
```diff
@@ -0,0 +1,173 @@
...
+}
+
+.migration-controls {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+}
+
+.migrations {
+  .icinga-form { // Reset Icinga Form layout styles
```
nilmerg, line 72: If this is necessary, remove the class from the form instead.

nilmerg (changes requested):

library/Icinga/Application/MigrationManager.php
```diff
@@ -0,0 +1,398 @@
...
+            'charset'  => $config->charset
+        ], $db->getAdapter()->getOptions($config)));
+    }
+
+    protected function load(): void
+    {
+        $this->pendingMigrations = [];
+
+        /** @var DbMigrationHook $hook */
+        foreach (Hook::all('migration') as $hook) {
```
nilmerg, line 302: `dbMigration`

schema/mysql-upgrades/2.12.0.sql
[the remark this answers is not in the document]
lippserd: Yes, I would not try to make the scripts idempotent at the moment.

library/Icinga/Application/Hook/MigrationHook.php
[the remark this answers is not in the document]
nilmerg: Minus the `r|`. We don't make a difference yet between them. And once we do, the `r` prefixed files probably won't define a version, which excludes them from any version compare or specificity to a particular release. If we allow the `r` prefix now, we may have to do so in the future without a reason.

library/Icinga/Application/Hook/DbMigrationHook.php
```diff
@@ -0,0 +1,416 @@
...
+                    $migration->getVersion()
+                );
+
+                $this->storeState($migration->getVersion(), null);
+            } catch (Exception $e) {
+                Logger::error(
+                    "Failed to apply %s pending migration version %s \n%s",
+                    $this->getName(),
+                    $migration->getVersion(),
+                    $e->getMessage()
```
nilmerg, line 228: Please pass just `$e`. This way the logged message includes more details about the error and not everything is lost in case debug mode is not enabled.

library/Icinga/Application/Hook/DbMigrationHook.php
```diff
@@ -0,0 +1,416 @@
...
+        }
+
+        foreach ($this->getMigrations() as $migration) {
+            try {
+                $migration->apply($conn);
+
+                $this->version = $migration->getVersion();
+                unset($this->migrations[$migration->getVersion()]);
+
+                Logger::info(
```
nilmerg, line 216: apply may return false, so it doesn't make any sense that a success is logged here in that case. Remove the logs from apply and log a different message in case of an error here instead.

library/Icinga/Application/MigrationManager.php
```diff
@@ -0,0 +1,399 @@
...
+     * @param ?array<string, string> $elevateConfig
+     *
+     * @return bool
+     */
+    public function apply(DbMigrationHook $hook, array $elevateConfig = null): bool
+    {
+        if ($hook->isModule() && $this->hasMigrations(DbMigrationHook::DEFAULT_MODULE)) {
+            Logger::error(
+                'Please apply the Icinga Web pending migration(s) first or apply all the migrations instead'
+            );
```
nilmerg, line 139: Should throw I think, a LogicException for example

library/Icinga/Application/Hook/DbMigrationHook.php
nilmerg, line 216: This should also log who performed the migration and which one. Please use `AuditHook::logActivity` for this.

nilmerg (changes requested):

library/Icinga/Application/Hook/DbMigrationHook.php
```diff
@@ -0,0 +1,416 @@
...
+     */
+    public static function tableExists(Connection $conn, string $table): bool
+    {
+        /** @var false|int $exists */
+        $exists = $conn->prepexec(
+            'SELECT EXISTS(SELECT 1 FROM information_schema.tables WHERE table_name = ?) AS result',
+            $table
+        )->fetchColumn();
+
+        return $exists !== false;
```
nilmerg, line 73: You issue an exists query above. fetchColumn will only return false, if there are no rows returned, which an exists query does never do. So $exists is 0 if the table does not exist.

library/Icinga/Application/Hook/DbMigrationHook.php
yhabteab: `DbMigrationStepp::apply()` doesn't return boolean types, but either executes the statements successfully or fails badly with an exception. Therefore, this particular version of the migration script is successfully appleid at this point and I don't see why you wouldn't log any information here.

yhabteab: You just requested a change above to remove this log message.

nilmerg: aye, I've incorrectly identified the apply being called here as MigrationManager's apply.

library/Icinga/Application/MigrationManager.php
nilmerg: scratch that

nilmerg (approved):
At long last, it's done.

Thoroughly tested several migration paths, with MySQL and PostgreSQL.
Being it a wizard based installation or manual setup.

It works. Seemingly.

lippserd (approved):

PR #5049, opened 2023-06-28
ui.js: Add event `column-moved`

sukhwinder33445:


nilmerg (changes requested):
The event should be fired on the new column and include the id of the source column.

So that `event.target.id === 'col2' && source === 'col1'` is true if `col1` content has moved to `col2`.

nilmerg (approved):

PR #5055, opened 2023-07-06
Fix:Remove detail url when no item selected

### LINKED ISSUE #5053: Detail view url is still present in the browser address bar when the list item is deselected
## Describe the bug
The detail view URL is still present in the browser address bar when the list item is deselected.

## To Reproduce

1. Select a list item (host/service/comment etc)
2. Deselect it by pressing `ctrl + click on item`

sukhwinder33445:
fixes #5053

### DIFF
--- a/public/js/icinga/loader.js
+++ b/public/js/icinga/loader.js
@@ -1061,7 +1061,6 @@
             if (req.addToHistory && ! req.autorefresh) {
                 req.$target.data('icingaRefresh', 0);
                 req.$target.data('icingaUrl', url);
-                icinga.history.pushCurrentState();
             }
 
             if (typeof req.progressTimer !== 'undefined') {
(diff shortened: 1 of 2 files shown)


public/js/icinga/ui.js
```diff
@@ -37,6 +37,8 @@
         });
     };
 
+    var globalCOl = 'empty';
+
```
nilmerg, line 41: ??

nilmerg (changes requested):

public/js/icinga/loader.js
```diff
@@ -1057,9 +1056,6 @@
                         'Request to ' + url + ' has been aborted for ',
                         req.$target
                     );
-
-                    // Aborted requests should not be added to browser history
-                    req.addToHistory = false;
```
nilmerg, line 1062: Manual requests that have been aborted should still be added to history, as discussed yesterday. Though, scripted requests not. So set this to false, but only if `req.scripted`.

nilmerg (changes requested):

nilmerg (approved):

PR #5057, opened 2023-07-07
js: Don't trigger beforerender or rendered for discarded content

### LINKED ISSUE #5056: multiple "1 Item(s) selected"
## Describe the bug
In Firefox a selected item leads to multiple "1 Item(s) selected"-messages at the bottom of the view (1+n times)

## To Reproduce
1. open icingaweb in Firefox 102.12.0esr
2. Menu "Icinga DB" -> Hosts
3. klick in the search-field, choose "Host Name (CI)", add a host and search
4. select the host-event and wait a few moments
5. each automatic refresh adds a "1 Item(s) selected"-message at the bottom of the window

## Expected behavior
Only one message should appear if only one line is selected

## Screenshots
image

## Your Environment
* Icinga DB Web version (System - About): 1.0.2
* Icinga Web 2 version (System - About): 2.11.4
* Web browser: Firefox 102.12.0esr (64-Bit)
* Icinga 2 version (`icinga2 --version`): 2.13.7-1
* Icinga DB version (`icingadb --version`): 1.1.0
* PHP version used (`php --version`): 8.0.29
* Server operating system and version: SLES 15 SP4

## Additional context
Not reproducible in Chrome

nilmerg:
A listener for `rendered` should not need to check if the content really changed. Though, if such a listener is not triggered anymore, `beforerender` listeners also must not be triggered, as they might assume that the content is really being updated and their accompanied `rendered` listener is triggered. (e.g. input-enrichment.js)

This might be breaking change and any `Behavior.renderHook` implementation needs to be checked against it. Potentially also in third party modules. As if such an implementation updates the container on its own, `beforerender` listeners only have access to the updated container after this change, while they had access to the original beforehand.

`rendered` listeners should not be that much affected, as for them the change results in the same behavior as if no update has ever been scheduled for the container.

fixes #5056

### DIFF
--- a/doc/80-Upgrading.md
+++ b/doc/80-Upgrading.md
@@ -15,6 +15,10 @@ v2.6 to v2.8 requires to follow the instructions for v2.7 too.
   developers should not encounter breaking changes in their Less files. The `wikimedia/less.php`
   [changelog](https://github.com/wikimedia/less.php/blob/master/CHANGES.md) covers all changes introduced between
   these versions.
+* Our JavaScript framework does not trigger the events `beforerender` and `rendered` on containers anymore,
+  in case the response has been discarded by a behavior with a `renderHook` implementation. Other postprocessing
+  such as updates according to the request header `X-Icinga-Extra-Updates` or fulfilling redirects is also skipped
+  in this case.
 
 ## Upgrading to Icinga Web 2.13
 
--- a/public/js/icinga/loader.js
+++ b/public/js/icinga/loader.js
@@ -875,7 +875,7 @@
                     }
                 })
             } else {
-                this.renderContentToContainer(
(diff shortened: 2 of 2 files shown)


nilmerg:
The referenced issue has been closed, as it isn't reproducible anymore. That's probably due to a new version of Firefox.esr. The underlying problem with how discarded content is handled, still stands though.

nilmerg:
This is definitely incompatible with `renderHook` implementations that apply updates partially on their own, [as does the graphite module](https://github.com/NETWAYS/icingaweb2-module-grafana/blob/d98f5b01ca687d5973d6fe2b222145b3534870c1/public/js/behavior/iframe.js#L64).

PR #5059, opened 2023-07-13
Allow Enabling Strict Content Security Policy (CSP)

lippserd:
> Content Security Policy ([CSP](https://developer.mozilla.org/en-US/docs/Glossary/CSP)) is an added layer of security that helps to detect and mitigate certain types of attacks, including Cross-Site Scripting ([XSS](https://developer.mozilla.org/en-US/docs/Glossary/Cross-site_scripting)) and data injection attacks. These attacks are used for everything from data theft, to site defacement, to malware distribution.

([MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP))

This PR allows to enable strict CSP by introducing the following features:

* A setting to enable strict CSP in the application configuration.
* A helper class for setting/recreating/accessing nonces for dynamic CSS persisted in window-aware sessions, and for generating the CSP header.
* Logic in our base action controller that is responsible for creating the nonce and adding the CSP header when the page is first loaded or reloaded (document requests): since the first CSP header sent defines the policy, it must also contain the allowed nonces and thus cannot be deferred to Ajax responses.

Strict here means that all content must come from Icinga Web's own origin (also excluding subdomains), plus prohibiting inline JS and allowing inline CSS only if it comes with the accepted nonce.

At the moment, the only thing missing is that Icinga Web has to perform a complete page reload after changing the setting. And of course documentation.

### DIFF
--- a/application/controllers/ConfigController.php
+++ b/application/controllers/ConfigController.php
@@ -96,7 +96,18 @@ class ConfigController extends Controller
         $this->assertPermission('config/general');
         $form = new GeneralConfigForm();
         $form->setIniConfig(Config::app());
-        $form->handleRequest();
+        $form->setOnSuccess(function (GeneralConfigForm $form) {
+            $config = Config::app();
+            $useStrictCsp = (bool) $config->get('security', 'use_strict_csp', false);
+            if ($form->onSuccess() === false) {
+                return false;
+            }
+
+            $appConfigForm = $form->getSubForm('form_config_general_application');
+            if ($appConfigForm && (bool) $appConfigForm->getValue('security_use_strict_csp') !== $useStrictCsp) {
+                $this->getResponse()->setReloadWindow(true);
+            }
+        })->handleRequest();
 
         $this->view->form = $form;
         $this->view->title = $this->translate('General');
--- a/doc/20-Advanced-Topics.md
+++ b/doc/20-Advanced-Topics.md
@@ -4,6 +4,7 @@ This chapter provides details for advanced Icinga Web 2 topics.
 
 * [Global URL parameters](20-Advanced-Topics.md#global-url-parameters)
 * [VirtualHost configuration](20-Advanced-Topics.md#virtualhost-configuration)
+* [Content Security Policy (CSP)](20-Advanced-Topics.md#advanced-topics-csp)
 * [Advanced Authentication Tips](20-Advanced-Topics.md#advanced-topics-authentication-tips)
 * [Source installation](20-Advanced-Topics.md#installing-from-source)
 * [Automated setup](20-Advanced-Topics.md#web-setup-automation)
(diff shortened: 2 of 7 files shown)


library/Icinga/Web/Controller/ActionController.php
```diff
@@ -450,6 +451,22 @@ public function preDispatch()
             $form->handleRequest();
         }
         $this->_helper->layout()->autoRefreshForm = $form;
+
+        if (! $this->isXhr()
+            // Layout is only enabled for initial page loads or page reloads
+            && $this->_helper->layout()->isEnabled()
```
nilmerg, line 457: In which case is it not an XHR request and without layout?

library/Icinga/Web/Controller/ActionController.php
```diff
@@ -450,6 +451,22 @@ public function preDispatch()
...
+            && Config::app()->get('security', 'use_strict_csp', false)
+        ) {
+            $response = $this->getResponse();
+            // Window ID must be explicitly set here for the window-aware CSP nonces to work,
+            // since we only set the window ID upon XHR requests.
+            if (! $response->getHeader('X-Icinga-WindowId')) {
+                $response->setHeader('X-Icinga-WindowId', $this->Window()->getId(), true);
```
nilmerg, line 464: I don't quite understand why this header is needed. It won't be processed anyway, if it's really a page load. Can you explain when/where the header is required?

library/Icinga/Util/Csp.php
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Add Content-Security-Policy header with a nonce for dynamic CSS
+     *
+     * Note that {@see static::createNonces()} must be called beforehand.
+     *
+     * @param Response $response
+     *
+     * @throws RuntimeException If no nonce set for CSS
+     */
+    public static function addHeader(Response $response)
```
yhabteab, line 44:
```suggestion
    public static function addHeader(Response $response): void
```

application/controllers/ConfigController.php
```diff
@@ -96,7 +96,20 @@ public function generalAction()
         $this->assertPermission('config/general');
         $form = new GeneralConfigForm();
         $form->setIniConfig(Config::app());
-        $form->handleRequest();
+        $form->setOnSuccess(function (GeneralConfigForm $form) {
+            $config = Config::app();
+            $useStrictCsp = (bool) $config->get('security', 'use_strict_csp', false);
+            if (! $form->onSuccess()) {
```
nilmerg, line 102: Please check for false explicitly and drop the added return in the form.

application/controllers/ConfigController.php
```diff
@@ -96,7 +96,20 @@ public function generalAction()
...
+                return false;
+            }
+
+            $appConfigForm = $form->getSubForm('form_config_general_application');
+            if ($appConfigForm && (bool) $appConfigForm->getValue('security_use_strict_csp') !== $useStrictCsp) {
+                $this->getResponse()->setReloadWindow(true);
+            }
+
+            return true;
```
nilmerg, line 111: shouldn't be necessary as well

doc/20-Advanced-Topics.md
```diff
@@ -115,6 +116,39 @@ Reload Apache and open the FQDN in your web browser.
 systemctl reload httpd
 ```
 
+## Advanced Security <a id="advanced-topics-security"></a>
+
+### Content Security Policy (CSP) <a id="advanced-topics-security-csp"></a>
```
nilmerg, line 121: Make this a level 2 title please. `Advanced Security` sounds weird.

doc/20-Advanced-Topics.md
```diff
@@ -115,6 +116,39 @@ Reload Apache and open the FQDN in your web browser.
...
+
+Elevate your security standards to an even higher level bei using Icinga Web 2 [Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) policy.
```
nilmerg, line 123: Icinga Web ~~2~~

nilmerg: `Policy (CSP) policy.`?
Better: `Policy (CSP).`

doc/20-Advanced-Topics.md
```diff
@@ -115,6 +116,39 @@ Reload Apache and open the FQDN in your web browser.
...
+Enabling strict CSP can prevent your Icinga Web 2 environment from becoming a potential target of [Cross-Site Scripting (XSS)](https://developer.mozilla.org/en-US/docs/Glossary/Cross-site_scripting)
```
nilmerg, line 124: Icinga Web ~~2~~

nilmerg: You'll get the point, I'm sure ;)

nilmerg (changes requested):

library/Icinga/Web/Controller/ActionController.php
nilmerg: To answer my own question: Static asset (e.g. images) requests and requests with the `_disableLayout` query parameter.

There are other responses without layout. Some of them include JSON or CSV content. Though some still include HTML.

So to clarify what's happening here, I'd suggest to check if the content type is `text/html` which it is only if this is true:

* Not a static asset request
* Not an api request
* Not an export

Though, to check the content type, the CSP header must not be set until the response is ready. i.e. as part of `postDispatch` (which won't be used in some of the cases above anyway)

nilmerg (changes requested):

doc/20-Advanced-Topics.md
```diff
@@ -115,6 +116,37 @@ Reload Apache and open the FQDN in your web browser.
 systemctl reload httpd
 ```
 
+### Content Security Policy (CSP) <a id="advanced-topics-csp"></a>
+
+Elevate your security standards to an even higher level bei using Icinga Web [Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP).
```
nilmerg, line 121: ~~bei~~by

nilmerg: *by enabling the `<csp link>` for Icinga Web.*

nilmerg (changes requested):

library/Icinga/Web/Controller/ActionController.php
nilmerg: Changed it. The header is now registered by the response though. https://github.com/Icinga/icingaweb2/pull/5059/commits/b96894202402950b9c3725e5f01aad4a049f6f51

nilmerg:
Creating the CSP nonce in the controller's constructor seems safe. Once the window is accessed, it's ensured an id is set. The session is also set up at this stage, as the first session access is during authentication, which is part of the bootstrapping.

doc/20-Advanced-Topics.md
```diff
@@ -115,6 +116,37 @@ Reload Apache and open the FQDN in your web browser.
...
+### Content Security Policy (CSP) <a id="advanced-topics-csp"></a>
+
+Elevate your security standards to an even higher level by enabling the [Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) for Icinga Web.
+Enabling strict CSP can prevent your Icinga Web environment from becoming a potential target of [Cross-Site Scripting (XSS)](https://developer.mozilla.org/en-US/docs/Glossary/Cross-site_scripting)
+and data injection attacks. After enabling this feature Icinga Web defines all the required CSP headers. Subsequently,
+only content coming from Icinga Web's own origin is accepted, inline JS is prohibited, and inline CSS is accepted only
+if it contains the nonce set in the response header. 
+
+We decided against enabling this by default as we cannot guarantee that all the modules out there will function correctly.
+Therefore, you have to manually enable this policy explicitly and accept the risks that this might probably break some of
```
nilmerg, line 128: ~~probably~~, *might* is sufficient

doc/20-Advanced-Topics.md
```diff
@@ -115,6 +116,37 @@ Reload Apache and open the FQDN in your web browser.
...
+| Icinga Reporting                  | [v1.0.0](https://github.com/Icinga/icingaweb2-module-reporting/releases/tag/v1.0.0)       |
+| Icinga IDO Reports                | [v0.10.1](https://github.com/Icinga/icingaweb2-module-idoreports/releases/tag/v0.10.1)    |
+| Icinga Cube                       | [v1.3.2](https://github.com/Icinga/icingaweb2-module-cube/releases/tag/v1.3.2)            |
+| Icinga Business Process Modeling  | [v2.5.0](https://github.com/Icinga/icingaweb2-module-businessprocess/releases/tag/v2.5.0) |
+| Icinga Certificate Monitoring     | [v1.3.0](https://github.com/Icinga/icingaweb2-module-x509/releases/tag/v1.3.0)            |
+| Icinga PDF Export                 | [v0.10.2](https://github.com/Icinga/icingaweb2-module-pdfexport/releases/tag/v0.10.2)     |
+| Icinga Web Jira Integration       | [v1.3.2](https://github.com/Icinga/icingaweb2-module-jira/releases/tag/v1.3.2)            |
+| Icinga Web Graphite Integration   | [v1.3.0](https://github.com/Icinga/icingaweb2-module-graphite/releases/tag/v1.3.0)        |
+| Icinga Web GenericTTS Integration | [v2.1.0](https://github.com/Icinga/icingaweb2-module-generctts/releases/tag/v2.1.0)       |
```
nilmerg, line 145: gener**i**ctts, in the url

nilmerg (changes requested):

PR #5066, opened 2023-07-20
js: Always load redirects to their designated target

### LINKED ISSUE #4881: Annoying redirection magic
## What's this about?

There's this [line](https://github.com/Icinga/icingaweb2/blob/62d7ca8e97fc4e7aa1e0bcfb56c9a8bc49c83b76/public/js/icinga/loader.js).

## What does the line do?

It reroutes redirects targeting `#col2` to `#col1` if the target-url path is equal to the url-path currently being used by `#col1`.

## Why is it required?

We added this long ago to automatically close `#col2` and *refresh* `#col1` in case a form has been submitted in `#col2`.

## What does rely on it?

Configuration forms for example where the index of available configuration objects is in `#col1` and the user adds or removes one by interacting with a form in `#col2`.

The form then redirects to the index route, which triggers it.

## What's the problem?

It's just wrong to assume that the same url path should trigger such a behavior. If it would at least be the full url with query, it may be fine, but it just looks at the path. :sob: 

This already led to false positives. Just take a look at [this](https://github.com/Icinga/icingaweb2/commit/18188f71858457e81e32845ab507a09db17af5ff). This was necessary, as in Icinga DB Web there might be the same views in both columns, just with different filters. Interacting there with the view mode switcher triggered this automatism.

Now I face the same issue, though with a landing redirect in `#col2`. This requires a yet another workaround: #4880

## My proposal?

Remove the line. And all workarounds. It's that simple.

## Are there alternatives?

Yes! Such a form can redirect to `__CLOSE__` since Icinga Web v2.9.0 to have the same effect. This will close `#col2` and refresh `#col1` just like an auto-refresh would do.

By looking at the request header `X-Icinga-Container` (contains the column where the response will be loaded in), which is also always transmitted since Icinga Web v2.9.0, it's possible to not tell the client to close `#col1` accidentally. This isn't possible anyway, but the redirect will be ignored.

## What happens if the alternative can't be used?

`#col2` stays open which results in duplicate content in both columns. (Outdated left, Updated right)

## Why not just remove it?

Because it's impossible to identify all cases relying on this. That's where I would like to have some input. Especially by @Thomas-Gelf.

We would need to make use of `__CLOSE__` everywhere instead. And then either require Icinga Web v2.9.0 as minimum or live with the duplicate content.

If not now, when is it feasible? Is there a less intrusive way of getting this removed that I don't see?

nilmerg:
The server applies the redirection magic now, but only for compatibility reasons.

resolves #4881
closes #4880

### DIFF
--- a/library/Icinga/Web/Response.php
+++ b/library/Icinga/Web/Response.php
@@ -315,14 +315,29 @@ class Response extends Zend_Controller_Response_Http
      */
     protected function prepare()
     {
+        $request = $this->getRequest();
         $redirectUrl = $this->getRedirectUrl();
-        if ($this->getRequest()->isXmlHttpRequest()) {
+        if ($request->isXmlHttpRequest()) {
             if ($redirectUrl !== null) {
-                if ($this->getRequest()->isGet() && Icinga::app()->getViewRenderer()->view->compact) {
+                if ($request->isGet() && Icinga::app()->getViewRenderer()->view->compact) {
                     $redirectUrl->getParams()->set('showCompact', true);
                 }
 
-                $this->setHeader('X-Icinga-Redirect', rawurlencode($redirectUrl->getAbsoluteUrl()), true);
+                $encodedRedirectUrl = rawurlencode($redirectUrl->getAbsoluteUrl());
+
+                // TODO: Compatibility only. Remove once v2.14 is out.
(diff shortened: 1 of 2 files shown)


nilmerg:
Also includes changes proposed in #4880

PR #5078, opened 2023-08-09
Test: Register module namespaces correctly

yhabteab:
Currently, if the environment variable is set as follows: `ICINGAWEB_MODULE_DIRS=/icingaweb2/modules:/usr/share/icingaweb2-modules` the base name of these paths is mistakenly assumed to be the module name and registers the namespaces accordingly like: `Icinga\\Modules\\Modules` etc., which is wrong.

### DIFF
--- a/library/Icinga/Application/Test.php
+++ b/library/Icinga/Application/Test.php
@@ -2,6 +2,7 @@
 
 namespace Icinga\Application;
 
+use DirectoryIterator;
 use Icinga\Web\Request;
 use Icinga\Web\Response;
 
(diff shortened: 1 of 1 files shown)


yhabteab:
It is not a bug, I guess. One has to register the modules in this way: `ICINGAWEB_MODULE_DIRS=/icingaweb2/modules/reporting:/usr/share/icingaweb2-modules/x509`

nilmerg:
Yes, there is `ICINGAWEB_MODULES_DIR` for what you initially tried.

PR #5109, opened 2023-09-07
Response: Fix backwards compatibilty for `__CLOSE__`

### LINKED ISSUE #5081: Enhance module en/disable redirects
## Describe the bug
Enabling and disabling a module just closes the `col2` container and doesn't even refresh the `col1` container.

## To Reproduce
Just en/disable a module that doesn't provide `JS` files (if a module provides JS files, the whole window will be reloaded) e.g the `x509` or `reporting` module.

## Expected behavior
It shouldn't close the `col2` container, but refresh `col1` and `col2` instead.

## Additional context
It's related to these fixes https://github.com/Icinga/icingaweb2/pull/5066

Changing this line of code...
https://github.com/Icinga/icingaweb2/blob/9c7d19ea9558db5e82b261c5f81f8862dfbacd05/application/controllers/ConfigController.php#L195

... to this helps, but I'm not sure if this is the right solution.
```php
$this->getResponse()->setHeader('X-Icinga-Extra-Updates', '#col1', true);
$this->redirectNow(Url::fromPath('config/module', ['name' => $moduleName]));
```

nilmerg:
If the redirect target changes, `__CLOSE__` has no effect

fixes #5081

### DIFF
--- a/library/Icinga/Web/Response.php
+++ b/library/Icinga/Web/Response.php
@@ -329,7 +329,13 @@ class Response extends Zend_Controller_Response_Http
 
                 // TODO: Compatibility only. Remove once v2.14 is out.
                 $targetId = $request->getHeader('X-Icinga-Container');
-                if ($request->isPost() && $targetId === 'col2' && $request->getHeader('X-Icinga-Col2-State')) {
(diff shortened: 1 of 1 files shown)


yhabteab (approved):
Works fine!

PR #5110, opened 2023-09-07
Add `never` return type to all redirect methods

yhabteab:


library/Icinga/Web/Controller/ActionController.php
```diff
@@ -413,6 +413,13 @@ public function isXhr()
         return $this->getRequest()->isXmlHttpRequest();
     }
 
+    /**
+     * Redirect xhr requests to the specified url
```
nilmerg, line 417: *Issue a redirect that's performed with XHR by the client*

library/Icinga/Web/Controller/ActionController.php
```diff
@@ -428,6 +435,13 @@ protected function redirectXhr($url)
         $response->redirectAndExit($url);
     }
 
+    /**
+     * Redirect http requests to the specified url
```
nilmerg, line 439: *Issue a redirect that's performed as a native HTTP request by the client*

*This will effectively reload the window*

nilmerg (changes requested):

nilmerg (approved):

PR #5141, opened 2023-10-24
Url migration and restriction

raviks789:
resolves #5140

### DIFF
--- a/modules/migrate/application/clicommands/NavigationCommand.php
+++ b/modules/migrate/application/clicommands/NavigationCommand.php
@@ -31,6 +31,8 @@ class NavigationCommand extends Command
      *  --user=<username>  Migrate monitoring navigation items only for
      *                     the given user. (Default *)
      *
+     *  --override         Override the existing Icinga DB navigation items
+     *
      *  --delete           Remove the legacy files after successfully
      *                     migrated the navigation items.
      */
@@ -120,6 +122,7 @@ class NavigationCommand extends Command
     private function migrateNavigationItems($config, $path, $shared, &$rc)
     {
         $deleteLegacyFiles = $this->params->get('delete');
+        $override = $this->params->get('override');
         $newConfig = $this->readFromIni($path, $rc);
 
         /** @var ConfigObject $configObject */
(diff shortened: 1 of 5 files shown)


modules/migrate/application/clicommands/NavigationCommand.php
```diff
@@ -31,165 +21,18 @@ class NavigationCommand extends Command
...
-
-    /**
-     * Migrate the given config to the given new config path
-     *
-     * @param Config $config
-     * @param string $path
-     * @param int    $rc
-     */
-    private function migrateNavigationItems($config, $path, &$rc)
+    public function indexAction(): void
```
nilmerg, line 29: This should still default to `--user *`, as described in the help. Though, only here, so backwards compatibility is still kept

modules/migrate/application/clicommands/ToicingadbCommand.php
```diff
@@ -0,0 +1,584 @@
...
+                    }
+                }
+
+                if (isset($role[$monitoringPropertyBlackList])) {
+                    if (! isset($role[$icingadbPropertyDenyList]) || $override) {
+                        Logger::info(
+                            'Migrating monitoring blacklisted properties for role "%s" to the Icinga DB Web deny list',
+                            $name
+                        );
+                        $role[$icingadbPropertyDenyList] = $role[$monitoringPropertyBlackList];
```
nilmerg, line 259: these can't just be copied over. See https://icinga.com/docs/icinga-db-web/latest/doc/10-Migration/#monitoringblacklistproperties and https://icinga.com/docs/icinga-web/latest/modules/monitoring/doc/10-Restrict-Custom-Variables/

nilmerg (changes requested):

modules/migrate/application/clicommands/ToicingadbCommand.php
```diff
@@ -0,0 +1,596 @@
...
+    {
+        $override = $this->params->get('override');
+
+        /** @var ?string $groupName */
+        $groupName = $this->params->get('group');
+        /** @var ?string $roleName */
+        $roleName = $this->params->get('role');
+
+        if ($roleName === null && $groupName === null) {
+            $this->fail("One of the parameters 'group' or 'role' must be supplied");
```
nilmerg, line 198: please also fail if both are given, since --role overrules --group, this is better than just ignoring --group

nilmerg:
Moved here https://github.com/Icinga/icingadb-web/pull/929

PR #5170, opened 2024-01-09
list/history: Use intl's default locale to format time and date

nilmerg:


sukhwinder33445 (approved):

nilmerg:
I've just raised the required ipl version as well, as otherwise nothing would *set* a default locale for intl.

PR #5191, opened 2024-02-21
Controller::setupPaginationControl: Avoid passing magic property to parameter `$query`

raviks789:


modules/monitoring/application/controllers/ShowController.php
```diff
@@ -6,6 +6,7 @@
 use Icinga\Data\Filter\FilterEqual;
 use Icinga\Module\Monitoring\Backend\MonitoringBackend;
 use Icinga\Module\Monitoring\Controller;
+use Icinga\Module\Monitoring\DataView\DataView;
```
nilmerg, line 9:
```suggestion
```

nilmerg (changes requested):

nilmerg (approved):

nilmerg:
Not merging now. Doesn't seem to address any phpstan notices at the moment.

Issue #5202, opened 2024-03-26
LdapConnection.php - Undefined array key LDAP_CONTROL_PAGEDRESULTS


lydon42:
There is a Undefined array key "1.2.840.113556.1.4.319" ([LdapConnection.php:1007](https://github.com/Icinga/icingaweb2/blob/cd2daeb2cb8537c633d343a29eb76c54cd2ebbf2/library/Icinga/Protocol/Ldap/LdapConnection.php#L1007C1-L1007C91)) if the LDAP result does not have the paging key. This happened while using LDAP Proxy (with icingaweb2:2.12.1).

Possible solution: add `?? null` at the end. This fixed the problem for me.

nilmerg:
Aye, [a comment above](https://github.com/Icinga/icingaweb2/blob/cd2daeb2cb8537c633d343a29eb76c54cd2ebbf2/library/Icinga/Protocol/Ldap/LdapConnection.php#L904) clearly marks this as optional. The remaining code should account for that.

PR #5206, opened 2024-04-03
LdapConnection: Access array key only if its set

sukhwinder33445:
fixes https://github.com/Icinga/icingaweb2/issues/5202


sukhwinder33445:
Phpstan errors are fixen in https://github.com/Icinga/icingaweb2/pull/5201.

nilmerg (approved):

PR #5207, opened 2024-04-05
php: Split phpstan tests into a separate workflow

nilmerg:


.github/workflows/phpstan.yml
```diff
@@ -0,0 +1,18 @@
+name: PHPStan
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  phpstan:
+    uses: icinga/.github/.github/workflows/phpstan.yml@main
```
nilmerg, line 10: Should we really use `icinga/.github` for this? This is actually only recommended explicitly for starter workflows, but such are not useful here as they're essentially templates, which are used for the workflow wizard here on Github. i.e. Not referenced in the end when commiting one.

So we might want to use a separate repository for such re-usable workflows as this.

This would be especially useful if we decide to reference particular versions. Such workflows may require input variables, such as visible below, which one might argue could change in the future, breaking any checks that are not updated yet.

The downside of this is, we lose the option to just update a single location and have the change be applied where ever it has been referenced.

Soo:

a) icinga/`.github`, with versioning
b) icinga/`.github`, without
c) icinga/`<somethingelse>`, with versioning
d) icinga/`<somethingelse>`, without

I'd prefer d)

PR #5211, opened 2024-04-08
Displaying items in a menu with many items

olimpo88:
When creating many elements in the menu, they do not appear. After verifying the code, it was detected that it was caused because the menu has a fixed height. By eliminating this style, all the menu elements can now be displayed.

Selection_59360

### DIFF
--- a/public/css/icinga/layout.less
+++ b/public/css/icinga/layout.less
@@ -125,7 +125,6 @@
         // Remove gutter for tabs
         margin-left: -1 * @gutter;
         margin-right: -1 * @gutter;
-        height: 2.5em;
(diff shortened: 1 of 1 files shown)


bobapple:
@cla-bot check

nilmerg:
Hi, thanks. Though, there is #3461 for this which [describes a different solution](https://github.com/Icinga/icingaweb2/issues/3461#issuecomment-492153077) than what you've presented here. There's also #3787 which already attempted to solve this, but was postponed due to changes in the layout. Since then, nobody attempted it again. If you want, you can have a go at it.

Issue #5236, opened 2024-07-24
Enabled but not configured module breaks About page in Icinga Web


julianbrost:
## Describe the bug
Some code from the x509 module is run when accessing the global About page in Icinga Web (`/icingaweb2/about`) that throws an exception if the module wasn't configured yet.

## To Reproduce

1. Install Icinga Web 2 as per the usual installation instructions from packages
2. Install the `icinga-x509` package (which automatically enables the module)
3. Access the About page

Instead of the About page, you are greeted by the following exception:
```
Cannot load resource config "x509". Resource does not exist

#0 /usr/share/icingaweb2/modules/x509/library/X509/Common/Database.php(42): Icinga\Data\ResourceFactory::getResourceConfig()
Icinga/icingaweb2-module-x509#1 /usr/share/icingaweb2/modules/x509/library/X509/Common/Database.php(29): Icinga\Module\X509\Common\Database::getDb()
Icinga/icingaweb2-module-x509#2 /usr/share/icingaweb2/modules/x509/library/X509/ProvidedHook/DbMigration.php(88): Icinga\Module\X509\Common\Database::get()
Icinga/icingaweb2-module-x509#3 /usr/share/php/Icinga/Application/Hook/DbMigrationHook.php(293): Icinga\Module\X509\ProvidedHook\DbMigration->getDb()
Icinga/icingaweb2-module-x509#4 /usr/share/php/Icinga/Application/Hook/DbMigrationHook.php(172): Icinga\Application\Hook\DbMigrationHook->load()
Icinga/icingaweb2-module-x509#5 /usr/share/php/Icinga/Application/MigrationManager.php(306): Icinga\Application\Hook\DbMigrationHook->getMigrations()
Icinga/icingaweb2-module-x509#6 /usr/share/php/Icinga/Application/MigrationManager.php(56): Icinga\Application\MigrationManager->load()
Icinga/icingaweb2-module-x509#7 /usr/share/php/Icinga/Application/MigrationManager.php(106): Icinga\Application\MigrationManager->getPendingMigrations()
Icinga/icingaweb2-module-x509#8 /usr/share/php/Icinga/Application/MigrationManager.php(69): Icinga\Application\MigrationManager->count()
Icinga/icingaweb2-module-x509#9 /usr/share/icingaweb2/application/views/scripts/about/index.phtml(99): Icinga\Application\MigrationManager->hasPendingMigrations()
[...]

PR #5239, opened 2024-07-31
About Page: Suppress the MigrationManager exception

sukhwinder33445:
fixes https://github.com/Icinga/icingaweb2/issues/5236


nilmerg:
Previously the migration overview crashed with the same error. While this is bad in the about section, it's *okay* in the migration overview. So please catch any error in the about section only, so that they're still visible in the migration overview.

nilmerg (approved):

PR #5240, opened 2024-07-31
Css: Do not allow `:hover` to affect the `.active` list elements

sukhwinder33445:
The `.active` state should not be overwritten by `:hover`.


nilmerg:
Why?

sukhwinder33445:
If you click on a list-item but do not move the cursor away, the active state is not visible. Icingadb-web shows it.

nilmerg (approved):

PR #5241, opened 2024-08-06
Fix/update font family stacks

flourish86:
Not sure who changed the font stack for the fixed font and why. But it seems like a good opportunity to update our default font-stacks. ✌️

https://css-tricks.com/snippets/css/system-font-stack/


nilmerg:
> Not sure who changed the font stack for the fixed font and why.

Sounds passive-aggressive to me and I want to know which change you're talking about? The lines you touch here were last changed about a decade ago.

flourish86:
> Sounds passive-aggressive to me …

It wasn't meant that way. I stumbled over it and was just wondering, why we aren't using the default OS fonts for the monospaced font stack. But you're right, the change for the fixed family seems to not having been changed.

I was convinced, that when we updated the regular font stack, we also did for the fixed font-family, which was wrong. Hopefully that clarifies it.

Besides that, I'm still convinced that it made sense to do so now.

There's also a relatively new value called `system-ui` which unifies the `apple-system` and `BlinkMacSystemFont` value (see css-tricks article). Theoretically we could also drop `"Segoe UI", "Noto Sans"`, since we don't support IE anymore.

It should also be sufficiently supported in non bleeding edge browsers.
https://caniuse.com/?search=system-ui

nilmerg:
Okay. I was just triggered by this allegation that wasn't backed by any references.

Anyway.

Supporting bleeding-edge font aliases is fine, there are fallbacks for this reason.
And of course, we can drop all IE related stuff. At least, we shouldn't introduce new IE only stuff. (i.e. Noto Sans)

nilmerg:
Quickly tested this and discussed this with @flourish86. Seems fine now, though I noticed that `system-ui` results in a way more bolder appearance of sidebar labels, which is why its `font-weight` needs to be decreased. Though this shows that we should thoroughly test this across OSes to make sure the impact isn't as *bold*. :rofl:

lippserd:
@flourish86 Is this change still up to date?

flourish86:
> @flourish86 Is this change still up to date?

I guess. But the testing still has to be done.

PR #5250, opened 2024-08-26
Add hint for comments in ini files

tbauriedel:
Adds a small hint about comments in .ini files.

### DIFF
--- a/doc/03-Configuration.md
+++ b/doc/03-Configuration.md
@@ -14,6 +14,8 @@ File/Directory                                          | Description
 enabledModules                                          | Symlinks to enabled modules
 modules                                                 | Directory for module specific configuration
 
+Comments in the .ini files follow the (de facto) standard for .ini files in the form of a `;` at the beginning of the line.  
+For multi-line comments or to comment out a entire section, a `;` is required in each line that is to be commented out.
 
 ## General Configuration <a id="configuration-general"></a>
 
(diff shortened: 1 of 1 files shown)


tbauriedel:
ref/NC/827716

doc/03-Configuration.md
```diff
@@ -14,6 +14,8 @@ File/Directory                                          | Description
 enabledModules                                          | Symlinks to enabled modules
 modules                                                 | Directory for module specific configuration
 
+Comments in the .ini files follow the (de facto) standard for .ini files in the form of a `;` at the beginning of the line.  
+For multi-line comments or to comment out a entire section, a `;` is required in each line that is to be commented out.
```
nilmerg, line 18: * *in the .ini files*: The article is redundant since *.ini* narrows it already down
* *standard for …*: There's no need to mention that a file *type* follows the same *type* standard
* *of a `;`*: I usually mention the name of the character and just use its symbol as example
* *the line*: There's no particular line to speak of, so it's *a line*
* The entire second sentence is redundant to me, once it's just **a** line. To clarify that there's no such thing as a multiline comment, an entire sentence is a little over emphasized, don't you think?

```suggestion
Comments in .ini files follow the (de facto) standard in the form of a semicolon (`;`) at the beginning of a (each) line.
```

nilmerg (changes requested):

doc/03-Configuration.md
nilmerg: If you don't mind my suggestion, please just commit it from the UI. I'd squash the result on merge anyway.

nilmerg (approved):
Thank you!

PR #5255, opened 2024-09-04
`JavaScript::optimizeDefine()`: Take multiple `define()` calls in same file into account

sukhwinder33445:
**File changed:**

library/Icinga/Web/JavaScript.php — specifically the optimizeDefine() method.

**The bug it fixes:**

The original code used preg_match(), which only finds the first define() call in a JS file. If a file contained multiple AMD define() calls (e.g. a bundled JS file), only the first one would have its module name and dependencies rewritten to use fully-qualified package paths. All subsequent define() calls were left untouched.

**What the fix does:**

The core change is switching from preg_match() to preg_match_all(..., PREG_SET_ORDER), which captures all define() matches in the file. The processing logic is then wrapped in a foreach ($matches as $match) loop so each define() call gets the same name/dependency optimization treatment. At the end of each iteration, str_replace() now patches the JS string in-place and the loop continues, with a final return $js after all matches are processed (instead of the previous early return str_replace(...)).

The logic inside the loop is otherwise identical to the original — asset name resolution, dependency path rewriting, JSON encoding, and the define.amd guard check all remain the same.

### DIFF
--- a/library/Icinga/Web/JavaScript.php
+++ b/library/Icinga/Web/JavaScript.php
@@ -213,57 +213,67 @@ class JavaScript
      */
     public static function optimizeDefine($js, $filePath, $basePath, $packageName)
     {
-        if (! preg_match(self::DEFINE_RE, $js, $match) || strpos($js, 'define.amd') !== false) {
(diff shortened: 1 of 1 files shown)


nilmerg (dismissed):

lippserd (dismissed):
This PR lacks a description.

nilmerg (approved):

Issue #5271, opened 2024-10-04
PluginOutputRenderer gets called twice in monitoring module


moreamazingnick:
## Describe the bug
The render is called seperately for output and long_output wich causes the PluginOutputRenderHook to be called twice see here:

https://github.com/Icinga/icingaweb2/blob/main/modules/monitoring/application/views/scripts/show/components/output.phtml

## To Reproduce
Provide a link to a live example, or an unambiguous set of steps to reproduce this issue. Include configuration, logs, etc. to reproduce, if relevant.

1. Create a PluginOutputHook implementation
2. implement function render($command, $output, $detail)

## Expected behavior
* Use some parameter to let the Pluginoutput hook decide if it is about the output or the long_output.

or change the output.html into something like this:
```php
<?= $this->pluginOutput($object->output."\n".$object->long_output, false, $object->check_command) ?>
```
this would also get rid of the space between output and long_output.

## Screenshots
![Screenshot 2024-10-04 155921](https://github.com/user-attachments/assets/e7695ccf-2cfe-4ca6-b188-cf21fff249f2)

## Your Environment
Include as many relevant details about the environment you experienced the problem in

* Icinga Web 2 version and modules (System - About): 2.12.2

Im happy to provide a pull request.

Best Regards
Nicolas

nilmerg:
> or change the output.html into something like this:

:+1: (But with spaces around the concatenation operator (`.`))

Feel free to open a PR, thanks :D

PR #5272, opened 2024-10-07
Fix PluginOutputHookImplementation gets called twice #5271

moreamazingnick:
Concatenate Output.

This might break workarounds used implemented in a PluginOutputHookImplementation but I haven't found any out in the wild.

This removes the annoying space between the first line and the other lines in the monitoring module plugin output

Alternatives :
Introduce a monitoring setting that disables this fix for a fast switch back in case this breaks something

Best Regards
Nicolas

### DIFF
--- a/doc/80-Upgrading.md
+++ b/doc/80-Upgrading.md
@@ -2,11 +2,13 @@
 
 Specific version upgrades are described below. Please note that upgrades are incremental. An upgrade from
 v2.6 to v2.8 requires to follow the instructions for v2.7 too.
+
 ## Upgrading to Icinga Web 2.12.2
 
 **Framework changes affecting third-party code**
 
 * `Icinga\Module\Monitoring\Hook\PluginOutputHook` When rendering the Icinga check output, the output and long_output fields are now concatenated with a newline (\n) before any post-processing occurs, such as through a PluginOutputHook
+
 ## Upgrading to Icinga Web 2.12.0
 
 **Database Schema**
--- a/doc/80-Upgrading.md
+++ b/doc/80-Upgrading.md
@@ -2,7 +2,11 @@
 
 Specific version upgrades are described below. Please note that upgrades are incremental. An upgrade from
 v2.6 to v2.8 requires to follow the instructions for v2.7 too.
+## Upgrading to Icinga Web 2.12.2
 
+**Framework changes affecting third-party code**
+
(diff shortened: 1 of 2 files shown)


modules/monitoring/application/views/scripts/show/components/output.phtml
```diff
@@ -1,5 +1,4 @@
 <h2><?= $this->translate('Plugin Output') ?></h2>
 <div id="check-output-<?= $this->escape(str_replace(' ', '-', $object->check_command)) ?>" class="collapsible" data-visible-height="100">
-    <?= $this->pluginOutput($object->output, false, $object->check_command) ?>
-    <?= $this->pluginOutput($object->long_output, false, $object->check_command) ?>
+    <?= $this->pluginOutput($object->output. "\n" . $object->long_output, false, $object->check_command) ?>
```
nilmerg, line 3: a space **before** and **after** the `.` operator

moreamazingnick:
@nilmerg do you mean I should edit this as well?
https://github.com/Icinga/icingaweb2/blob/main/CHANGELOG.md

nilmerg:
No, the [upgrade](https://icinga.com/docs/icinga-web/latest/doc/80-Upgrading/) documentation. You can refer to the 2.11.x framework changes to get an idea how.

nilmerg:
Yep. But instead of `(monitoring/ido)`, mention the absolute class path of the hook. (With code highlighting using single backticks)

doc/80-Upgrading.md
```diff
@@ -2,7 +2,11 @@
 
 Specific version upgrades are described below. Please note that upgrades are incremental. An upgrade from
 v2.6 to v2.8 requires to follow the instructions for v2.7 too.
+## Upgrading to Icinga Web 2.12.2
 
+**Framework changes affecting third-party code**
+
+* `Icinga\Module\Monitoring\Hook\PluginOutputHook` When rendering the Icinga check output, the output and long_output fields are now concatenated with a newline (\n) before any post-processing occurs, such as through a PluginOutputHook
 ## Upgrading to Icinga Web 2.12.0
```
nilmerg, line 10: We try to keep markdown readable in raw format, so please add an empty line before any heading :)

nilmerg (changes requested):

doc/80-Upgrading.md
moreamazingnick: done

nilmerg (approved):
LGTM!

PR #5281, opened 2024-10-29
Corrected kiosk mode config

log1-c:
Only got the documented steps to work if the [autologin] part was added, so I suppose it is missing from this documentation

### DIFF
--- a/doc/20-Advanced-Topics.md
+++ b/doc/20-Advanced-Topics.md
@@ -425,6 +425,13 @@ If you want to show a specific Dashboard you can enforce this onto the kiosk use
 
 Replace Remote_Addr with the IP where the kiosk user is accessing the Web to restrict further usage from other IPs.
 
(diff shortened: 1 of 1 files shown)


nilmerg (changes requested):
Please drop the example and link to `05-Authentication.md#authentication-configuration-external-authentication` instead. I also don't see a need for the configuration path here.

And please capitalize the first word of a new sentence. ;)

log1-c:
@nilmerg 
Added the link (though I personally don't see the benefit of "jumping" to another docs page for just two lines of code).
Full path was used because it is used everywhere else in this docs page as well.

nilmerg (approved):
Well, ok, but the link points to the main location where external auth is described. In case the config ever changes, it's just a single place to update.

nilmerg:
Eh, and thank you :D

PR #5282, opened 2024-10-31
fix(csv): fix #5279 always force subject to be a string in str_replace

### LINKED ISSUE #5279: PHP 8.1.27 - Deprecated:  str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /usr/share/php/Icinga/File/Csv.php on line 40
## Describe the bug
I want to print host-objects with some custom-vars.
Each line that contains a null value in the custom-var, raises a php error.

Example command:
`icingacli monitoring list hosts --format csv --columns "host,_host_costcenter_name"`

Example Output:
image

## Your Environment

* Icinga 2 version used (`icinga2 --version`): r2.14.2-1
* PHP version used (`php --version`): 8.1.27
* Server operating system and version: RHEL 9
* Icingaweb-Modules / Packages: image

Yoda-BZH:
When performing a str_replace, subject should always be a string. If passing a non-existent column or empty column, subject may be null

This ensure str_replace is called when subject is actually a string (not NULL)

### DIFF
--- a/library/Icinga/File/Csv.php
+++ b/library/Icinga/File/Csv.php
@@ -37,7 +37,7 @@ class Csv
             }
             $out = array();
             foreach ($row as & $val) {
-                $out[] = '"' . str_replace('"', '""', $val) . '"';
+                $out[] = '"' . ($val ? str_replace('"', '""', $val) : '') . '"';
             }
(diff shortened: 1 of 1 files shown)


nilmerg (approved):
Thank you!

PR #5311, opened 2025-01-08
Log user roles

jrauh01:
Log the roles assigned to a user during login in the debug log.


lippserd:
@jrauh01 Can you please rebase and only implement a minimal change required for logging the groups? Diff is rather big.

jrauh01:
I have rebased and dropped the commits that just refactored stuff.

PR #5319, opened 2025-02-03
Prevent unintentional closing of a modal

### LINKED ISSUE #5307: Prevent unintentional closing of a modal
A modal has currently the following ways to be closed:

1) Click on the X in the top right
2) Click anywhere outside the modal
3) Push the Escape key
4) Successfully submit a form
5) Click on the cancel button (#5305)

Cases 2 and 3 can easily be triggered, without any intention of the user to close the modal.

To prevent this, once a form has been adjusted, both cases shouldn't allow to close the modal anymore.

An alternative would be to ask for confirmation. Though, I'd like to avoid this at all cost.

sukhwinder33445:
resolves #5307

### DIFF
--- a/public/js/icinga/behavior/modal.js
+++ b/public/js/icinga/behavior/modal.js
@@ -18,6 +18,7 @@
         this.icinga = icinga;
         this.$layout = $('#layout');
         this.$ghost = $('#modal-ghost');
+        this.hasChanges = false;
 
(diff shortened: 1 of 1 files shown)


nilmerg:
I had to rework this from scratch as most of the changes here didn't fully track the user's changes nor did respect normal user interaction. (i.e. the blur upon pushing Escape to trigger the change event)

Opened #5501 as a result.

PR #5337, opened 2025-03-20
Allow editing of the CSP trusted image sources.

### LINKED ISSUE #5333: Allow customization of the CSP
The current Content Security Policy is hardcoded, but not all deployments have the same requirements. It would be helpful to allow customizing the CSP, for example, from an ini file.

w1ll-i-code:
Hi, as discussed with @lippserd, we improved the CSP header and added the ability to whitelist certain trusted domains for the image sources. This closes #5333.


lippserd:
> Hi, as discussed with @lippserd, we improved the CSP header and added the ability to whitelist certain trusted domains for the image sources. This closes #5333.

For reference: We discussed a use case where an Icinga Web module depends on features from an external provider, such as OpenStreetMap. Without the ability to modify the CSP header, every user of the module would need to adjust or override the web server configuration. A more effective approach we considered is to implement specific functionality in Icinga Web to modify only certain parts of the header. Additionally, I think such functionality would also be necessary for https://github.com/nbuchwitz/icingaweb2-module-map for example.

nilmerg (changes requested):
There are some style issues as well. Please rebase with main and the actions will run and point this out.

nilmerg (changes requested):
Thanks for keeping at it!

I didn't look at it entirely yet. Only one thing for now:

The way you load navigation items is not correct yet. You forgot dashlets and you only load shared items the user is an owner of.

Take a look at `\Icinga\Web\Navigation\Navigation::load` for example. Though, please note that for dashlets a more complex solution is needed. Dashlets by modules can be loaded there (and should! Also modules may provide external URLs this way), but not user dashlets. This is done by `\Icinga\Web\Widget\Dashboard::load` instead.

Also, please remember that there are additional types of navigation items. A module can provide its own as well using `\Icinga\Application\Modules\Module::provideNavigationItem`. Monitoring's and Icinga DB Web's host and service actions are an example which may also result in an iframe.

So please restore the `NavigationController` and think about an alternative to the `NavigationItemHelper`.

gianlucapiccolo:
hello @nilmerg can you please give us a feedback on the changes we made? Thanks!!

library/Icinga/Application/Hook/CspDirectiveHook.php
```diff
@@ -0,0 +1,18 @@
+<?php
+/* Icinga Web 2 | (c) 2022 Icinga GmbH | GPLv2+ */
+
+namespace Icinga\Application\Hook;
+
+abstract class CspDirectiveHook
```
Al2Klimov, line 6: As this is a new hook, add the new methods just like in #5433:

* all()
* register()

Al2Klimov (changes requested):

application/controllers/NavigationController.php
```diff
@@ -444,4 +444,4 @@ public function dashboardAction()
         $this->view->navigation = $navigation;
         $this->view->title = $navigation->getLabel();
     }
-}
+}
```
nilmerg, line 447: Please restore the empty line at the end of the file as well.

library/Icinga/Application/Hook/CspDirectiveHook.php
```diff
@@ -0,0 +1,18 @@
+<?php
+/* Icinga Web 2 | (c) 2022 Icinga GmbH | GPLv2+ */
```
nilmerg, line 2:
```suggestion
// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
// SPDX-License-Identifier: GPL-3.0-or-later
```

library/Icinga/Util/Csp.php
```diff
@@ -108,4 +199,104 @@ protected static function getInstance(): self
...
+            $configShared = Config::navigation($type);
+            $configShared->getConfigObject()->setKeyColumn('name');
+            foreach ($configShared->select() as $itemConfig) {
+                if (Icinga::app()->hasAccessToSharedNavigationItem($itemConfig, $config) && $itemConfig->get("target", "") !== "_blank") {
+                    $menuItems[] = ["name" => $itemConfig->get('name'), "url" => $itemConfig->get('url')];
+                }
+            }
+        }
+        return $menuItems;
+    }
```
nilmerg, line 252: I was rather thinking about a solution like this: (Untested)

I didn't ensure this fits with your remaining code, so adjust it accordingly as you see fit.

```suggestion
    if (! Auth::isAuthenticated()) {
        return [];
    }

    $origins = [];
    foreach (Navigation::load($type) as $navItem) {
         foreach (self::yieldNavigation($navItem) as $name => $url) {
             $origins[] = $url->getScheme() . '://' . $url->getHost();
         }
    }
    
    return $origins;
}

protected static function yieldNavigation(NavigationItem $item): Generator
{
    if ($item->hasChildren()) {
        foreach ($item as $child) {
            yield from self::yieldNavigation($child);
        }
    } else {
        if ($item->getTarget() !== '_blank' && $item->getUrl()->isExternal()) {
            yield $item->getName() => $item->getUrl();
        }
    }
}

```

library/Icinga/Application/Web.php
```diff
@@ -176,7 +176,7 @@ public function getViewRenderer()
         return $this->viewRenderer;
     }
 
-    private function hasAccessToSharedNavigationItem(&$config, Config $navConfig)
+    public function hasAccessToSharedNavigationItem(&$config, Config $navConfig)
```
nilmerg, line 179: I don't think that's needed if you use my proposal.

library/Icinga/Util/Csp.php
```diff
@@ -44,20 +54,97 @@ private function __construct()
...
+                Logger::debug("$errorSource: Skipping invalid url: $host");
+                continue;
+            }
+
+            $scheme = parse_url($navigationItem["url"], PHP_URL_SCHEME);
+
 
+            if ($host === null) {
+                continue;
+            }
```
nilmerg, line 105: The validation is already performed at least partially in the other methods. Please make sure to do this only once. Preferably during the fetch logic, as in my proposal.

nilmerg (changes requested):
Please rebase with main as the CI configuration changed.

Didn't test yet but it looks promising already!

v2.13 is on the horizon, so ensure to get back to it soon if you want this being part of it. Or drop a comment to let me delegate this further if you don't have the time. (But then allow collaborator access for us)

lippserd:
Please excuse me for not informing everyone, but I discussed with WP that we will take over this PR.

library/Icinga/Application/Hook/CspDirectiveHook.php
TheSyscall: Implemented in https://github.com/Icinga/icingaweb2/pull/5477/changes/b9393406afe8a8a1a6888cc7eacbdbfdaca51abb (https://github.com/Icinga/icingaweb2/pull/5477)

application/controllers/NavigationController.php
TheSyscall: Changes to `NavigationController`have been reverted in https://github.com/Icinga/icingaweb2/pull/5477

library/Icinga/Application/Hook/CspDirectiveHook.php
Al2Klimov: Neither I'd add GPLv3 files in a (yet) GPLv2 project (our corporate lawyer has left the chat 🙈), nor I'd mix license header formats in one and the same project. Latter especially because the GPLv3 PR (which doesn't even exist yet, by the way!) will likely use xargs to apply the same command to all files. The more files have the same header format, the better.

library/Icinga/Util/Csp.php
TheSyscall: Implemented your suggestion in #5477

library/Icinga/Application/Web.php
TheSyscall: Yes. Reverted it in #5477

library/Icinga/Util/Csp.php
```diff
@@ -44,20 +54,97 @@ private function __construct()
...
+            $policy = $host;
+            if ($scheme !== null) {
+                $policy = "$scheme://$host";
+            }
+
+            $cspDirectives['frame-src'][] = $policy;
+        }
+        // Allow modules to add their own csp directives in a limited fashion.
+        /** @var CspDirectiveHook $hook */
+        foreach (Hook::all('CspDirective') as $hook) {
```
Al2Klimov, line 116: This method is deprecated since fd5f0e3c7182d942880b22bfdf556043bab48dda.

Al2Klimov (changes requested):

library/Icinga/Util/Csp.php
lippserd: > This method is deprecated since [fd5f0e3](https://github.com/Icinga/icingaweb2/commit/fd5f0e3c7182d942880b22bfdf556043bab48dda).

We haven taken over this PR. Nothing to do here.

PR #5341, opened 2025-03-25
Pdf: Ensure dompdf can create temporary files

nilmerg:
This is required since dompdf seems to automatically load our custom font and complains otherwise.


nilmerg:
Based on https://github.com/dompdf/dompdf/wiki/About-Fonts-and-Character-Encoding

PR #5347, opened 2025-03-28
Dashlet: Properly embed iframe urls

### LINKED ISSUE #5346: External URLs set up as dashlets are not *embedded* the same as navigation items
## Describe the bug
Users are able to set up custom dashlets in the dashboard. By default external urls are not permitted. However, the user can utilize the iframe endpoint to circumvent this, manually.

This leads to embedded content, that isn't detected as such and hence the iframe endpoint refuses to load the url right away. The user has to manually approve this.

## To Reproduce

1. Create a new dashlet
2. Use an external url and prepend `iframe?url=`
3. Open the dashboard where the dashlet is part of

## Expected behavior
The url is automatically loaded.

## Your Environment

* Icinga Web 2 version and modules (System - About): 2.12.4

## Additional context
ref/IP/58637

nilmerg:
fixes #5346

### DIFF
--- a/library/Icinga/Web/Widget/Dashboard/Dashlet.php
+++ b/library/Icinga/Web/Widget/Dashboard/Dashlet.php
@@ -3,9 +3,9 @@
 
 namespace Icinga\Web\Widget\Dashboard;
 
+use Icinga\Web\Session;
 use Icinga\Web\Url;
 use Icinga\Data\ConfigObject;
-use Icinga\Exception\IcingaException;
 
 /**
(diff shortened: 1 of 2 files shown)


raviks789:
The urlHash is lost once the page is refreshed and you get the warning message instead of the content that was embedded.

nilmerg:
If, by refreshing you mean a full reload, this cannot be avoided.

raviks789 (approved):
LGTM

PR #5351, opened 2025-04-07
Patch-Csv.php-for-0-values

Tqnsls:


library/Icinga/File/Csv.php
```diff
@@ -37,7 +37,7 @@ public function __toString()
             }
             $out = array();
             foreach ($row as & $val) {
-                $out[] = '"' . ($val ? str_replace('"', '""', $val) : '') . '"';
+                $out[] = '"' . ($val === '0' ? '0' : ($val ? str_replace('"', '""', $val) : '')) . '"';
```
nilmerg, line 40: Just in case it's really zero ;)

```suggestion
                $out[] = '"' . ($val == '0' ? '0' : ($val ? str_replace('"', '""', $val) : '')) . '"';
```

nilmerg (changes requested):

nilmerg (approved):

PR #5355, opened 2025-04-23
Fix: spacing and text-breaking

### LINKED ISSUE #5244: Collapsed menu issues
## Describe the bug
1) When the sidebar is collapsed ellipsisizing doesn't work on level-2 nav-items in flyouts.
2) Also there's a margin on the icon, which produces an unnecessary gap.

Screenshot 2024-08-13 at 11 05 45

## To Reproduce
for 1)
1. Collapse sidebar
2. Open a level-2 menu flyout containing a nav-item with overflowing text
3. Notice the overflowing text not being ellipsisized

for 2)
1. Collapse sidebar
2. Open any level-2 menu flyout
3. Notice the unnecessary gap in the place, where the item icon would be

## Expected behavior
Should be clear from the description

## Your Environment
* icingaweb2 main

Jan-Schuppik:
Fix #5244

### DIFF
--- a/public/css/icinga/layout.less
+++ b/public/css/icinga/layout.less
@@ -333,6 +333,10 @@
     .nav-level-1 > .nav-item i {
       font-size: 1.5em;
       margin-right: .5em;
+
+      &.icon-letter {
+        margin-right: 0;
+      }
     }
 
     > .search-control {
--- a/public/css/icinga/layout.less
+++ b/public/css/icinga/layout.less
@@ -332,10 +332,9 @@
 
     .nav-level-1 > .nav-item i {
       font-size: 1.5em;
-      margin-right: .5em;
 
-      &.icon-letter {
-        margin-right: 0;
+      &:not(.nav-level-2 > .nav-item i) {
+        margin-right: .5em;
       }
     }
 
--- a/public/css/icinga/layout.less
+++ b/public/css/icinga/layout.less
@@ -330,12 +330,10 @@
       padding-left: .75em;
     }
 
-    .nav-level-1 > .nav-item i {
+    .nav-level-1 > .nav-item > a > i,
+    .nav-level-1 > .nav-item > span > i {
       font-size: 1.5em;
-
-      &:not(.nav-level-2 > .nav-item i) {
-        margin-right: .5em;
-      }
+      margin-right: .5em;
     }
 
     > .search-control {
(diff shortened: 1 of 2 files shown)


flourish86 (dismissed):
Looks to me like the issues have been solved.

public/css/icinga/layout.less
```diff
@@ -333,6 +333,10 @@
     .nav-level-1 > .nav-item i {
       font-size: 1.5em;
       margin-right: .5em;
```
nilmerg, line 335: This is actually related to the specificity of its rule. This margin is only relevant to level 1 nav items when the sidebar is collapsed. At the moment, level 2 nav items are also affected because they are contained and also have items with icons. The icon `.nav-item i` selector must be limited to only affect icons related to level 1 nav items.

public/css/icinga/menu.less
```diff
@@ -286,6 +286,8 @@
 
         &:first-of-type {
           padding-left: 1.5em;
+          text-overflow: ellipsis;
+          overflow: hidden;
```
nilmerg, line 290: There is already a rule for this, albeit not for the collapsed menu. Please re-use it.

public/css/icinga/layout.less
```diff
@@ -332,7 +332,10 @@
 
     .nav-level-1 > .nav-item i {
       font-size: 1.5em;
-      margin-right: .5em;
+
+      &:not(.nav-level-2 > .nav-item i) {
```
nilmerg, line 336: This can be solved more easily. Also, both properties are only relevant to level 1. So just extend the existing selector to: 
```
.nav-level-1 > .nav-item > a > i,
.nav-level-1 > .nav-item > span > i {
```

nilmerg (changes requested):

nilmerg (changes requested):
The new flyout headings are still cut off:
image

public/css/icinga/menu.less
```diff
@@ -537,9 +542,10 @@ html.no-js  #toggle-sidebar {
 
   span {
     padding-left: 1.375em;
+    padding-right: 0.364em;
```
nilmerg, line 545: The anchors below use `0.545em`

nilmerg (changes requested):

nilmerg (approved):

PR #5356, opened 2025-04-23
about/index.php: Add message if database connection fail

### LINKED ISSUE #5155: Fix stacktrace for configuration database in the general configuration
## Describe the bug
When the configuration backend is not set correctly (e.g. throuch installing with the ansible collection ) the about section throws an inaccurate stacktrace. See: https://community.icinga.com/t/icingadbweb2-v1-1-icinga2-r2-14-0-1-error-in-about-section/12912

## To Reproduce
Provide a link to a live example, or an unambiguous set of steps to reproduce this issue. Include configuration, logs, etc. to reproduce, if relevant.

1. Install Icingaweb2 with ansible collection or select an invalid database in general config
2. Click about

## Expected behavior
An error message that guides me to the necessary option

## Your Environment

JolienTrog:
fixes #5155

### DIFF
--- a/application/controllers/ErrorController.php
+++ b/application/controllers/ErrorController.php
@@ -8,6 +8,7 @@ namespace Icinga\Controllers;
 use Icinga\Application\Hook\DbMigrationHook;
 use Icinga\Application\MigrationManager;
 use Icinga\Exception\IcingaException;
+use Throwable;
 use Zend_Controller_Plugin_ErrorHandler;
 use Icinga\Application\Icinga;
 use Icinga\Application\Logger;
@@ -98,17 +99,21 @@ class ErrorController extends ActionController
                         $mm = MigrationManager::instance();
                         $action = $this->getRequest()->getActionName();
                         $controller = $this->getRequest()->getControllerName();
-                        if ($action !== 'hint' && $controller !== 'migrations' && $mm->hasMigrations($moduleName)) {
-                            // The view renderer from IPL web doesn't render the HTML content set in the respective
-                            // controller if the error_handler request param is set, as it doesn't support error
-                            // rendering. Since this error handler isn't caused by the migrations controller, we can
-                            // safely unset this.
-                            $this->setParam('error_handler', null);
-                            $this->forward('hint', 'migrations', 'default', [
-                                DbMigrationHook::MIGRATION_PARAM => $moduleName
-                            ]);
+                        try {
+                            if ($action !== 'hint' && $controller !== 'migrations' && $mm->hasMigrations($moduleName)) {
+                                // The view renderer from IPL web doesn't render the HTML content set in the respective
+                                // controller if the error_handler request param is set, as it doesn't support error
+                                // rendering. Since this error handler isn't caused by the migrations controller, we can
+                                // safely unset this.
+                                $this->setParam('error_handler', null);
+                                $this->forward('hint', 'migrations', 'default', [
+                                    DbMigrationHook::MIGRATION_PARAM => $moduleName
+                                ]);
 
-                            return;
+                                return;
+                            }
+                        } catch (Throwable $e) {
+                            Logger::error("%s\n%s", $e, IcingaException::getConfidentialTraceAsString($e));
                         }
 
                         $this->getResponse()->setHttpResponseCode(500);
--- a/application/controllers/MigrationsController.php
+++ b/application/controllers/MigrationsController.php
@@ -7,8 +7,10 @@ namespace Icinga\Controllers;
 
 use Icinga\Application\Hook\DbMigrationHook;
 use Icinga\Application\Icinga;
+use Icinga\Application\Logger;
 use Icinga\Application\MigrationManager;
 use Icinga\Common\Database;
+use Icinga\Exception\IcingaException;
 use Icinga\Exception\MissingParameterException;
 use Icinga\Forms\MigrationForm;
 use Icinga\Web\Notification;
@@ -18,8 +20,11 @@ use ipl\Html\Attributes;
 use ipl\Html\FormElement\SubmitButtonElement;
 use ipl\Html\HtmlElement;
 use ipl\Html\Text;
+use ipl\Web\Common\CalloutType;
 use ipl\Web\Compat\CompatController;
 use ipl\Web\Widget\ActionLink;
+use ipl\Web\Widget\Callout;
+use Throwable;
 
 class MigrationsController extends CompatController
 {
@@ -56,7 +61,20 @@ class MigrationsController extends CompatController
 
         $migrateListForm = new MigrationForm();
         $migrateListForm->setAttribute('id', $this->getRequest()->protectId('migration-form'));
-        $migrateListForm->setRenderDatabaseUserChange(! $mm->validateDatabasePrivileges());
+        try {
+            $hasPrivileges = $mm->validateDatabasePrivileges();
+        } catch (Throwable $e) {
+            Logger::error("%s\n%s", $e, IcingaException::getConfidentialTraceAsString($e));
+            $this->addContent(new Callout(
+                CalloutType::Error,
+                sprintf($this->translate('Please check the log for details: %s'), $e->getMessage()),
+                $this->translate('Failed to load pending migrations'),
+            ));
+
+            return;
+        }
+
+        $migrateListForm->setRenderDatabaseUserChange(! $hasPrivileges);
 
         if ($canApply && $mm->hasPendingMigrations()) {
             $migrateAllButton = new SubmitButtonElement(sprintf('migrate-%s', DbMigrationHook::ALL_MIGRATIONS), [
--- a/application/controllers/MyDevicesController.php
+++ b/application/controllers/MyDevicesController.php
@@ -5,11 +5,19 @@
 
 namespace Icinga\Controllers;
 
+use Icinga\Application\Logger;
 use Icinga\Common\Database;
-use Icinga\Web\Notification;
 use Icinga\Web\RememberMe;
 use Icinga\Web\RememberMeUserDevicesList;
+use ipl\Html\Attributes;
+use ipl\Html\HtmlElement;
+use ipl\Html\Text;
+use ipl\Web\Common\CalloutType;
 use ipl\Web\Compat\CompatController;
+use ipl\Web\Widget\Callout;
+use ipl\Web\Widget\Icon;
+use ipl\Web\Widget\Link;
+use Throwable;
 
 /**
  * MyDevicesController
@@ -48,6 +56,30 @@ class MyDevicesController extends CompatController
 
     public function indexAction()
     {
+        try {
+            $this->getDb();
+        } catch (Throwable $e) {
+            Logger::error("%s\n%s", $e, $e->getTraceAsString());
+            if ($this->hasPermission('config/*')) {
+                $errorMessage = $this->translate(
+                    'To establish a valid database connection set the Configuration'
+                    . ' Database field in the Application Settings.'
+                );
+            } else {
+                $errorMessage = $this->translate(
+                    'You don`t have permission to change this setting. Please contact an administrator.'
+                );
+            }
+
+            $this->addContent(new Callout(
+                CalloutType::Error,
+                $errorMessage,
+                $this->translate('The configuration database has not been configured'),
+            ));
+
+            return;
+        }
+
         $name = $this->auth->getUser()->getUsername();
 
         $data = (new RememberMeUserDevicesList())
@@ -56,12 +88,6 @@ class MyDevicesController extends CompatController
             ->setUrl('my-devices/delete');
 
         $this->addContent($data);
-
-        if (! $this->hasDb()) {
-            Notification::warning(
-                $this->translate("Users can't stay logged in without a database configuration backend")
-            );
-        }
     }
 
     public function deleteAction()
--- a/application/forms/Config/General/ApplicationConfigForm.php
+++ b/application/forms/Config/General/ApplicationConfigForm.php
@@ -5,6 +5,7 @@
 
 namespace Icinga\Forms\Config\General;
 
+use Icinga\Application\Config;
 use Icinga\Application\Icinga;
 use Icinga\Data\ResourceFactory;
 use Icinga\Web\Form;
@@ -89,6 +90,16 @@ class ApplicationConfigForm extends Form
                 'label'         => $this->translate('Configuration Database')
             ]
         );
+
+        if (! isset(Config::app()->getSection('global')->config_resource)) {
+            $this->warning(
+                $this->translate(
+                    'The configuration database has not been configured.'
+                    . ' Please set the required field to establish a valid database connection.'
+                ),
+                false,
+            );
+        }
 
         return $this;
     }
--- a/application/forms/PreferenceForm.php
+++ b/application/forms/PreferenceForm.php
@@ -19,6 +19,7 @@ use Icinga\Web\Notification;
 use Icinga\Web\Session;
 use Icinga\Web\StyleSheet;
 use ipl\Html\HtmlElement;
+use ipl\Html\Text;
 use ipl\I18n\GettextTranslator;
 use ipl\I18n\Locale;
 use ipl\I18n\StaticTranslator;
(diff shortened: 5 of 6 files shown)


nilmerg:
Seems like this was already in part addressed by #5236, which hid the stacktrace completely. You're now re-introducing it, but with a different message, even though I said to not show a stacktrace anymore!

But anyway, let's start from scratch here. Please do the following:

* Drop the option `config_resource` (Section `global`) from `/etc/icingaweb2/config.ini`
  This will simulate the case that no configuration database has been configured
* Note down the cases where migrations are checked
  Utilize PHPStorm's `Find Usages` for this
* Visit the routes that will trigger a check and see how they cope with a missing/invalid resource
  We know already that the about page will suppress the error, but others might not

Then, depending on the individual case, please propose a solution how it can be improved.
In case of the about page, a message at the location where pending migrations are usually shown for example. (Above loaded libraries)

application/controllers/MigrationsController.php
```diff
@@ -55,24 +58,48 @@ public function indexAction(): void
...
+                    new Icon('warning'),
+                    new HtmlElement('ul', null),
+                    new HtmlElement(
+                        'p',
+                        null,
+                        new Text(
+                            $this->translate('   No Configuration Database selected. 
+                            To establish a valid database connection set the Configuration Database field.   ')
+                        )
+                    ),
```
raviks789:
```suggestion
                        new Text($this->translate(
                            'No Configuration Database selected. '
                            . 'To establish a valid database connection set the Configuration Database field.'
                        ))
```

application/controllers/MigrationsController.php
```diff
@@ -55,24 +58,48 @@ public function indexAction(): void
...
+                    new HtmlElement('ul', null),
+                    new Link($this->translate('Configuration Database'), 'config/general/')
```
raviks789:
```suggestion
                    new Link($this->translate('Configuration Database'), 'config/general')
```

No need to add the trailing '/' in the url.

application/controllers/MyDevicesController.php
```diff
@@ -49,6 +54,46 @@ public function init()
...
+                            new Text($this->translate(
+                                'No Configuration Database selected.'
+                                 . 'You don`t have permission to change this setting. Please contact an administrator.'
+                            ))
+                        )
+                    )
+                );
+            }
+            return;
+        }
```
raviks789:
```suggestion
            $configLink = new HtmlDocument();
            if ($hasConfigPermission) {
                $warningMessage = $this->translate(
                    'No Configuration Database selected.'
                    . 'To establish a valid database connection set the Configuration Database field.'
                );

                $configLink = new Link($this->translate('Configuration Database'), 'config/general');
            } else {
                $warningMessage = $this->translate(
                    'No Configuration Database selected.'
                    . 'You don`t have permission to change this setting. Please contact an administrator.'
                );
            }

            $this->addContent(
                new HtmlElement(
                    'div',
                    new Attributes(['class' => 'db-connection-warning']),
                    new Icon('warning'),
                    new HtmlElement(
                        'p',
                        null,
                        Text::create($warningMessage),
                    ),
                    $configLink
                )
            );

            return;
        }

```
With this the common code will be kept outside the if-else block. And the repetition of code will be avoided.

application/forms/PreferenceForm.php
```diff
@@ -185,7 +186,15 @@ public function createElements(array $formData)
                 false
             );
         }
-
+        $config = Config::app()->getSection('global');
```
raviks789, line 192: Add new line after if block.

application/forms/PreferenceForm.php
```diff
@@ -185,7 +186,15 @@ public function createElements(array $formData)
...
+        if (!isset($config->config_resource)) {
```
raviks789:
```suggestion
        if (! isset($config->config_resource)) {
```
There should be a space after `!` operator.

application/forms/PreferenceForm.php
```diff
@@ -185,7 +186,15 @@ public function createElements(array $formData)
...
+            $missingConfigResource =
+                Text::create($this->translate(
+                    'No Configuration Database selected.'
+                    . 'To establish a valid database connection set the Configuration Database field.'
+                ));
+            $this->warning($missingConfigResource, false);
```
raviks789:
```suggestion
            $this->warning(
                $this->translate(
                    'No Configuration Database selected.'
                    . 'To establish a valid database connection set the Configuration Database field.'),
                false
            );
```
No need for `Text::create` and variable `$missingConfigResource`.

application/forms/Config/General/ApplicationConfigForm.php
```diff
@@ -100,6 +102,16 @@ public function createElements(array $formData)
...
 
+        $config = Config::app()->getSection('global');
+        if (!isset($config->config_resource)) {
+            $missingConfigResource =
+                Text::create(
+                    $this->translate("No Configuration Database selected. 
+                    Please set the field to establish a valid database connection.")
+                );
+            $this->warning($missingConfigResource, false);
+        }
```
raviks789:
```suggestion
        if (! isset($config->config_resource)) {
            $this->warning(
                $this->translate('No Configuration Database selected.' 
                    . 'Please set the field to establish a valid database connection.'
                ),
                false
            );
        }
```

application/views/scripts/about/index.phtml
```diff
@@ -104,10 +104,35 @@ use ipl\Web\Widget\StateBadge;
...
+        <div class="pending-migrations clearfix">
+            <?php if($mm->getPendingMigrations() || (!isset($config->config_resource))) : ?>
+            <h2><?= $this->translate('Pending Migrations') ?></h2>
+            <?php endif ?>
+            <table >
+                <tr>
+                    <th>
+                  <?php  if (!isset($config->config_resource)) : ?>
+                      <?=  new EmptyState(
+                              $this->translate('No Configuration Database selected. To establish a valid database connection set: '))
```
raviks789:
```suggestion
                              $this->translate('No Configuration Database selected. To establish a valid database connection set: ')
                    )
```

application/views/scripts/about/index.phtml
```diff
@@ -104,10 +104,35 @@ use ipl\Web\Widget\StateBadge;
...
+                  ?><td>
+                  <?= $this->qlink(
+                      $this->translate('Configuration Database'),
+                      'config/general/',
+                      array('name' => ('Configuration Database')),
```
raviks789:
```suggestion
                      [],
```
No Url parameters needed here.

application/views/scripts/about/index.phtml
```diff
@@ -104,10 +104,35 @@ use ipl\Web\Widget\StateBadge;
...
+                      array('title' => sprintf($this->translate('Go to Application/General')), 'Configuration Database')
```
raviks789:
```suggestion
                      ['title' => sprintf($this->translate('Go to Application/General')), 'Configuration Database']
```

raviks789 (changes requested):

application/forms/PreferenceForm.php
JolienTrog: Really? Is this always the case? I've seen it without spaces elsewhere.

application/controllers/ErrorController.php
```diff
@@ -96,17 +97,23 @@ public function errorAction()
...
-                            // controller if the error_handler request param is set, as it doesn't support error
-                            // rendering. Since this error handler isn't caused by the migrations controller, we can
-                            // safely unset this.
-                            $this->setParam('error_handler', null);
-                            $this->forward('hint', 'migrations', 'default', [
-                                DbMigrationHook::MIGRATION_PARAM => $moduleName
-                            ]);
-
-                            return;
+                        $hasPending = false;
```
raviks789: Is there a reason you need this variable `$hasPending`.

I don't see this variable used anywhere. In the try block, you're assigning it a value, but it's not used afterward. You could consider removing it.

application/forms/PreferenceForm.php
raviks789: I see that this is not resolved. There is still no newline after this if block.

application/views/scripts/about/index.phtml
```diff
@@ -104,10 +104,36 @@ use ipl\Web\Widget\StateBadge;
...
+        </div>
+        <?php $config = Config::app()->getSection('global') ?>
+        <div class="pending-migrations clearfix">
+            <?php if($mm->getPendingMigrations() || (!isset($config->config_resource))) : ?>
+            <h2><?= $this->translate('Pending Migrations') ?></h2>
+            <?php endif ?>
+            <table >
+                <tr>
+                    <th>
+                  <?php  if (!isset($config->config_resource)) : ?>
```
raviks789:
```suggestion
                  <?php  if (! isset($config->config_resource)) : ?>
```
We usually add a space after the ! (not) operator for readability.

application/controllers/MyDevicesController.php
```diff
@@ -49,6 +54,42 @@ public function init()
...
+                    new Attributes(['class' => 'db-connection-warning']),
+                    new Icon('warning'),
+                    new HtmlElement(
+                        'p',
+                        null,
+                        Text::create($warningMessage),
+                    ),
+                    $configLink
+                )
+            );
```
raviks789:
```suggestion
            $hasConfigPermission = $this->hasPermission('config/*');
            if ($hasConfigPermission) {
                $configLink = new Link($this->translate('Configuration Database'), 'config/general');
                $warningMessage = sprintf($this->translate(
                    'No configuration database is selected. '
                    . 'To establish a valid database connection set the %s field.'
                ), $configLink);
            } else {
                $warningMessage = Text::create(
                    $this->translate(
                        'No configuration database is selected. '
                        . 'You don`t have permission to change this setting. Please contact an administrator.'
                    )
                );
            }

            $this->addContent(
                new HtmlElement(
                    'div',
                    new Attributes(['class' => 'db-connection-warning']),
                    new Icon('warning'),
                    new HtmlElement(
                        'p',
                        null,
                        HtmlString::create($warningMessage),
                    )
                )
            );
```

I have changed the message and also included the link within the message, which made more sense to me. Also, add `ipl\Html\HtmlString` to the usages at the top of the file.

application/forms/Config/General/ApplicationConfigForm.php
```diff
@@ -100,6 +102,17 @@ public function createElements(array $formData)
             )
         );
 
+        $config = Config::app()->getSection('global');
+        if (! isset($config->config_resource)) {
+            $this->warning(
+                $this->translate(
+                    'No Configuration Database selected.'
+                    . 'Please set the field to establish a valid database connection.'
```
raviks789:
```suggestion
                    'The configuration database has not been selected. '
                    . 'Please set the required field to establish a valid database connection.'
```

application/forms/PreferenceForm.php
```diff
@@ -185,7 +186,16 @@ public function createElements(array $formData)
                 false
             );
         }
-
+        $config = Config::app()->getSection('global');
+        if (! isset($config->config_resource)) {
+            $this->warning(
+                $this->translate(
+                    'No Configuration Database selected.'
+                    . 'To establish a valid database connection set the Configuration Database field.'
```
raviks789:
```suggestion
                    'The configuration database is selected. '
                    . 'To establish a valid database connection set the configuration database field.'
```

I am curious why the link to set the configuration database was not included here like in `MyDevicesController`.

application/views/scripts/about/index.phtml
```diff
@@ -104,10 +104,36 @@ use ipl\Web\Widget\StateBadge;
...
-        if ($hasPending): ?>
-            <div class="pending-migrations clearfix">
-                <h2><?= $this->translate('Pending Migrations') ?></h2>
-                <table class="name-value-table migrations">
+        ?>
+            </div>
+        </div>
+        <?php $config = Config::app()->getSection('global') ?>
+        <div class="pending-migrations clearfix">
+            <?php if($mm->getPendingMigrations() || (!isset($config->config_resource))) : ?>
```
raviks789:
```suggestion
            <?php if($mm->getPendingMigrations() || (! isset($config->config_resource))) : ?>
```
We usually add a space after the ! (not) operator for readability.

application/views/scripts/about/index.phtml
```diff
@@ -104,10 +104,36 @@ use ipl\Web\Widget\StateBadge;
...
+                              $this->translate('No Configuration Database selected. To establish a valid database connection set: ')
+                    )
+                  ?><td>
+                  <?= $this->qlink(
+                      $this->translate('Configuration Database'),
+                      'config/general/',
+                      [],
+                      ['title' => sprintf($this->translate('Go to Application/General')), 'Configuration Database']
+                  ) ?>
+                    </td>
```
raviks789:
```suggestion
                      <?=  new EmptyState(
                              HtmlString::create(sprintf(
                                      $this->translate('No configuration database is selected. To establish a valid database connection set the %s'),
                                      new Link($this->translate('Configuration Database'), 'config/general')
                              ))
                    )
                  ?>
```

`HtmlString` can render the html in the string. And it looks better to me.
You need to add `ipl\Html\HtmlString` and `ipl\Web\Widget\Link` at the top of the file.

raviks789 (changes requested):

application/forms/PreferenceForm.php
JolienTrog: I believe there’s a small mistake in the suggested text. The correct version should be: “The configuration database has not been selected.”
Also, if I remember correctly, it wasn’t possible to include a link in a Zend Form warning message, since the link would not be rendered. Perhaps you know a workaround?

application/views/scripts/about/index.phtml
JolienTrog: In other parts of the code, `qlink` has always been used. Should I now use `new Link `just this one time?

JolienTrog: Unfortunately, the text and the link are not displayed on the same line, that’s why I used a colon (:), because to me it looks better that way.

image

your suggestion:
image

What do you think?

application/forms/PreferenceForm.php
raviks789: > I believe there’s a small mistake in the suggested text. The correct version should be: “The configuration database has not been selected.”

You are right, use the correct version has you have mentioned.

> Also, if I remember correctly, it wasn’t possible to include a link in a Zend Form warning message, since the link would not be rendered. Perhaps you know a workaround?

Good question, I will look into it and will let you know if this is possible :)

application/views/scripts/about/index.phtml
raviks789: You need to check why this is happening. The issue appears to be caused by the current CSS rules, which are preventing the links from displaying on the same line. Consider updating the layout properties that are causing this.

raviks789: > In other parts of the code, `qlink` has always been used. Should I now use `new Link `just this one time?

It provides a straightforward way to include the link in the EmptyState widget. Hence, you could use it here.

application/forms/PreferenceForm.php
raviks789: @JolienTrog, easiest way is to add an EmptyState widget directly in the view script `application/views/scripts/account/index.phtml`, just like you did in `application/views/scripts/about/index.phtml` and use some css styling to make it appear like a warning. This way you could also include the link.

application/controllers/MigrationsController.php
```diff
@@ -55,24 +58,48 @@ public function indexAction(): void
...
+                $migrateAllButton->setWrapper(
+                    new HtmlElement('div', Attributes::create(['class' => 'icinga-controls']))
+                );
 
-            $this->controls->getAttributes()->add('class', 'default-layout');
-            $this->addControl($migrateAllButton);
+                $this->controls->getAttributes()->add('class', 'default-layout');
+                $this->addControl($migrateAllButton);
+            }
+        } catch (Throwable $e) {
```
Al2Klimov: Are you sure you need to `try {` such a large code fragment?

Al2Klimov: Given your specific error message, re-consider your caught type. (Throwable)

application/controllers/MigrationsController.php
```diff
@@ -55,24 +58,48 @@ public function indexAction(): void
...
+                        new Text($this->translate(
+                            'No Configuration Database selected. '
+                            . 'To establish a valid database connection set the Configuration Database field.'
+                        )),
+                        new HtmlElement('ul', null),
+                        new Link($this->translate('Configuration Database'), 'config/general')
+                    )
+                )
+            );
+            return;
```
Al2Klimov: Even if the framework has no configuration DB to migrate:

* Modules (the code below) may have pending migrations – but you're blocking them as well
* I'd even guess(!) no configuration DB = no need to migrate framework stuff = fine

Al2Klimov (changes requested):

JolienTrog:
@nilmerg :
I talked to @flourish86  and he suggested adding a notification similar to the health check and integrating it into the footer. Compared to a regular warning message, this notification is persistent and doesn’t disappear after a short time. That’s why I didn’t use the regular warning notification, but instead displayed the message at the top of the page. However, this approach is not consistent with the overall UI/UX.
The disadvantage would be that the existing development would basically have to be completely reworked. So, aside from the learning effect for me, a lot of development work would be wasted.

What do you think? Should I keep it like this or use the HealthHook instead?

Workaround version:
image


HealthHook verson:
image

PR #5360, opened 2025-04-25
Show title in level-2 flyouts

### LINKED ISSUE #4968: Sidebar: Show the parent level-1 label in level-2 flyouts
We should display the title in level-2 flyouts in the sidebar navigation, when the sidebar is collapsed.

While it’s totally fine, when the sidebar is expanded, it’s kind of confusing when the sidebar is collapsed.

Artboard

This could look something like this.

Artboard Copy

jrauh01:
Add a title to the navigation level-2 flyouts. The title is only displayed if the sidebar is collapsed.

Closes #4968

### DIFF
--- a/library/Icinga/Web/Navigation/Renderer/RecursiveMenuNavigationRenderer.php
+++ b/library/Icinga/Web/Navigation/Renderer/RecursiveMenuNavigationRenderer.php
@@ -1,0 +1,30 @@
+<?php
+/* Icinga Web 2 | (c) 2025 Icinga GmbH | GPLv2+ */
+
+namespace Icinga\Web\Navigation\Renderer;
+
+use Icinga\Web\Navigation\NavigationItem;
+
+/**
+ * Renderer for the multi level navigation in the sidebar menu
+ */
+class RecursiveMenuNavigationRenderer extends RecursiveNavigationRenderer
+{
+    public function beginChildren(): void
+    {
+        parent::beginChildren();
+
+        $parentItem = $this->getInnerIterator()->current()->getParent();
+        $item = new NavigationItem($parentItem->getName());
+        $item->setLabel($parentItem->getLabel());
+        $item->setCssClass('nav-item-header');
+
+        $renderer = new NavigationItemRenderer();
+        $renderer->setEscapeLabel(false);
+        $content = $renderer->render($item);
+
+        $this->content[] = $this->getInnerIterator()->beginItemMarkup($item);
+        $this->content[] = $content;
+        $this->content[] = $this->getInnerIterator()->endItemMarkup();
+    }
+}
--- a/public/css/icinga/menu.less
+++ b/public/css/icinga/menu.less
@@ -246,6 +246,10 @@
   margin-right: .5em
 }
 
+#layout:not(.sidebar-collapsed) #menu .nav-level-1 > .nav-item:not(.active).hover > .nav-level-2 {
+  padding-top: @vertical-padding;
+}
+
 // Hovered menu
 #layout:not(.minimal-layout).sidebar-collapsed #menu .nav-level-1 > .nav-item.hover,
 #layout:not(.minimal-layout) #menu .nav-level-1 > .nav-item:not(.active).hover {
@@ -255,7 +259,7 @@
     border-color: @gray-light;
     border-radius: .25em;
     box-shadow: 0 0 1em 0 rgba(0,0,0,.3);
-    padding: @vertical-padding 0;
+    padding-bottom: @vertical-padding;
     width: 14em;
     position: fixed;
     z-index: 1;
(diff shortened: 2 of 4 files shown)


nilmerg:
I'd rather like to see a no-JS approach

jrauh01:
Unlike discussed, I would use the parent name instead of label to avoid displaying state badges.

Using label:
Screenshot 2025-04-28 at 12 05 33

Using name:
Screenshot 2025-04-28 at 12 05 49

library/Icinga/Web/Navigation/Renderer/RecursiveMenuNavigationRenderer.php
```diff
@@ -0,0 +1,27 @@
...
+/* Icinga Web 2 | (c) 2025 Icinga GmbH | AGPLv3 */
+
+namespace Icinga\Web\Navigation\Renderer;
+
+/**
+ * Renderer for the multi level navigation in the sidebar menu
+ *
+ * @method NavigationRenderer getInnerIterator() {
+ *     {@inheritdoc}
+ * }
```
nilmerg, line 11: There's no reason to include this here, please remove.

library/Icinga/Web/Navigation/Renderer/RecursiveMenuNavigationRenderer.php
```diff
@@ -0,0 +1,27 @@
+<?php
+/* Icinga Web 2 | (c) 2025 Icinga GmbH | AGPLv3 */
```
nilmerg, line 2: Icinga Web is gplv2+

library/Icinga/Web/Navigation/Renderer/RecursiveMenuNavigationRenderer.php
```diff
@@ -0,0 +1,27 @@
...
+ */
+class RecursiveMenuNavigationRenderer extends RecursiveNavigationRenderer implements NavigationRendererInterface
+{
+    public function beginChildren(): void
+    {
+        parent::beginChildren();
+
+        $currentItemName = $this->getInnerIterator()->current()->getParent()->getName();
+
+        if ($this->getDepth() + 1 > 1) {
```
nilmerg, line 21: a) Here are children processed
b) The current name is accessed using the parent

Due to both, this check is redundant and always true.

library/Icinga/Web/Navigation/Renderer/RecursiveMenuNavigationRenderer.php
```diff
@@ -0,0 +1,27 @@
...
+            $this->content[] = '<li class="nav-item nav-item-header">'
+                . '<p><i aria-hidden="true" class="icon-letter"></i> ' . $currentItemName . '</p>'
+                . '</li>';
```
nilmerg, line 24: Please render this the usual way. (Like it's done in method render already.)
i.e. not *by hand*

nilmerg (changes requested):
When not collapsed, the first entry has none or a narrower gap at the top.

**Previous**
Screenshot from 2025-04-28 13-22-35

**After**
Screenshot from 2025-04-28 13-24-01

The height of the new element is not the same as the one in the sidebar:
image

library/Icinga/Web/Navigation/Renderer/RecursiveMenuNavigationRenderer.php
```diff
@@ -0,0 +1,25 @@
...
+ */
+class RecursiveMenuNavigationRenderer extends RecursiveNavigationRenderer implements NavigationRendererInterface
+{
+    public function beginChildren(): void
+    {
+        parent::beginChildren();
+
+        $item = new NavigationItem($this->getInnerIterator()->current()->getParent()->getName());
+        $item->setCssClass('nav-item-header');
+        $item->render();
```
nilmerg, line 19: 1. You're still ignoring the label.
2. The render result is ignored. You're rendering twice at the moment.

nilmerg: Remember to hide the label's badge using CSS.

library/Icinga/Web/Navigation/Renderer/RecursiveMenuNavigationRenderer.php
```diff
@@ -0,0 +1,25 @@
...
+/* Icinga Web 2 | (c) 2025 Icinga GmbH | GPLv2+ */
+
+namespace Icinga\Web\Navigation\Renderer;
+
+use Icinga\Web\Navigation\NavigationItem;
+
+/**
+ * Renderer for the multi level navigation in the sidebar menu
+ */
+class RecursiveMenuNavigationRenderer extends RecursiveNavigationRenderer implements NavigationRendererInterface
```
nilmerg, line 11: The base already implements the interface

public/js/icinga/behavior/navigation.js
```diff
@@ -319,7 +319,8 @@
             $target.siblings().not($target).removeClass('hover');
             $target.addClass('hover');
 
-            var targetHeight = $target.offset().top + $target.outerHeight();
+            let target = $target[0];
```
nilmerg, line 322: placement is still off, especially once the menu is scrollable. Retrieve top offset and height by using getBoundingClientRect instead.

nilmerg (changes requested):

public/css/icinga/menu.less
```diff
@@ -514,6 +518,28 @@ html.no-js  #toggle-sidebar {
...
+#layout.sidebar-collapsed #menu .nav-level-1 > .nav-item.hover .nav-level-2 > .nav-item-header {
+  background-color: @menu-bg-color;
+  border-bottom: 1px solid @gray-light;
+  border-top-left-radius: .25em;
+  border-top-right-radius: .25em;
+
+  span {
+    font-weight: @font-weight-bold;
+    line-height: 2.167em;
+    padding: .5em .5em .5em 1.375em;
```
nilmerg, line 530: Don't use line height and padding to achieve a desired height. Use a plain height instead. With #5363 there's a variable for this. (`@nav-item-height`)

To align the content, use flex box.

public/css/icinga/menu.less
```diff
@@ -514,6 +518,28 @@ html.no-js  #toggle-sidebar {
...
+    margin: 0;
```
nilmerg, line 531: spans have no margin so why do you reset it?

nilmerg (changes requested):
Please rebase with #5363

nilmerg (approved):

nilmerg (approved):

Issue #5263, opened 2024-09-20
filtering for older-than events with relative time does not work


restena-sw:
## Describe the bug

Time-based filters take the operators <, <= and >, >= and the value can be an absolute or relative time.

Using an absolute time for the filter value always works correctly. E.g. querying for

Service Time Last Ok >= 2024-01-01 selects all services that were OK sometime this year; and the complement
Service Time Last Ok < 2024-01-01 selects all services that were not in OK state this year, but earlier

Using a relative time for the filter value works only for > and >=; it always yields 0 results when attempted with < and <=. E.g. querying for

Service Time Last Ok >= 9m selects all services that were OK sometime in the last nine months; but the complement
Service Time Last Ok < 9m selects *nothing*. 

I also tried "-9m", "9 months" and "-9 months"; none of which produced the desired result.

The use case for such relative-time filter strings is the creation of a dashlet that shows problems that persist for a long time.

## To Reproduce

For a given set of services with issues, find a point in time that has start/resolve occurences before and after that point in time. Then filter as discussed above using a relative time string to express the chosen point in time.

## Expected behavior
The behaviour should be identical for absolute time and relative time values pointing to the same moment in time.

## Screenshots
Can add some (redacted for privacy) if needed; I hope I'm clear enough with words.

## Your Environment

* Icinga Web 2 version and modules (System - About): 2.9 (but confirmed in a 2.12 that relative time still does not work, maybe in a different way)
* Web browser used: Chromiusm 128
* Icinga 2 version used (`icinga2 --version`): 2.13.2, OS package
* Server operating system and version: CentOS 7 (now Suse Liberty Linux 7)

nilmerg:
Hi,

`9m` is an invalid value. Try `9 minutes`.

To see what's working, try https://strtotime.co.uk/

Though, we are aware that this is far from being useful and there is a proper UI validation/assistance required.

restena-sw:
I did try "x timespan" like as I wrote initially. So here are the screenshots with a "2 weeks" filter proving that it does not work:

- filtering for <= and an absolute time working fine, showing various hits for "old" issues:
![icinga_abs_time_working](https://github.com/user-attachments/assets/b49cab94-769b-4b54-9fa0-f4c40d2509d7)

- filtering with <= and both "2 weeks" and "-2 weeks" not showing any results
- 
[...]

PR #5362, opened 2025-04-28
ServicestatusQuery: Fix that timestamp columns work with relative time filter

sukhwinder33445:
fixes https://github.com/Icinga/icingaweb2/issues/5263


doc/80-Upgrading.md
```diff
@@ -3,6 +3,16 @@
 Specific version upgrades are described below. Please note that upgrades are incremental. An upgrade from
 v2.6 to v2.8 requires to follow the instructions for v2.7 too.
 
+## Upgrading to Icinga Web 2.12.5
```
nilmerg, line 6: This will only be part of 2.13

nilmerg (changes requested):

nilmerg (approved):

PR #5367, opened 2025-05-07
Menu navigation: Improve flyout transition

### LINKED ISSUE #5160: Sidebar Navigation: Level 2 Usability Improvement
We're not the first to have this issue.

When navigating to level 2 items in the sidebar, the flyout unexpecedly closes when the user tries to move the cursor to the desired level 2 menu item.

On the direct path to the item the cursors leaves the hover area of the main item, which makes the flyout disappear.

Frame 28

We've been affirmed that most users ran into the problem at first and have to learn how to avoid this issue.

https://github.com/Icinga/icingaweb2/assets/6977434/a07c0a3f-7784-4294-9bf6-b21a5c4f0fca

We're also not the first to solve this. Here's an article coming up with different solutions.
https://css-tricks.com/dropdown-menus-with-more-forgiving-mouse-movement-paths/

raviks789:
resolves #5160

### DIFF
--- a/public/js/icinga/behavior/navigation.js
+++ b/public/js/icinga/behavior/navigation.js
@@ -9,7 +9,6 @@
     try {
         var d3 = require("icinga/icinga-php-thirdparty/mbostock/d3");
     } catch (e) {
-        d3 = null;
         console.warn('D3.js library is unavailable. Navigation to flyout may not work as expected.');
     }
 
@@ -18,7 +17,7 @@
         this.on('click', '#menu a', this.linkClicked, this);
         this.on('click', '#menu tr[href]', this.linkClicked, this);
         this.on('rendered', '#menu', this.onRendered, this);
-        if (d3 !== null) {
+        if (typeof d3 !== "undefined") {
             this.on('mousemove', '#menu .primary-nav .nav-level-1 > .nav-item', this.onMouseMove, this);
         }
 
@@ -321,7 +320,7 @@
         const $target = $(this);
 
         if (
-            d3 !== null
+            typeof d3 !== "undefined"
             && ! _this.extendedFlyoutZone.includes(undefined)
             && d3.polygonContains(_this.extendedFlyoutZone, [e.clientX, e.clientY])
         ) {
@@ -355,15 +354,25 @@
 
         const $target = $(this);
 
-        if (
-            $target[0].matches(':has(.nav-level-2)')
-            && ! $target.hasClass('hover')
-            && ! _this.extendedFlyoutZone.includes(undefined)
-            && d3.polygonContains(_this.extendedFlyoutZone, [e.clientX, e.clientY])
-        ) {
-            _this.flyoutTimer = setTimeout(function() {
+        if (! $target[0].matches(':has(.nav-level-2)')) {
+            return;
+        }
+
+        if (! $target.hasClass('hover')) {
+            if (
+                ! _this.extendedFlyoutZone.includes(undefined)
+                && d3.polygonContains(_this.extendedFlyoutZone, [e.clientX, e.clientY])
+            ) {
+                _this.flyoutTimer = setTimeout(function() {
+                    _this.showFlyoutMenu($target);
+                }, 200);
+            } else {
+                // The extended flyout zone keeps shrinking when the mouse moves towards the target's flyout.
+                // Hence, if the mouse is moved and stopped over a new target, sometimes it could be slightly outside
+                // the extended flyout zone. This in turn will not trigger the flyoutTimer.
+                // Hence, the showFlyoutMenu should be manually called.
                 _this.showFlyoutMenu($target);
-            }, 200);
+            }
         }
 
         _this.extendedFlyoutZone[0] = [e.clientX, e.clientY];
--- a/public/js/icinga/behavior/navigation.js
+++ b/public/js/icinga/behavior/navigation.js
@@ -9,7 +9,8 @@
     try {
         var d3 = require("icinga/icinga-php-thirdparty/mbostock/d3");
     } catch (e) {
-        console.warn('[Navigation] requires d3.js library');
+        d3 = null;
+        console.warn('D3.js library is unavailable. Navigation to flyout may not work as expected.');
     }
 
     var Navigation = function (icinga) {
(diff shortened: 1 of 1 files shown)


public/js/icinga/behavior/navigation.js
```diff
@@ -282,63 +318,102 @@
...
         if ($layout.hasClass('minimal-layout')) {
             return;
         }
 
-        var $target = $(this);
-        var $flyout = $target.find('.nav-level-2');
-
+        let $target = $(this);
+        let $flyout = $target.find('.nav-level-2');
         if (! $flyout.length) {
```
nilmerg, line 334: This can be simplified to `! $target[0].matches(':has(.nav-level-2)')` which doesn't require a reference to the flyout itself.

public/js/icinga/behavior/navigation.js
```diff
@@ -6,14 +6,22 @@
...
         Icinga.EventListener.call(this, icinga);
         this.on('click', '#menu a', this.linkClicked, this);
         this.on('click', '#menu tr[href]', this.linkClicked, this);
         this.on('rendered', '#menu', this.onRendered, this);
-        this.on('mouseenter', '#menu .primary-nav .nav-level-1 > .nav-item', this.showFlyoutMenu, this);
+        this.on('mousemove', '#menu .primary-nav .nav-level-1 > .nav-item', this.onMouseMove, this);
+        this.on('mouseenter', '#menu .primary-nav .nav-level-1 > .nav-item', this.onMouseEnter, this);
         this.on('mouseleave', '#menu .primary-nav', this.hideFlyoutMenu, this);
         this.on('click', '#toggle-sidebar', this.toggleSidebar, this);
+        this.on('resize', '', this.toggleConfigFlyout, this);
```
nilmerg, line 24: ?

nilmerg (changes requested):
Please use always `const` instead of `let`, unless `let` is required.

public/js/icinga/behavior/navigation.js
raviks789: I think, I was trying to trigger an event on flyout resize and just left it as it was unnecessary.  This should be cleaned up.

flourish86 (approved):
Obviously the preview triangle should be removed in the end but as tested, this works really well now.

public/js/icinga/behavior/navigation.js
```diff
@@ -282,63 +314,96 @@
...
+     * @param $target
+     */
+    Navigation.prototype.showFlyoutMenu = function($target) {
+        const $flyout = $target.find('.nav-level-2');
+        if (! $target.is(':hover')) {
+            return;
+        }
+
+        $target.siblings().not($target).removeClass('hover');
+        $target.addClass('hover');
```
nilmerg, line 381: I've meant to move multiple lines before. Take a look at the noted range of lines in the screenshot. It's obviously bugged right now, since the diff is outdated, but it's clearly not just a single line what this comment is about!

image

Now take note of the range of **this** comment: +376 to +381

public/js/icinga/behavior/navigation.js
```diff
@@ -282,63 +314,96 @@
...
+        if (! $target[0].matches(':has(.nav-level-2)')) {
             $layout.removeClass('menu-hovered');
             $target.siblings().not($target).removeClass('hover');
             return;
         }
 
-        var delay = 300;
+        $layout.addClass('menu-hovered');
+        _this.coordinates[0] = [e.clientX, e.clientY];
+        _this.showFlyoutMenu($target, _this.coordinates[0]);
```
nilmerg, line 342: You don't need to pass the cursor position anymore

public/js/icinga/behavior/navigation.js
```diff
@@ -30,6 +36,32 @@
          */
         this.active = null;
 
+        /**
+         * The co-ordinates of the triangle formed from the previous cursor position
+         * and the left corners of the flyout
+         *
+         * @type {Array}
+         */
+        this.coordinates = new Array(3);
```
nilmerg, line 45: *coordinates* is not really a good name. Without looking at the doc block, it reads as *navigation coordinates*, which is obviously wrong. Please find an alternative. (Without the term *coordinates*)

public/js/icinga/behavior/navigation.js
```diff
@@ -30,6 +36,32 @@
...
+
+        this.flyoutTimer = null;
```
nilmerg, line 47: Add a doc block

public/js/icinga/behavior/navigation.js
```diff
@@ -6,15 +6,21 @@
 
     Icinga.Behaviors = Icinga.Behaviors || {};
 
+    try {
+        var d3 = require("icinga/icinga-php-thirdparty/mbostock/d3");
+    } catch (e) {
+        console.warn('[Navigation] requires d3.js library');
```
nilmerg, line 12: If you only *warn* here, make sure the remaining code doesn't break in case `d3` is undefined

public/js/icinga/behavior/navigation.js
```diff
@@ -282,63 +314,96 @@
...
         }
 
-        setTimeout(function() {
-            try {
-                if (! $target.is(':hover')) {
-                    return;
-                }
-            } catch(e) { /* Bypass because if IE8 */ }
+        if ($target[0].matches(':has(.nav-level-2)')) {
+            _this.flyoutTimer = setTimeout(function() {
```
nilmerg, line 362: This runs out in case movement stops while still hovering the same target. Don't start the timeout in case the shown flyout is the one of the target!

nilmerg (changes requested):
And remove the visible triangle now please

public/js/icinga/behavior/navigation.js
raviks789: I see, but I think the removal of the hover class from the sibling navigation item and adding that class to current target needs to be done in `showFlyoutMenu` method. Orelse I have to do that in onMouseMove method again. That is why I misunderstood the comment. But the lines from +376 to +379 could be moved to onMouseEnter.

public/js/icinga/behavior/navigation.js
```diff
@@ -6,15 +6,25 @@
 
     Icinga.Behaviors = Icinga.Behaviors || {};
 
+    try {
+        var d3 = require("icinga/icinga-php-thirdparty/mbostock/d3");
+    } catch (e) {
+        d3 = null;
```
nilmerg, line 12: you don't need that, use `typeof d3 !== "undefined"` instead

public/js/icinga/behavior/navigation.js
```diff
@@ -282,63 +307,101 @@
...
+        const _this = e.data.self;
+        clearTimeout(_this.flyoutTimer);
+
+        const $target = $(this);
+
+        if (
+            $target[0].matches(':has(.nav-level-2)')
+            && ! $target.hasClass('hover')
+            && ! _this.extendedFlyoutZone.includes(undefined)
+            && d3.polygonContains(_this.extendedFlyoutZone, [e.clientX, e.clientY])
```
nilmerg, line 362: Something is off here. For me, the flyout doesn't disappear after 200ms even though I'm clearly not over the original target anymore. I'll show in person how this behaves.

nilmerg (changes requested):

nilmerg (approved):
It even works perfect in the collapsed menu :tada:

PR #5372, opened 2025-05-26
Fix typo in getProgressLabel()

jrauh01:
The function was named `getProgressLabe()`.


nilmerg:
I'm not sure if I want to introduce a semi-breaking change like this on such a legacy part of the code, which is to be replaced anyway… It just results in more hassle than necessary.

PR #5375, opened 2025-06-04
menu.less: Apply same `bg-color` to the caret in the `.nav-item-header`

sukhwinder33445:
## Before:
Screenshot 2025-06-04 at 09 03 52

## After:
Screenshot 2025-06-04 at 09 08 01

### DIFF
--- a/public/css/icinga/menu.less
+++ b/public/css/icinga/menu.less
@@ -534,25 +534,31 @@ html.no-js  #toggle-sidebar {
   margin-right: 0;
 }
 
-#layout.sidebar-collapsed #menu .nav-level-1 > .nav-item.hover .nav-level-2 > .nav-item-header {
-  background-color: @menu-bg-color;
-  color: @menu-color;
-  border-bottom: 1px solid @gray-light;
-  border-top-left-radius: .25em;
-  border-top-right-radius: .25em;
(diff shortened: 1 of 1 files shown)


public/css/icinga/menu.less
```diff
@@ -532,24 +532,30 @@ html.no-js  #toggle-sidebar {
...
-  border-top-left-radius: .25em;
-  border-top-right-radius: .25em;
-
-  span {
-    padding-left: 1.375em;
-    padding-right: 0.545em;
-    height: @nav-item-height;
-    line-height: @nav-item-height;
-    display: block;
+#layout:not(.minimal-layout).sidebar-collapsed #menu .nav-level-1 > .nav-item.hover > .nav-level-2 {
```
nilmerg, line 537: Why `:not(.mimimal-layout)`?

sukhwinder33445: Otherwise, the color of the caret (:after) cannot be changed, as it has been defined with higher specificity (:not.minimal-layout). I have used the [same](https://github.com/Icinga/icingaweb2/blob/fe6d568345a39dfcce782232f8851549b6d8aa27/public/css/icinga/menu.less#L261) specificity to change the color.
And the caret and the `.nav-item-header` are not present in the `.minimal-layout`. So this change cannot have a negative effect.

nilmerg (approved):

PR #5379, opened 2025-06-12
base.less: Define scrollbar color

### LINKED ISSUE #5378: Dark mode shows light mode scrollbar
## Describe the bug
The scroll bar in dark mode looks exactly the same as in light mode.

1. Have a look at a page that contains a scroll bar

## Screenshot
Image

## Expected behavior

Image

sukhwinder33445:
fixes #5378 

requires: https://github.com/Icinga/ipl-web/pull/296

This does not work for safari (with webkit css).

### DIFF
--- a/public/css/icinga/base.less
+++ b/public/css/icinga/base.less
@@ -193,7 +193,10 @@
 // Make padding not affect the final computed width of an element
 html {
   box-sizing: border-box;
+
+  scrollbar-color: var(--scrollbar-thumb-color, @scrollbar-thumb-color) transparent;
 }
+
 details > * {
   // children somehow default to content-box no matter the inheritance
   box-sizing: border-box;


lippserd:
@flourish86 Is this the correct approach?

flourish86:
Sorry, maybe I'm missing context, but I think this just how chrome behaves and browser or even OS UI should be altered, imho. This might not be sustainable, because vendors might change that randomly.

Additionally to be noted is that scroll bars on macOS appear only, when you scroll as set by default. The circumstances, this affects user are therefore very narrow, I'd say.

So I don't see a point in changing this in the first place.

If you insist in changing it anyway, have you tried using `scrollbar-color: auto`?

See https://caniuse.com/mdn-css_properties_scrollbar-color_auto

Besides that: A popular GPT tool confirmed, that `scrollbar-color` is the best way to change the scrollbar color via css, but it might not get displayed properly and consistently in every browser.

Side note: It works perfectly fine in Safari and Firefox without the fix.

## FF
Screenshot 2026-03-24 at 16 14 48

## Safari
Screenshot 2026-03-24 at 16 15 09

sukhwinder33445:
The scrollbar works correctly in Safari and other browsers without any changes now. Only Chrome shows a white scrollbar (in dark mode), and unfortunately `scrollbar-color: auto` does not fix it, it is also the default value anyway.
Closing this PR since introducing a Chrome-only workaround is not appropriate.

PR #5381, opened 2025-06-17
Fix mobile/minimal layout layout

### LINKED ISSUE #5023: Header scrollable on mobile view
## Describe the bug
The header section on the mobile view of icingaweb2 is scrollable, so that you lose (direct) access to the menu

## To Reproduce
Open Icingaweb on a mobile device (or emulate one using developer tools) and try to scroll in the header section.

## Expected behavior
That section should not be scrollable, so that the menu (and logo) is visible all the time.

## Screenshots
HeaderScrollable

## Your Environment
* Icinga Web 2 version and modules (System - About): 2.11.4
* Web browser used: Chrome 112.0.5615.121
* Icinga 2 version used (`icinga2 --version`): r2.13.7-1
* PHP version used (`php --version`): 8.0.27
* Server operating system and version: 

## Possible Solution
adding "overflow-x: hidden" and "overflow-y: hidden" fixed the issue (only tested briefly)

#layout.minimal-layout #sidebar {
    width: 100%;
    overflow: auto;
    overflow-x: hidden;
    overflow-y: hidden;
}

### LINKED ISSUE #5386: Improve Minimal layout
#### The sidebar has a scrollbar, because of `.skip-links`.

Image

---

#### The extended sidebar does not have the full height.

Image

---

#### The minimal layout does not support multiple select, but the message is displayed.

Image

sukhwinder33445:
fix #5386, fix #5023

### DIFF
--- a/public/css/icinga/responsive.less
+++ b/public/css/icinga/responsive.less
@@ -73,6 +73,7 @@
 
     &.expanded {
       overflow: auto;
+      height: 100%;
     }
   }
 


flourish86 (commented):
✅ The header scrolling problem seems to be fixed.

✅ I noticed that the expanded menu takes now full height, which is a good side effect.

⚠️ The message in the list footer mentioned in #5386 is still there. We should either fix this here or keep track of it in another issue:

I also noticed weird hovering behavior in Safari with the menu. It won't happen in the Chrome inspector on Desktop, because it has hover effects disabled. It should though be tested on Android Chrome.

https://github.com/user-attachments/assets/0086fb71-ed44-4b72-b94b-ab309f2ba0ca

If the problem is Safari desktop only, we can ignore it, since it's not a very common use case. But on the other hand, it shouldn't be difficult to disable the opening of the '.level-2' submenu on hover.

flourish86:
> I also noticed weird hovering behavior in Safari with the menu. It won't happen in the Chrome inspector on Desktop, because it has hover effects disabled. It should though be tested on Android Chrome.
> 
>  Screen.Recording.2025-11-10.at.09.09.04.mov 
> 
> If the problem is Safari desktop only, we can ignore it, since it's not a very common use case. But on the other hand, it shouldn't be difficult to disable the opening of the '.level-2' submenu on hover.

We just investigated this and tested it on an actual phone. The Chrome inspector disables the hover behavior completely, so this is not reproducible there (Firefox also has an option to enable "touch mode"). iOS Safari still recoginzes the hover effect regardlessly.

Besides that, the level-2 menus still appear on active/selected nav-items. We should completely disable the level-2 menus in .minimal mode to make sure the we don't run into any weird side-effects, so the navigation works via the dashboard sheets as intended.

@sukhwinder: I'd suggest we track this in a different issue.

nilmerg:
please rebase

nilmerg:
@cla-bot check

PR #5387, opened 2025-07-08
Streamline authentication process

### LINKED ISSUE #5265: CustomUserBackend is loaded too late
# History
In the meantime, some time has passed as @nilmerg explained me how to create a own custom backend. The requirement was to call Icinga from an application portal (beside the possability via DbBackend) and use its authentication information for the login. Because it's a kind of external backend I should use as role model https://github.com/Icinga/icingaweb2/blob/c4b6e4bdda691820fd3c16841436b2ef796f44b1/library/Icinga/Authentication/User/ExternalBackend.php#L13. 
Now I have finally managed to do this, but I noticed a few things that I would like to describe here

## Describe the bug
During the development I recognized that my module with the own backend is loaded too late and/or the authentifcation call was after putting the credentials into the LoginForm. 
Depending on the situation icingaweb produces a lot of error messages like:
`Can't create authentication backend "portal". An exception was thrown: <- Icinga\Exception\ConfigurationError in /usr/share/php/Icinga/Authentication/User/UserBackend.php with message: Authentication configuration for user backend "xxx" defines an invalid backend type. Backend type "xxx" is not supported
`
It doesn't matter in which order I configured my backend and the DbBackend at `/etc/icingaweb2/authentication.ini`. Because of that I had to debug the complete login sequence.  

## To Reproduce
The first problem (or actually hurdle) I encourtered is the function https://github.com/Icinga/icingaweb2/blob/c4b6e4bdda691820fd3c16841436b2ef796f44b1/library/Icinga/Authentication/Auth.php#L239 which is triggerd by https://github.com/Icinga/icingaweb2/blob/c4b6e4bdda691820fd3c16841436b2ef796f44b1/library/Icinga/Authentication/Auth.php#L88. The problem what I saw is here, that the function authExternal checks if the backend is a instace of the class `ExternalBackend`. So If I create my own backend with the implemation of the Interface `UserBackendInterface`. The function returns false everytime and consequently the authentication fails. Here is the workarround for now that the own backend extends the class `ExternalBackend` although the most things isn't needed from that class. Maybe to create an own interface, which also will be checked in `authExternal` instead of the class name? Just to produce nicer code.

The second problem I encourtered is this call https://github.com/Icinga/icingaweb2/blob/c4b6e4bdda691820fd3c16841436b2ef796f44b1/library/Icinga/Authentication/User/UserBackend.php#L91. All installed and activated modules are loaded alphabetically. For example I choose as name for my module "Mybackend". If there are many other modules loaded before like "audit", "icingadb" etc. it takes time, until the own module is loaded.  And here happens a big problem: During loading every module and register possible custom backends, following function gets triggerd https://github.com/Icinga/icingaweb2/blob/c4b6e4bdda691820fd3c16841436b2ef796f44b1/library/Icinga/Application/Modules/Module.php#L1374  This function includes the configuration.php and excutes the existing PHP code. If there is a code like (from icingadb-web):
```
$auth = Auth::getInstance();
$authenticated = Icinga::app()->isWeb() && $auth->isAuthenticated();
```
the call will fail, because the own modul isn't loaded and the own backend isn't registred yet. This produces the mentioned error message above and happens as long as, until the own module gets loaded! Depending on the number of modules installed, the log file becomes filled with error messages!
The only workarround for this is to rename the own module to a little senseless name like "aaamybackend" to provoke the fact that the own module with the own backend is loaded first.

## Expected behavior
I know this is a still rare case. Maybe not everybody is crazy like my/us 😆 to create a own backend. So get a solution for this will be low prio I think. For loading in a specific order instead of alphabetical could be to set something like a load order. So that, the call
`Icinga::app()->getModuleManager()->getLoadedModules()` returns the correct and needed order. 
And for the "problem" with `authExternal` a solution could be an own interface as I wrote before. But that's only that the code will be nicer, not more. 
I just wanted to draw attention to it.

## Your Environment

* Web browser used: Edge (128.0.2739.79) / Firefox (115.15.0esr)
* Icinga 2 version used (`icinga2 --version`):  2.14.2-1
* PHP version used (`php --version`): 8.0.30
* Server operating system and version: RHEL  9.4

### LINKED ISSUE #5385: Users who are not allowed to change the theme, cannot change the theme mode either
## Describe the bug
A user who has only the theme mode selection available, can choose whatever they like but storing it has no effect.

## To Reproduce

1. Go the general application configuration
1. Prohibit changing the theme, globally
3. Choose the Icinga default theme as default
2. Save
4. Go to the preferences
5. Switch to another theme mode
6. Save
7. Notice that the mode change has no effect

## Expected behavior
The user should be able to change the theme mode.

## Your Environment

* Icinga Web 2 version and modules (System - About): latest

## Additional context
This is related to #5265, somewhat. Here, also the businessprocess module is the culprit. Enabling it, solves this issue. Having it disabled, authentication has not been performed at the stage the theme mode is evaluated, **while theme selection has been disabled**. (Which, if enabled, would also trigger authentication in time.)

nilmerg:
This is split into four distinct parts: (One commit is only performing a clean up)

### 1. EmbeddedWeb: Explicitly perform authentication

It is nowadays no exception that stylesheet may be dependent on who's using the app. So to avoid race conditions like in https://github.com/Icinga/icingaweb2/issues/5385 authentication is an explicit step during bootstrap now.
`EmbeddedWeb` now knows about and is responsible to initially set up the user object. So this does not change when `Web` sets up the user, it just allows `EmbeddedWeb` to do it.

A small side-effect introduced by this change is fully optional, but doesn't break anything either: Method `Web::setupSession()` and the accompanying `Web::$session` property are gone now as the property was obsolete already. Initializing the session is still done at the previous stage though, but implicitly by `Web::setupNotifications()`.

### 2. Auth: Perform authentication only once and not lazily

Since authentication is now performed even for static resources, there's no reason anymore to support implicit authentication. This also limits authentication attempts to a single one, previously failed attempts were repeated. Requiring authentication during bootstrapping, i.e. before authentication has been performed, will now throw an error.

This conflicts with #5430 as it also affects `Auth::isAuthenticated()`. But this should be easy to resolve.

### 3. Manager: Perform module loading asynchronously

So that authentication can suspend it. There are cases, e.g. cube, where authentication is required in run.php. During bootstrapping loading modules is mostly required to load libraries, register routes and hooks. Most of the time authentication is not required for these, but if it is, evaluation is now interrupted and continued after authentication has actually been performed.

I don't see a real risk for any breaking change here, since authentication happens shortly after. It actually avoids a breaking change, since without this, cube's Icinga DB support would break or at least malfunction.

And cube is only a single example. As long as the *monitoring* module is still supported by them, they will keep being affected.

Enable the *monitoring* module and Icinga DB Web in an environment and look at the debug log to see this change in effect.

### 4. Module: Expect user and group backends upon registration

Providing a user or user group backend in configuration.php now triggers a deprecation or warning. They are expected to be announced in run.php, just like hooks. Authentication attempts during a configuration.php run, while Authentication has not been performed yet, will be also suspended now. This ensures that the current behavior, being broken or working, stays the same.

You might think that this is a breaking change, but it was rather always broken and only worked in such particular situations that custom backends were never used in conjunction with any of the official modules mentioned in 3.

Removing the compatibility layer (the suspension part) with v2.15 might indeed be a bit hasty. But I rather like to get rid of this as early as possible and I think those who have custom extensions like this, in a functional state, are a minority and known to us. (i.e. Icinga partners which we can forewarn)

### DIFF
--- a/doc/80-Upgrading.md
+++ b/doc/80-Upgrading.md
@@ -5,6 +5,12 @@ v2.6 to v2.8 requires to follow the instructions for v2.7 too.
 
 ## Upgrading to Icinga Web 2.14
 
+**Deprecations**
+
+* Providing a custom user or group backend inside a module's configuration.php is deprecated and won't work as of v2.15.
+  * It continues to work under the same circumstances as before and remains broken where it was already broken.
+  * To keep it working as of v2.15 or make it work now, provide the custom backend inside the module's run.php instead.
+
 **Framework changes affecting third-party code**
 
 * Our JavaScript implementation to update relative times in the UI has been removed from Icinga Web, and an updated
--- a/doc/80-Upgrading.md
+++ b/doc/80-Upgrading.md
@@ -5,6 +5,12 @@ v2.6 to v2.8 requires to follow the instructions for v2.7 too.
 
 ## Upgrading to Icinga Web 2.14
 
+**Deprecations**
+
+* Providing a custom user or group backend inside a module's configuration.php is deprecated and won't work as of v2.15.
+  * It will still work as previously under the right circumstances, but will still be broken if it broke beforehand.
+  * To keep it working as of v2.15 or make it work now, provide the custom backend inside the module's run.php instead.
+
 **Framework changes affecting third-party code**
 
 * Our JavaScript implementation to update relative times in the UI has been removed from Icinga Web, and an updated
--- a/library/Icinga/Application/Modules/Manager.php
+++ b/library/Icinga/Application/Modules/Manager.php
@@ -5,6 +5,7 @@
 
 namespace Icinga\Application\Modules;
 
+use Fiber;
 use Icinga\Application\ApplicationBootstrap;
 use Icinga\Application\Icinga;
 use Icinga\Application\Logger;
@@ -189,8 +190,23 @@ class Manager
     public function loadEnabledModules()
     {
         if (! $this->loadedAllEnabledModules) {
+            $async = ! $this->app->isCli();
             foreach ($this->listEnabledModules() as $name) {
-                $this->loadModule($name);
+                if ($async) {
+                    // May be suspended during authentication and resumed upon finish
+                    (new Fiber(function () use ($name) {
+                        Logger::debug(
+                            'Loading enabled module "%s" asynchronously (Process %d; Fiber %d)',
+                            $name,
+                            getmypid() ?: 0,
+                            spl_object_id(Fiber::getCurrent())
+                        );
+
+                        $this->loadModule($name);
+                    }))->start();
+                } else {
+                    $this->loadModule($name);
+                }
             }
 
             $this->loadedAllEnabledModules = true;
--- a/library/Icinga/Application/Modules/Module.php
+++ b/library/Icinga/Application/Modules/Module.php
@@ -6,6 +6,7 @@
 namespace Icinga\Application\Modules;
 
 use Exception;
+use Fiber;
 use Icinga\Application\ApplicationBootstrap;
 use Icinga\Application\Config;
 use Icinga\Application\Hook;
@@ -1084,7 +1085,22 @@ class Module
      */
     public function getUserBackends()
     {
-        $this->launchConfigScript();
+        // TODO: Remove with v2.15
+        if (Fiber::getCurrent() === null) {
+            (new Fiber(function () {
+                Logger::debug(
+                    'Running config script of module "%s" asynchronously (Process %d; Fiber %d)',
+                    $this->getName(),
+                    getmypid() ?: 0,
+                    spl_object_id(Fiber::getCurrent())
+                );
+
+                $this->launchConfigScript();
+            }))->start();
+        } else {
+            $this->launchConfigScript();
+        }
+
         return $this->userBackends;
     }
 
@@ -1095,7 +1111,22 @@ class Module
      */
     public function getUserGroupBackends()
     {
-        $this->launchConfigScript();
+        // TODO: Remove with v2.15
+        if (Fiber::getCurrent() === null) {
+            (new Fiber(function () {
+                Logger::debug(
+                    'Running config script of module "%s" asynchronously (Process %d; Fiber %d)',
+                    $this->getName(),
+                    getmypid() ?: 0,
+                    spl_object_id(Fiber::getCurrent())
+                );
+
+                $this->launchConfigScript();
+            }))->start();
+        } else {
+            $this->launchConfigScript();
+        }
+
         return $this->userGroupBackends;
     }
 
@@ -1196,6 +1227,21 @@ class Module
      */
     protected function provideUserBackend($identifier, $className)
     {
+        if ($this->registered) {
+            trigger_error(sprintf(
+                'Module "%s" already registered. Providing user backend "%s" in configuration.php'
+                . ' has no effect as of version 2.15. Put it in run.php instead.',
+                $this->getName(),
+                $identifier
+            ), E_USER_DEPRECATED);
+            Logger::warning(
+                'Module "%s" already registered. Providing user backend "%s" in configuration.php'
+                . ' has no effect as of version 2.15. Put it in run.php instead.',
+                $this->getName(),
+                $identifier
+            );
+        }
+
         $this->userBackends[strtolower($identifier)] = $className;
         return $this;
     }
@@ -1210,6 +1256,21 @@ class Module
      */
     protected function provideUserGroupBackend($identifier, $className)
     {
+        if ($this->registered) {
+            trigger_error(sprintf(
+                'Module "%s" already registered. Providing user group backend "%s" in configuration.php'
+                . ' has no effect as of version 2.15. Put it in run.php instead.',
+                $this->getName(),
+                $identifier
+            ), E_USER_DEPRECATED);
+            Logger::warning(
+                'Module "%s" already registered. Providing user group backend "%s" in configuration.php'
+                . ' has no effect as of version 2.15. Put it in run.php instead.',
+                $this->getName(),
+                $identifier
+            );
+        }
+
         $this->userGroupBackends[strtolower($identifier)] = $className;
         return $this;
     }
--- a/library/Icinga/Application/Modules/Module.php
+++ b/library/Icinga/Application/Modules/Module.php
@@ -20,7 +20,6 @@ use Icinga\Web\Widget;
 use ipl\I18n\GettextTranslator;
 use ipl\I18n\StaticTranslator;
 use ipl\I18n\Translation;
-use Zend_Controller_Router_Route;
 use Zend_Controller_Router_Route_Abstract;
 use Zend_Controller_Router_Route_Regex;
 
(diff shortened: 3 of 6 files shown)


library/Icinga/Authentication/Auth.php
```diff
@@ -90,14 +110,26 @@ public function getAuthChain()
...
+                    spl_object_id($fiber),
+                    getmypid() ?: 0
+                );
+
+                $this->suspendedFibers[] = $fiber;
+                Fiber::suspend();
+            } else {
+                throw new RuntimeException(
+                    'This should not happen. Please report this as a bug.'
+                    . 'Include the full stack trace in your report and a list of *all* installed modules.'
```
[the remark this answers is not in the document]
nilmerg: done

library/Icinga/Application/Modules/Module.php
```diff
@@ -1085,7 +1086,22 @@ public function getSetupWizard()
...
+                    spl_object_id(Fiber::getCurrent())
+                );
+
+                $this->launchConfigScript();
+            }))->start();
+        } else {
+            $this->launchConfigScript();
+        }
+
         return $this->userBackends;
```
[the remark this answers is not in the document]
nilmerg: This is essentially why I stated:

> This ensures that the current behavior, being broken or working, stays the same.

If this case occurs, it means user group backends never worked. We just ensure it doesn't break, as this never broke either.

library/Icinga/Application/Modules/Module.php
```diff
@@ -1096,7 +1112,22 @@ public function getUserBackends()
...
+                );
+
+                $this->launchConfigScript();
+            }))->start();
+        } else {
+            $this->launchConfigScript();
+        }
+
         return $this->userGroupBackends;
     }
```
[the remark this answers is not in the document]
nilmerg: same reasoning as for getUserBackends.

library/Icinga/Application/Web.php
```diff
@@ -18,14 +18,11 @@
 use Zend_Paginator;
```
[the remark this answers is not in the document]
nilmerg: done, also affected `Module.php`.

library/Icinga/Authentication/Auth.php
```diff
@@ -90,14 +109,26 @@ public function getAuthChain()
...
+                );
+
+                $this->suspendedFibers[] = $fiber;
+                Fiber::suspend();
+            } else {
+                throw new LogicException(
+                    'Authentication has not been performed yet.'
+                    . ' Call Auth::authenticate() during bootstrap before calling isAuthenticated().'
+                );
+            }
```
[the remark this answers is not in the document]
nilmerg: fixed

library/Icinga/Authentication/Auth.php
```diff
@@ -216,6 +247,51 @@ public function setUser(User $user)
...
+                'Authentication performed, resuming fiber %d of process %d.',
+                spl_object_id($fiber),
+                getmypid() ?: 0
+            );
+
+            $fiber->resume();
+        }
+
+        $this->suspendedFibers = [];
+
```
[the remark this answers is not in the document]
nilmerg: No idea how stateful PHP behaves, but unless we support this, this isn't an issue. Besides many many other parts…

library/Icinga/Authentication/Auth.php
lippserd: `setAuthenticated()` never sets `$this->authenticated = true`. This is inconsistent with `setUser()`, which does set the flag.
Currently, `setAuthenticated()` is only called from `authExternal()` and `authHttp()`, which both run inside `authenticate()` so the flag ends up correct regardless.
But `setAuthenticated()` is public, and any caller that invokes it directly (unit tests, external modules, future contributors) will leave `$authenticated = false`, causing the next `isAuthenticated()` call to throw `LogicException`.

```suggestion
        $this->setUser($user);
```

library/Icinga/Authentication/Auth.php
```diff
@@ -52,6 +54,23 @@ class Auth
...
     private $user;
 
+    /**
+     * Whether the user has been authenticated
+     *
+     * If true, this does NOT mean authentication was successful.
+     *
+     * @var bool
+     */
+    private bool $authenticated = false;
```
lippserd: This naming creates ambiguity and contradicts `isAuthenticated()` and `setAuthenticated()`: `$authenticated` actually is the outcome, but here it's the process gate. Please rename to something like `$authPerformed` . That would also make the comment unnecessary.

doc/80-Upgrading.md
```diff
@@ -5,6 +5,12 @@ v2.6 to v2.8 requires to follow the instructions for v2.7 too.
 
 ## Upgrading to Icinga Web 2.14
 
+**Deprecations**
+
+* Providing a custom user or group backend inside a module's configuration.php is deprecated and won't work as of v2.15.
+  * It will still work as previously under the right circumstances, but will still be broken if it broke beforehand.
```
lippserd:
```suggestion
  * It continues to work under the same circumstances as before, and remains broken where it was already broken.
```

library/Icinga/Application/Modules/Manager.php
```diff
@@ -189,8 +190,23 @@ private function detectEnabledModules()
...
+            $async = ! $this->app->isCli();
             foreach ($this->listEnabledModules() as $name) {
-                $this->loadModule($name);
+                if ($async) {
+                    // May be suspended during authentication and resumed upon finish
+                    (new Fiber(function () use ($name) {
+                        Logger::debug(
+                            'Loading enabled module "%s" asynchronously (Process %d; Fiber %d)',
+                            $name,
+                            getmypid() ?: 0,
```
lippserd, line 201: Any specific reason to use `getmypid()`? It won't change during a request and we already have `spl_object_id` for identification.

lippserd (changes requested):
Just some minor adjustments. Well done overall! 👍 

I tested the PR with multiple mocked modules and couldn't find any flaws.

The only thing we have to keep in mind or document is that hooks that should run during authentication (and the new two factor hook) must be registered before any call to `isAuthenticated()` is performed.

library/Icinga/Application/Modules/Manager.php
nilmerg: Parallel requests by different sessions have different pids.

nilmerg: ....but not necessarily different fiber ids.

library/Icinga/Authentication/Auth.php
nilmerg: done

library/Icinga/Authentication/Auth.php
nilmerg: done

doc/80-Upgrading.md
nilmerg: done, minus a comma

library/Icinga/Application/Modules/Manager.php
lippserd: Makes sense. Given that we do not have something like a request ID, let's keep it as is.

lippserd (approved):

PR #5400, opened 2025-08-06
Add opacity on hover

### LINKED ISSUE #5399: Add hover effect on big donut label
## Problem
The big donut label does not have a hover effect.

Add opacity on hover.

Image

Image

jrauh01:
Add opacity on the big donut label if hovered.

Closes #5399


flourish86 (approved):
Looking good! ✨

PR #5401, opened 2025-08-06
forms.less: Apply fixed height to input elements

sukhwinder33445:
ref https://github.com/Icinga/ipl-web/pull/305

### DIFF
--- a/public/css/icinga/forms.less
+++ b/public/css/icinga/forms.less
@@ -228,11 +228,20 @@ form.icinga-form {
   }
 }
 
-form.icinga-form select:not([multiple]),
-form.icinga-form input[type="datetime-local"] {
-  // Compensate inconsistent select height calculations
-  line-height: 1em;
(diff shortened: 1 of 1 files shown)


public/css/icinga/forms.less
```diff
@@ -228,11 +228,16 @@ form.icinga-form {
   }
 }
 
-form.icinga-form select:not([multiple]),
-form.icinga-form input[type="datetime-local"] {
-  // Compensate inconsistent select height calculations
```
nilmerg, line 233: why is this comment gone?

nilmerg (changes requested):

public/css/icinga/forms.less
sukhwinder33445: This is because we no longer compensate for this, as all input elements that we support are also added to the rule. The rest has its own style, e.g. checkbox, button.

PR #5402, opened 2025-08-07
Remove whitespace around config which couldn't be stored

Al2Klimov:
If I change settings in the web UI, but latter can't write to /etc/icingaweb2, it shows me the desired file contents, so I can deploy them myself – with strange whitespace around it. Screenshots say it all:

## Before

Bildschirmfoto 2025-08-07 um 11 20 31

## After

Bildschirmfoto 2025-08-07 um 11 19 44


flourish86 (approved):
Didn't check it in the browser. But from the screenshots this looks reasonable.

Issue #4945, opened 2022-11-21
Open Redirects for logged in users


mburring:
## Describe the bug
When logged in the underlying URL redirection system appears to be open and will will redirect to any requested URL

## To Reproduce
1) https://my.icingaweb2.server/icingaweb2/authentication/login?redirect=http://google.com
Will issue a 302 redirect to google.com

![icingawe2redirect](https://user-images.githubusercontent.com/118503693/202985134-d5df151b-30fc-49a7-8910-fcb8a456a30e.png)

## Expected behavior
All open redirects should be denied. While not a direct exploit it can be used as a potential attack surface for misdirection

## Your Environment
Include as many relevant details about the environment you experienced the problem in

* Icinga Web 2 version and modules (System - About): 2.11.2
* Web browser used: Chrome, Firefox
* Icinga 2 version used (`icinga2 --version`): r2.13.6-1
* PHP version used (`php --version`): 7.4.3
* Server operating system and version: Ubuntu 20.04

PR #5403, opened 2025-08-08
Form: don't redirect to external resources

Al2Klimov:
by throwing SecurityException if Form#setRedirectUrl() is called with an external URL. This prevents open redirects, like #4952. For the rationale, see #4945.


Al2Klimov:
Damn. E.g. SSO forms MUST redirect the user externally. 🙈

Edit: but they could override the setter.

nilmerg:
I already figured something like this. There was a reason why the LoginForm prevents it on its own, because the redirect is influenced by a GET parameter. A prevention like this by default collides with valid cases like OAuth. An intended external redirect should be possible without needing to override anything.

PR #5404, opened 2025-08-11
Use NotWritableError for ini config

jrauh01:
Replaced the unused `Zend_Config_Exception` with a more appropriate `NotWritableError` in `IniWriter::write()`.

- The old `Zend_Config_Exception` was never caught anywhere in the codebase.
- Changed to `NotWritableError` to better reflect file permission and write issues.
- Added `@` suppression to the `file_put_contents()` call so that the method can throw `NotWritableError` instead of crashing on PHP warnings.


nilmerg:
Please add a note to the upgrading documentation

PR #5408, opened 2025-08-14
Declare a redirect to __REFRESH__ as autorefresh

Al2Klimov:
to preserve scroll position when a modal is closed.

* E.g https://github.com/Icinga/netrp/issues/150

If this is explicitly unwanted in a particular case(?), we can add a `__RELOAD__` with the old behavior.


nilmerg:
The related PR is closed, so I don't think this needs consideration right now.

PR #5412, opened 2025-08-20
Remove `text-overflow: ellipsis` for user-nav-item link

### LINKED ISSUE #5368: Improve title and label of user menu item in collapsed mode
## Describe the bug

Image

1. in collapsed mode the label is ellipsisized, which is unnecessary
2. The title reads "Iicingaadmin": it seems to include the Character from the icon

## Expected behavior
1. The label should be hidden
2. the title tooltip should show only the username without the initial character from the icon

jrauh01:
... if sidebar is collapsed. The title was automatically generated by `text-overflow: ellipsis` and only in Safari browser.

fixes #5368

### DIFF
--- a/public/css/icinga/configmenu.less
+++ b/public/css/icinga/configmenu.less
@@ -4,6 +4,10 @@
 
 .sidebar-collapsed #menu {
   margin-bottom: 8em;
+
+  .config-menu > ul .user-nav-item > a {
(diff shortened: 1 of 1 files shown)


flourish86 (approved):
Looks perfect to me now. Also the title is now gone.

PR #5418, opened 2025-08-25
Fix: add login hook to ext backends

### LINKED ISSUE #5415: Authentication hooks are not called with external backends
## Describe the bug
The authentication hooks (onLogin) are not called when the authentication is performed by an external backend.

## To Reproduce

1. Setup an external backend
2. Add a module that attach an onLogin hook
3. Do the login
4. The attached hook function is never called 

## Expected behavior
The hook function must be called also when the authentication is done from an external backend.

## Your Environment

* Icinga Web 2 version and modules (System - About): 2.12.4
* PHP version used (`php --version`): 7.4.33

gianlucapiccolo:
Fixes #5415

### DIFF
--- a/library/Icinga/Authentication/Auth.php
+++ b/library/Icinga/Authentication/Auth.php
@@ -6,6 +6,7 @@ namespace Icinga\Authentication;
 use Exception;
 use Icinga\Application\Config;
 use Icinga\Application\Hook\AuditHook;
+use Icinga\Application\Hook\AuthenticationHook;
 use Icinga\Application\Icinga;
 use Icinga\Application\Logger;
 use Icinga\Authentication\User\ExternalBackend;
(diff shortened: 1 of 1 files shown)


library/Icinga/Authentication/Auth.php
```diff
@@ -245,7 +246,8 @@ protected function authExternal()
                     if (! $user->hasDomain()) {
                         $user->setDomain(Config::app()->get('authentication', 'default_domain'));
                     }
-                    $this->setAuthenticated($user);
+		    $this->setAuthenticated($user);
+		    AuthenticationHook::triggerLogin($user);
```
lippserd: Can you please correct the indentation and add a new line before the `return`?

lippserd (changes requested):

library/Icinga/Authentication/Auth.php
lippserd: Thanks!

lippserd: Ah wait, newline before return is still missing.

PR #5420, opened 2025-08-28
Revert "Sometimes you just herp the derp so hard it herpderps"

Al2Klimov:
This reverts commit c83659eff78438f1450f074322ffdaf2a852571b to allow logged in users accessing $requiresAuthentication=false controllers even without the module/MOD_NAME permission.


Al2Klimov:
This PR is required for front-channel logout, unless you

* allow me to silently inject the module/sso-example permission during SSO
* or require the admin to give the module/sso-example permission to all users

Issue #5428, opened 2025-09-30
Remove button consistency


martialblog:
Hi,

There are some inconsistencies with the remove buttons/confirmation pages in the "Access Control" and "Dashboard Settings".

First, the buttons to remove elements are different. X in the "Access Control" and a trash bin icon in the "Dashboard Settings".

<img width="691" height="83" alt="Image" src="https://github.com/user-attachments/assets/14570f0c-9986-4193-ad59-81278a589c93" />
<img width="687" height="98" alt="Image" src="https://github.com/user-attachments/assets/7df4bc50-6d26-42c9-8534-7fe62a7bf3fc" />

Second, the confirmation pages after clicking the buttons are very inconsistent:

<img width="477" height="267" alt="Image" src="https://github.com/user-attachments/assets/e1b8427e-87b2-49bf-9493-e10bf0c28ff6" />
<img width="665" height="270" alt="agreesImage" src="https://github.com/user-attachments/assets/42145ee8-b4b7-45de-9fcd-fd4f22d0c32d" />
<img width="654" height="192" alt="Image" src="https://github.com/user-attachments/assets/0e24da4a-77f0-400f-a9a9-0c03a6436a3e" />
<img width="666" height="257" alt="Image" src="https://github.com/user-attachments/assets/74ddc9c1-c6ce-4809-a828-853fa25cb6da" />

A very minor thing I know, but I bet @flourish86 has something so say about it.

martialblog:
Other occurrences

<img width="807" height="129" alt="Image" src="https://github.com/user-attachments/assets/26f282b9-843c-46b2-9937-67e386304cc8" />

Also, the X is not aligned in the Backends list:

<img width="389" height="299" alt="Image" src="https://github.com/user-attachments/assets/961390a1-37a4-4726-9df0-f590b0476176" />


<img width="873" height="235" alt="Image" src="https://github.com/user-attachments/assets/c522293f-9ba8-433b-b426-7af01a678052" />

grupperino:
I would like to work on this issue if its possible :) I am a first time contributor and this seems like something well within in my capabilities...

flourish86:
Thanks for your offer, @grupperino.

Since its functional so far, it only would make sense to make changes here, after we worked out a concept of how the ui could be improved here. At the moment we have higher priorities, though.

martialblog:
Makes sense to have a concept first. 

[...]

PR #5429, opened 2025-10-02
Normalize confirm remove

martialblog:
This PR normalized the use of the phrase "Confirm Removal" in forms that confirm removal.

It also normalizes the use of the "cancel" icon.

See #5428

### DIFF
--- a/application/controllers/DashboardController.php
+++ b/application/controllers/DashboardController.php
@@ -183,7 +183,6 @@ class DashboardController extends ActionController
             }
             return true;
         });
-        $form->setTitle($this->translate('Remove Dashlet From Dashboard'));
         $form->setRedirectUrl('dashboard/settings');
         $form->handleRequest();
         $this->view->pane = $pane;
--- a/application/controllers/GroupController.php
+++ b/application/controllers/GroupController.php
@@ -149,7 +149,7 @@ class GroupController extends AuthBackendController
                 'class'         => 'link-button spinner',
                 'value'         => 'btn_submit',
                 'decorators'    => array('ViewHelper'),
-                'label'         => $this->view->icon('trash'),
+                'label'         => $this->view->icon('cancel'),
                 'title'         => $this->translate('Remove this member')
             ));
             $this->view->removeForm = $removeForm;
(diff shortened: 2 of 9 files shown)


martialblog:
phpcs fails not due to my code changes. Seems to be the same issue in other PRs `Class name "Zend_View_Helper_Util" is not in PascalCase format`

lippserd:
@martialblog Please rebase.

martialblog:
@lippserd Branch is rebased.

application/controllers/DashboardController.php
jrauh01: As you removed the title for dashlet removal, please remove this title as well.

jrauh01: Should be only one tab to align with the other confirmation pages:

```php
$this->getTabs()->add('remove-pane', [
    'active' => true,
    'label'  => $this->translate('Remove Dashboard'),
    'url'    => Url::fromRequest()
]);

martialblog: Done. Let me know if there's anything else

martialblog: Done. Let me know if there's anything else

martialblog: FYI, I went for consistency with the other methods, so no newlines added even though I wanted to

jrauh01:
> * **Role removal button alignment**: The confirmation button is currently right-aligned. It should be left-aligned to stay consistent with the other confirmation buttons.

Did you overlook this one, or are you on it?

martialblog:
@jrauh01 Not 100% sure if this is the correct way to do it https://github.com/Icinga/icingaweb2/pull/5429/commits/cdc20f576bcb0fea441b951f7afdc9330593d2cc The RoleForm seems to be different than the other forms.

jrauh01:
Yes, unfortunately the forms are all implemented slightly different. But The `UserForm` does set the `.icinga-controls` class as well, so I think your solution fits well with the existing code.

jrauh01:
@martialblog I just noticed I overlooked, that the button to remove user groups is right-aligned as well. Your fix won't apply cleanly in this case, and I'm not quite sure if it has to be left-aligned, because of the form description. What do you think about that?

martialblog:
I'm not a certified UX expert so honestly I can't say. We could also probably change the form description to be an HTML element outside the form and align the button. For me personally it's OK as is and since the description relates to the button it probably best to keep them together e.g. for screenreaders.

jrauh01:
For me it's fine to keep as is as well.

PR #5430, opened 2025-10-02
Two-factor authentication

jrauh01:
This PR is based on #5397 and will therefore close it. Introduces basic two-factor authentication (**2FA**) using Time-based One-Time Passwords (**TOTP**) support for Icinga Web 2.

TOTP (Time-based One-Time Password) is a type of two-factor authentication (2FA) method. It generates short-lived numeric codes (usually 6-8 digits) based on:
1. A shared secret key (known to both the server and your authenticator app).
2. The current time (in 30-second intervals by default).

Since the code changes frequently and is only valid for a short window, it makes accounts harder to hack, even if someone steals your password.

After enabling 2FA, users authenticate with their usual Icinga Web credentials and are then prompted to provide a valid TOTP code from their authenticator app (e.g., Google Authenticator, Authy).

We will use the defaults of **6 digit tokens**, **30 seconds interval** and **sha-1 algorithm** for token generation. The generated token is valid for **10 seconds before and after the current time** to allow some clock drift.

### Current Authentication Process

Currently, Icinga Web authentication follows this workflow:
- First it checks if there is set a remember me cookie from past logins.
- If so, it attempts to log the user in using the encrypted credentials stored in the cookie.
- If there is no remember me cookie set, it will show a login form.
- If submitted it will compare the password with the hash in the database.
- If correct it will setup the user and store it in the database.
- Additionally it will create the remember me cookie if the slider for that in the login form is checked.
- Then it will redirect to the dashboard or another target set by url param `redirect`.

### 2FA Implementation

This implementation is split in two main parts.
1. The configuration of 2FA via the account settings
2. The actual authentication process with token input

#### Configuration

The configuration of TOTP-based 2FA is fully integrated into the user’s account settings. A new form (`TwoFactorConfigForm`) allows users to enable 2FA, verify, or remove a TOTP secret directly from their account settings.

When a user with the required permission (`user/two-factor-authentication`) accesses their account, the `AccountController` loads the current TOTP secret from the database (`TwoFactorTotp::loadFromDb()`). If none exists, a new secret is generated (`TwoFactorTotp::generate()`) and passed into the form.

The form (`TwoFactorConfigForm`) then guides the user through the following workflow:

- **Enable 2FA**: If no TOTP secret is stored, the user can opt in via a checkbox. Once enabled, a new secret is created, and the form displays both a QR code (rendered by `TwoFactorTotp::createQRCode()`) and a manual provisioning URI. The user can scan this code with an authenticator app or copy the URI via a **copy-to-clipboard** element.
- **Verification Step**: To prevent misconfiguration, the user must enter a valid TOTP code generated by their app. The code is checked using the verify method of the `TwoFactorTotp` class. Only if the verification succeeds will the secret be persisted to the database (`saveToDb()`).
- **Remove 2FA**: If a secret already exists in the database, the form instead provides the option to remove it. On form submit, the stored secret is deleted (`removeFromDb()`). **(Should be extended in the future to ask for confirmation!)**

#### Authentication

Once a user has a TOTP secret stored, the login flow gains a second step. After the username and password are verified, the session sets a `2fa_must_challenge_token` flag, stores a temporary user object, and if “Stay logged in” was checked, a temporary remember-me cookie in the session. The remember-me cookie is not persisted yet. This prevents bypassing 2FA by simply reusing the cookie. `Auth::isAuthenticated()` also enforces that a user is only considered fully logged in once the TOTP challenge has succeeded, unless explicitly skipped for trusted cookies.

The `LoginForm` is rewritten now extending `ipl\Web\Compat\CompatForm` and decides which elements it should render. If the `2fa_must_challenge_token` flag is set in the session, the elements to verify the TOTP token are displayed, otherwise the normal login form elements. A valid remember-me cookie can bypass both, the password login and the token challenge: if present and verified, the cookie is renewed, persisted, and the user is logged in.

When the `LoginForm` is submitted via the `btn_submit_verify_2fa` submit button, the secret is loaded from the database and the entered code is verified. A successful check sets `$twoFactorSuccessful` to `true` on the user object that was temporarily stored in the session after password login, clears the `2fa_must_challenge_token` session flag, persists any deferred remember-me cookie, and calls `AuthenticationHook::triggerLogin()`. Submitting the form via the `btn_submit_cancel_2fa` button lets users abort, which purges the session and resets the login.

### Future

Possible extensions for the future could be:
- Encrypt secrets in the database
- Option for administrators to enforce 2fa for users
- Ask for password verification to add or remove a TOTP secret in the account settings

Requires GD php extension for QR code generation
Requires https://github.com/Icinga/icinga-php-thirdparty/pull/53
Closes #5397


jrauh01:
The failed phpcs tests are not due to my changes.

Al2Klimov:
> Requires https://github.com/Icinga/icinga-php-thirdparty/pull/53

I'm afraid I have to convert this PR to draft in this case.

> The failed phpcs tests are not due to my changes.

Well, the current master is all-✅.

schema/mysql.schema.sql
```diff
@@ -64,5 +64,12 @@ CREATE TABLE icingaweb_schema (
   CONSTRAINT idx_icingaweb_schema_version UNIQUE (version)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=DYNAMIC;
 
+CREATE TABLE `icingaweb_totp`(
```
Al2Klimov: According to @nilmerg, IW2 PRs with schema changes shall include schema upgrades.

schema/mysql.schema.sql
```diff
@@ -64,5 +64,12 @@ CREATE TABLE icingaweb_schema (
...
+  `username` varchar(254) COLLATE utf8mb4_unicode_ci NOT NULL,
+  `secret`   varchar(255) NOT NULL,
+  `ctime`    timestamp NULL DEFAULT NULL,
```
Al2Klimov: Why NULL here? From a high-level PoV, 2FA entries always have a creation time.

schema/mysql.schema.sql
```diff
@@ -64,5 +64,12 @@ CREATE TABLE icingaweb_schema (
...
+  PRIMARY KEY (`username`)
```
Al2Klimov: mtime is missing.

library/Icinga/Model/TotpModel.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Icinga\Model;
```
Al2Klimov: License header missing. Also check other newly added files.

library/Icinga/Model/TotpModel.php
```diff
@@ -0,0 +1,38 @@
...
+
+use ipl\Orm\Model;
+
+class TotpModel extends Model
+{
+
```
Al2Klimov:
```suggestion
```

library/Icinga/Model/TotpModel.php
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * @inheritDoc
+     */
```
Al2Klimov: @lippserd @nilmerg Correct me if anything, but IIRC, we don't use such anymore.

library/Icinga/Model/TotpModel.php
```diff
@@ -0,0 +1,38 @@
...
+    public function getTableName()
```
Al2Klimov: Not a requirement, but I personally declare return types, even of overridden methods. https://github.com/Icinga/developer-guidelines-drafts/issues/83

application/forms/Security/RoleForm.php
```diff
@@ -573,6 +573,9 @@ public static function collectProvidedPrivileges()
             'user/password-change' => [
                 'description' => t('Allow password changes in the account preferences')
             ],
+            'user/two-factor-authentication'=> [
+                'description' => t('Allow 2FA configuration in the account preferences')
+            ],
```
Al2Klimov: You seriously have to ask for permission to enable 2FA in your account?

I'd rather use this mechanism for the opposite.

application/controllers/AccountController.php
```diff
@@ -67,6 +72,19 @@ public function indexAction()
             }
         }
 
+        if ($user->can('user/two-factor-authentication')) {
+            $totp = IcingaTotp::loadFromDb($this->getDb(), $user->getUsername());
+            if ($totp == null) {
```
Al2Klimov:
```suggestion
            if ($totp === null) {
```

application/forms/Account/TotpConfigForm.php
```diff
@@ -0,0 +1,217 @@
...
+use Icinga\Web\Notification;
+use Icinga\Web\Session;
+use ipl\Stdlib\Filter;
+
+/**
+ * Form for enabling and disabling TOTP or creating and updating the TOTP secret
+ *
+ * This form is used to manage the TOTP settings of a user account.
+ */
+class TotpConfigForm extends Form
```
Al2Klimov: An ipl form makes sense, unless this is planned to be used as a subform in established stuff.

application/forms/Account/TotpConfigForm.php
```diff
@@ -0,0 +1,217 @@
...
+                ]
+            );
+        }
+
+        if ($dbTotp = IcingaTotp::loadFromDb($this->getDb(), $this->user->getUsername())) {
+            $this->addElement(
+                'text',
+                'user_totp_secret',
+                [
+                    'value' => $dbTotp->getSecret()
```
Al2Klimov: Why'd you show an existing secret to the user on re-display of the form?

IMAO showing it makes sense only once for usage as recovery code.

application/forms/Account/TotpConfigForm.php
```diff
@@ -0,0 +1,217 @@
...
+                    ),
+                ]
+            );
+        }
+
+        if ($dbTotp = IcingaTotp::loadFromDb($this->getDb(), $this->user->getUsername())) {
+            $this->addElement(
+                'text',
+                'user_totp_secret',
+                [
```
Al2Klimov: Why should I as a user generate and provide my own secret?

schema/mysql.schema.sql
Al2Klimov: Even if I imagine for a moment I had a schema upgrade, when I log in as admin to apply it via web UI, IW2 greets me with:

> SQLSTATE[42S02]: Base table or view not found: 1146 Table 'iw2.icingaweb_totp' doesn't exist

application/forms/Account/TotpConfigForm.php
Al2Klimov: As a user I may get a new phone sooner or later. To transfer 2FA from one phone to another, I could imagine disabling it temporarily. But this may be not wanted for security or policy reasons. Then I have to add a second TOTP temporarily, but neither the code nor the schema permit this.

At least I'd allow replacing a TOTP atomically.

Al2Klimov: As a user I'd like to see the secret exactly once, after enabling/replacing 2FA, together with a hint that I should use the secret as recovery code.

Or at least a hint that I can save the QR code somewhere for recovery purposes.

library/Icinga/Authentication/IcingaTotp.php
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    protected function setTotp(TOTP $totp): static
+    {
+        $this->totp = $totp;
+
+        return $this;
+    }
+
+    public static function generate(string $user): static
```
Al2Klimov: > We will use the defaults of 6 digit tokens, 30 seconds interval and sha-1 algorithm for secret generation.

The default? So it's not strictly required? Good! Because SHA1 is insecure af and both Authy and Google support more than that.

(Bitte gleich gescheit machen.)

application/forms/Account/TotpConfigForm.php
```diff
@@ -0,0 +1,217 @@
...
+    public function setTotp(IcingaTotp $totp): static
+    {
+        $this->totp = $totp;
+
+        return $this;
+    }
+
+    public function createElements(array $formData): void
+    {
+        if (! IcingaTotp::hasDbSecret($this->getDb(), $this->user->getUsername())) {
```
Al2Klimov: Isn't this DB query superfluous as you load the TOTP below anyway?

Al2Klimov (changes requested):

schema/mysql.schema.sql
jrauh01: `mtime` is not needed, because the entry is never modified.

Al2Klimov: Isn't it? What about replacement https://github.com/Icinga/icingaweb2/pull/5430#discussion_r2459873573? Shouldn't it be NULLed on removal, so that I can always see when I changed 2FA the last time?

If no, why is it in the Model then?

library/Icinga/Authentication/IcingaTotp.php
jrauh01: After discussing with @julianbrost we came to the conclusion, that SHA1 is totally sufficient for this use case. It's commonly used and is significantly more widespread in terms of compatibility with authenticator apps.

schema/mysql.schema.sql
jrauh01: On removal the secret should be completely removed. The column in the model will be removed.

schema/mysql.schema.sql
jrauh01: I would simply stick to how all other `ctime` columns are created in the schema.

application/forms/Account/TotpConfigForm.php
jrauh01: As far as I know an ipl controller is needed for an ipl form. But the `AccountController` is an Icinga Web controller. But feel free to correct me if I'm wrong here.

application/forms/Security/RoleForm.php
jrauh01: Yes, it should be controllable if the user is allowed to enable 2FA. Additionally I thought about a feature to **enforce** users to enable 2FA. But this would be for the future and not part of this PR.

library/Icinga/Authentication/IcingaTotp.php
julianbrost: Some more context (as I'm not a huge fan "someone said" comments without rationale): The HOTP/TOTP specification allows the choice of a hash function with SHA1 being the default and still being commonly used[^1]. So SHA1 is the "it will just work" option. For other hash function, you have to check support in corresponding apps and/or have a higher chance of "doesn't work with $app" reports in the future. Sure, something like SHA512 is a nicer choice, but just take a look at the output, these are 6 digit combinations, that's just 10^6 possible values, so this isn't really about collision resistance (you will see collisions) but more about deriving a value that depends on the shared secret without leaking the shared secret.[^2]

[^1]: I just randomly sampled my TOTP app and didn't find a single one using something other than SHA1, not even among the more recently added ones.
[^2]: The security properties of HMAC are actually quite interesting, in particular, it doesn't depend on the collision resistance of the hash function: https://en.wikipedia.org/wiki/HMAC#Security. So HMAC-MD5 is more secure than you might think, it suffers more from MD5 outputting only 128 bit than not being collision resistant. Though you probably can't directly apply this here. HOTP is based on HMAC, but it serves a different purpose than HMAC, so it's unclear if you can directly apply this here. Take it more as a hand-waving argument that TOTP is probably more limited by it's output length than the choice of hash function.

application/forms/Account/TotpConfigForm.php
Al2Klimov: According to @nilmerg you can just pass ServerRequest::fromGlobals() to handleRequest, no ipl controller necessary.

jrauh01: Would you apply this only to the `TotpConfigForm`? Or to the `Cancel2FAForm` and `Challenge2FAForm` as well?

Al2Klimov: Also ask Eric/Johannes to be sure, but I personally learned to base all new forms on ipl (ex. subforms of old forms, necessary base class is a Zend Form, ...).

jrauh01: The text input that shows what secret is stored in the database, was only a temporary solution. I thought about removing it completely and show the secret only in the token verification form in the `totp_manual_token_url` textarea.

application/forms/Account/TotpConfigForm.php
jrauh01: Yes, this was a temporary solution for development.

application/forms/Account/TotpConfigForm.php
jrauh01: There could be a later extension for that, but for now the workflow should be the following:
- Enable 2FA
- Scan QR code with authenticator app
- Verify the secret

If you want to change the secret you have to disable 2FA and enable it again.

Maybe there could be a way to show the QR code afterwards to add multiple devices.

Al2Klimov: > Maybe there could be a way to show the QR code afterwards to add multiple devices.

To get a 2FA recovery code on some platform, I've saved their QR code for the case I lose my device.

So I can scan it with a new device.

jrauh01: I hope you saved the QR codes on a different device than the one with your authenticator app on it, but I'm sure you have. To stay on topic, the user already has the option to screenshot the QR code themselves. Maybe we could add a hint like you mentioned in https://github.com/Icinga/icingaweb2/pull/5430#discussion_r2459880083.

Al2Klimov: > screenshot the QR code

# 😆

Seriously speaking: right click, save image as...

> add a hint like you mentioned

While on it, this can be a download button.

jrauh01: Yes, I've thought about a download button too.

application/controllers/AccountController.php
```diff
@@ -3,20 +3,27 @@
...
 use Icinga\Authentication\User\UserBackend;
+use Icinga\Common\Database;
 use Icinga\Data\ConfigObject;
 use Icinga\Exception\ConfigurationError;
 use Icinga\Forms\Account\ChangePasswordForm;
+use Icinga\Forms\Account\TwoFactorConfigForm;
 use Icinga\Forms\PreferenceForm;
 use Icinga\User\Preferences\PreferencesStore;
 use Icinga\Web\Controller;
+use ipl\Html\Contract\Form;
```
Al2Klimov: Heads up! `ipl\Html\Contract\Form` was added recently to ipl, so our icingaweb2 2.13.0(?) packages MUST raise their dependencies. I didn't check all classes, though.

schema/pgsql-upgrades/2.13.0.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE "icingaweb_2fa" (
```
Al2Klimov: While I can login in order to apply this migration, /account says `SQLSTATE[42P01]: Undefined table: 7 ERROR: relation "icingaweb_2fa" does not exist` until applied.

Can't speak for others, but _I personally_ could imagine just doing nothing here – i.e let the users fix it by applying the migration.

application/forms/Account/TwoFactorConfigForm.php
```diff
@@ -0,0 +1,210 @@
...
+use ipl\Web\Url;
+use ipl\Web\Widget\ActionLink;
+use ipl\Web\Widget\CopyToClipboard;
+
+/**
+ * Form for enabling and disabling 2FA or creating and updating the 2FA TOTP secret
+ *
+ * This form is used to manage the 2FA settings of a user account.
+ */
+class TwoFactorConfigForm extends CompatForm
```
Al2Klimov: I was barely able to integrate https://github.com/Icinga/icinga-php-thirdparty/pull/53 for testing this.

* `composer update` on the PR said `The process "patch '-p1' --no-backup-if-mismatch -d '/Users/aklimov/NET/WS/icingaweb2/icinga-php-thirdparty/vendor/shardj/zf1-future' < '/Users/aklimov/NET/WS/icingaweb2/icinga-php-thirdparty/patches/shardj-zf1-future.patch'" exceeded the timeout of 300 seconds.`
* `composer require spomky-labs/otphp chillerlan/php-qrcode` on v0.14.0 missed the class for QR codes, so I had to comment those out

Please:

* Get https://github.com/Icinga/icinga-php-thirdparty/pull/53 ready
* Make me a commit based on it which contains vendor/, so I can check it out just like a release tag

Al2Klimov: As I had no QR code, I made one myself from the shown URL:

1. `qrencode -o qr.png`
2. Paste otpauth:// URL
3. Enter
4. Ctrl-D
5. `open qr.png`

But it didn't work:

IMG_5175

So I had to copy and paste just the secret to my phone.

application/forms/Authentication/LoginForm.php
```diff
@@ -6,122 +6,212 @@
...
+use ipl\Web\Common\CsrfCounterMeasure;
+use ipl\Web\Common\FormUid;
+use ipl\Web\Compat\CompatForm;
+use ipl\Web\Compat\FormDecorator\CheckboxDecorator;
 
 /**
  * Form for user authentication
  */
-class LoginForm extends Form
+class LoginForm extends CompatForm
```
Al2Klimov, line 32: Heads up, security hole! On Postgres, I can bypass 2FA for e.g icingaadmin by logging in as ICINGAADMIN.

Al2Klimov (changes requested):
Tbh, I as a reviewer don't wanna even think about whose (branch) fault the red GHA is.

Just (someone) get it green, please. Thanks.

(Notes for the future myself:

<details>
  <summary>Test system (I use NixOS btw)</summary>

```nix
{ pkgs, ... }: {
  imports = [
    ./hardware-configuration.nix


  ];

  boot.tmp.cleanOnBoot = true;
  networking.hostName = "aklimov-2fa";
  networking.domain = "";
  services.openssh.enable = true;
  users.users.root.openssh.authorizedKeys.keys = [''ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIBIroHYGSRaRNFxlK90SS0aHwWjEME30pK5J1N/V1w6a aklimov@ws-aklimov7777777.local'' ];
  system.stateVersion = "23.11";

networking.firewall.allowedTCPPorts = [ 80 ];

virtualisation.oci-containers = {
backend = "podman";
containers.postgresql = {
autoStart = true;
image = "postgres:9.1";
environment = {
POSTGRES_PASSWORD = "123456";
};
ports = [ "127.0.0.1:5432:5432" ];
volumes = [ "postgres:/var/lib/postgresql/data" ];
};
};

  services.icingaweb2 = {
    enable = true;
    #virtualHost = "iw2.aklimov.net-dump.de";
    generalConfig.global.config_resource = "iw2";
    modules.monitoring.enable = false;
    authentications = {
      pg = {
        backend = "db";
        resource = "iw2";
      };
    };
    resources.iw2 = {
        type = "db";
        db = "pgsql";
        host = "127.0.0.1";
        port = "5432";
        dbname = "postgres";
        username = "postgres";
        password = "123456";
        charset = "utf8";
      };
    roles = {
      adm = {
        users = "icingaadmin,Alexander A. Klimov";
        permissions = "*";
      };
    };
    modulePackages = {
    };
  };

nixpkgs.overlays = [
(_: prev: {
icingaweb2-ipl = prev.icingaweb2-ipl.overrideAttrs (old: {
src = prev.fetchFromGitHub {
owner = "Icinga";
repo = "icinga-php-library";
rev = "v0.18.1";
hash = "sha256-V2PkRtVzMnzBpNLUMsSl5NUSnrycgEHGXLGilbhcGfw=";
};
}); })
(_: prev: {
icingaweb2-thirdparty = prev.icingaweb2-thirdparty.overrideAttrs (old: {
src = prev.fetchFromGitHub {
owner = "Icinga";
repo = "icinga-php-thirdparty";
#rev = "274212c7e02223687eacdfbc78c48181878ac910"; # feature/2fa
#hash = "sha256-K+NRUDnQjGG0Lmi5AOtftfueZM+VE6qkOWXF9qLzi/Y=";
rev = "1983c72cd3bbc0f4f0df5f464db4985417b4e73d"; # aklimov/feature/2fa
hash = "sha256-3CIAfhSr9GGhytOPnSfnnt/H6xWJOlQzyrZpux0VZbI=";
};
}); })
    (_: prev: { icingaweb2 = prev.icingaweb2.overrideAttrs (old: {
      src = prev.fetchFromGitHub {
        owner = "Icinga";
        repo = "icingaweb2";
        #rev = "62f10871e9e270789def9a3776656ff15d094013"; # 2fa-new
        #hash = "sha256-iCoCCEGLO3vf9VdbDWmecBMaRuUAKJc7f9AeNKLQv0A=";
rev = "e0e2a22fbeb90ae721b5d301af235467aa22d3ef"; # aklimov/2fa-new
hash = "sha256-0SaOkROWkc20j7MxGWsx5q4aSQoRTDaQazd2SJjCPbU=";
#rev = "70f29827f921f24b2afc0556c4800dfa55a2d0c1"; # main
#hash = "sha256-fCd5kyJk7LXdKkZS7kppPO1wfeNTOD4esliqvSAtyV4=";
      };

      patches = [
        ./opcache_reset.patch

        # https://github.com/Icinga/icingaweb2/issues/5427
        ./migrations-db-no-pw.patch
      ];

      # https://github.com/NixOS/nixpkgs/pull/380065
      installPhase = old.installPhase + "\ncp -ra schema $out";
    }); })
  ];

}
```
</details>

application/forms/Authentication/LoginForm.php
Al2Klimov: @nilmerg FYI same applies to user preferences. (Not a security hole, though.) I.e ICINGAadmin and icingaADMIN have distinct user preferences.😅

schema/mysql.schema.sql
Al2Klimov: "all other ctime columns" are timestamp, yours a bigint.

Al2Klimov (commented):
Almost forgot! The commit history got out of hand. Consider making one or a few large commits with shared authorship where applicable:

https://docs.github.com/de/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

application/forms/Security/RoleForm.php
Al2Klimov: > Yes, it should be controllable if the user is allowed to enable 2FA.

But why?

lippserd: I don't remember if we discussed this, but I wouldn't introduce any permission for enabling 2FA either.

Al2Klimov: And for disabling? (Company policy, GitLab also has this, ...)

Al2Klimov (changes requested):
@lippserd Please write your requirements on 2FA + API requests.

* They just shall not work? (What a pity.)
* They just shall work with user:pass? (Security hole, our forms [already allow](https://git.icinga.com/icingaweb2/icingaweb2-module-partnerportal/-/issues/151#note_28370) API requests implicitly!)
* OTP header required?
* OTP is to be appended to the password? (`username:password;otp`)
* ...

Al2Klimov:
... or personal access token? (idea (c) @TheSyscall)

application/forms/Account/TwoFactorConfigForm.php
jrauh01: Done. Try to check out to `2686bdc3a0761519a4ffe432086a10cb19be921d`, or just pull https://github.com/Icinga/icinga-php-thirdparty/pull/53.

schema/pgsql-upgrades/2.13.0.sql
jrauh01: Yes, I also would let the user fix it by applying the migration.

application/forms/Security/RoleForm.php
jrauh01: There is a permission for changing passwords, so in my opinion there should be one for changing 2FA. Alternatively if a user has the permission to change their password, they could also be permitted to enable/disable 2FA.

For a future PR I would still keep in mind discussing an option to enforce 2FA for all users.

Al2Klimov: > if a user has the permission to change their password, they could also be permitted to enable/disable 2FA.

# 👍

application/controllers/AccountController.php
jrauh01: Done, in https://git.icinga.com/packages/icingaweb/-/merge_requests/10.

schema/mysql.schema.sql
jrauh01: Like we discussed offline I left them `bigint` and set to `NOT NULL`.

jrauh01:
Superseeded by #5514

PR #5432, opened 2025-10-08
Rename rule: ignore NotPascalCase in views/helpers

JolienTrog:
PHP_CodeSniffer 4.0 renamed the class name NotCamelCaps to
NotPascalCase.
Reference: https://github.com/PHPCSStandards/PHP_CodeSniffer/wiki/Version-4.0-User-Upgrade-Guide#upgrading-5

Because of this change, our current configuration was no longer compatible.
I tested the update inside a Docker container: after switching to the new class name `NotPascalCase`, the error message disappeared and the rules are applied as expected.


lippserd (changes requested):
Our commit messages start with a capital letter, so please use `Rename` instead of `rename`. Also, use the first paragraph of the PR description as the commit description and format it correctly in terms of line length.

PR #5433, opened 2025-10-10
Request hook

jrauh01:
Add a `RequestHook` that hooks into different states of a request. The implementation starts with hooking into after **non** xhr requests are dispatched. This is meant to be extended for more more states like for example before a request gets dispatched.

### DIFF
--- a/library/Icinga/Application/Hook/RequestHook.php
+++ b/library/Icinga/Application/Hook/RequestHook.php
@@ -27,12 +27,32 @@ abstract class RequestHook
      */
     final public static function postDispatch(Request $request): void
     {
-        foreach (Hook::all('Request') as $hook) {
+        foreach (static::all() as $hook) {
             try {
                 $hook->onPostDispatch($request);
             } catch (Throwable $e) {
                 Logger::error('Failed to execute hook on request: %s', $e);
             }
         }
+    }
+
+    /**
+     * Get all registered implementations
+     *
+     * @return static[]
+     */
+    public static function all(): array
+    {
+        return Hook::all('RequestHook');
+    }
+
+    /**
+     * Register the class as a RequestHook implementation
+     *
+     * Call this method on your implementation during module initialization to make Icinga Web aware of your hook.
+     */
+    public static function register(): void
+    {
+        Hook::register('RequestHook', static::class, static::class, true);
     }
 }


library/Icinga/Application/Hook/RequestHook.php
```diff
@@ -0,0 +1,33 @@
...
+    /**
+     * Call the onPostDispatch() method of all registered RequestHooks
+     *
+     * @param Request $request
+     *
+     * @return void
+     */
+    final public static function postDispatch(Request $request): void
+    {
+        array_map(
```
Al2Klimov: Why array_map? You don't use its result.

library/Icinga/Application/Hook/RequestHook.php
```diff
@@ -0,0 +1,38 @@
...
+     * @param Request $request
+     *
+     * @return void
+     */
+    final public static function postDispatch(Request $request): void
+    {
+        foreach (static::all('Request') as $hook) {
+            try {
+                $hook->onPostDispatch($request);
+            } catch (Throwable $e) {
```
Al2Klimov, line 33: Throwable is very broad. It's just like you'd catch BaseException in Python which includes SystemExit, GeneratorExit and KeyboardInterrupt. In PHP, Throwable includes Error which includes ArgumentCountError, ParseError, CompileError, ...

I just think that normally in PHP code we'd simply call other PHP code. But here... I'd catch just Exception.

library/Icinga/Web/Controller/ActionController.php
```diff
@@ -520,6 +521,8 @@ public function postDispatch()
 
         if ($this->isXhr()) {
             $this->postDispatchXhr();
+        } else {
+            RequestHook::postDispatch($req);
```
Al2Klimov: I think letting the hook decide would be clever AND smart. I mean, anyone could provide a hook... and wonder why some requests are missing. If `$this->isXhr()` is too heavy just with `$req`, we can pass additional arguments like a bool.

lippserd: I already discussed this w/ @jrauh01. We will call `postDispatch()` unconditionally.

library/Icinga/Application/Hook/RequestHook.php
lippserd: We (could) call PHP code that is not under our control. So why should we crash the application in this case? Catching `Throwable` is perfectly fine here. Especially since the error is logged. This is exactly the kind of use case `Throwable` is designed for.

library/Icinga/Web/Controller/ActionController.php
lippserd: @jrauh01 Please also add the following to our `Request`:

```php
public function isAutoRefresh(): bool
{
    return $this->getHeader('X-Icinga-Autorefresh');
}
```

library/Icinga/Application/Hook/RequestHook.php
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Call the onPostDispatch() method of all registered RequestHooks
+     *
+     * @param Request $request
+     *
+     * @return void
+     */
+    final public static function postDispatch(Request $request): void
+    {
+        foreach (static::all('Request') as $hook) {
```
Al2Klimov: Are you extending Hook just to use static:: here?

Al2Klimov (changes requested):
As this is a new hook, add the new methods just like in #5442:

* all()
* register()

jrauh01:
> As this is a new hook, add the new methods just like in #5442:
> 
> * all()
> * register()

I added them and I saw there is already an implementation of `Hook\Essentials` in #5474 . Depending on what PR will be merged first, the other PR has to be adjusted.

PR #5434, opened 2025-10-16
Add Service Display Name to Macros

svbroeker:
We had the Issue that we need a the Text from Service Display Names for Actions, so we add these to the Code from icingaweb2


svbroeker:
> Thank you for your pull request. Before we can look at it, you'll need to sign a Contributor License Agreement (CLA).
> 
> **Please follow instructions at https://icinga.com/company/contributor-agreement to sign the CLA.**
> 
> After that, please reply here with a comment and we'll verify.
> 
> Contributors that have not signed yet: @svbroeker

signed

bobapple:
@cla-bot check

nilmerg:
Thank you, but I have to point out that this is possible without changing the code. Just use `$service_display_name$`.

PR #5442, opened 2025-11-05
LoginButtonHook: additional buttons below the login form

Al2Klimov:
Introduces `LoginButtonHook`, a new hook for rendering additional buttons below the login form. Extend this class to display custom buttons on the Icinga Web login page — useful for alternative authentication flows such as SSO. Register your implementation by calling `YourLoginButtons::register()` during module initialization.

### DIFF
--- a/application/controllers/AuthenticationController.php
+++ b/application/controllers/AuthenticationController.php
@@ -3,9 +3,14 @@
 
 namespace Icinga\Controllers;
 
+use GuzzleHttp\Psr7\ServerRequest;
+use Icinga\Application\ClassLoader;
 use Icinga\Application\Hook\AuthenticationHook;
+use Icinga\Application\Hook\LoginButtonHook;
 use Icinga\Application\Icinga;
 use Icinga\Application\Logger;
+use Icinga\Authentication\LoginButton;
+use Icinga\Authentication\LoginButtonForm;
 use Icinga\Common\Database;
 use Icinga\Exception\AuthenticationException;
 use Icinga\Forms\Authentication\LoginForm;
@@ -14,6 +19,7 @@ use Icinga\Web\Helper\CookieHelper;
 use Icinga\Web\RememberMe;
 use Icinga\Web\Url;
 use RuntimeException;
+use Throwable;
 
 /**
  * Application wide controller for authentication
@@ -93,7 +99,33 @@ class AuthenticationController extends Controller
             }
             $form->handleRequest();
         }
+
+        $loginButtons = [];
+        $request = ServerRequest::fromGlobals();
+
+        foreach (LoginButtonHook::all() as $class => $hook) {
+            try {
+                foreach ($hook->getButtons() as $index => $button) {
+                    assert($button instanceof LoginButton);
+
+                    $loginButtons[] = (new LoginButtonForm(
+                        sha1("$class!$index"),
+                        $button,
+                        ClassLoader::classBelongsToModule($class) ? ClassLoader::extractModuleName($class) : null
+                    ))
+                        ->on(LoginButtonForm::ON_SUCCESS, function () use ($button): void {
+                            ($button->onClick)();
+                        })
+                        ->handleRequest($request);
+                }
+            } catch (Throwable $e) {
+                Logger::error('Failed to execute login button hook: %s', $e);
+                continue;
+            }
+        }
+
         $this->view->form = $form;
+        $this->view->loginButtons = $loginButtons;
         $this->view->defaultTitle = $this->translate('Icinga Web 2 Login');
         $this->view->requiresSetup = $requiresSetup;
     }
--- a/library/Icinga/Application/Hook/LoginButtonHook.php
+++ b/library/Icinga/Application/Hook/LoginButtonHook.php
@@ -1,0 +1,47 @@
+<?php
+/* Icinga Web 2 | (c) 2026 Icinga GmbH | GPLv2+ */
+
+namespace Icinga\Application\Hook;
+
+use Icinga\Application\Hook;
+use Icinga\Authentication\LoginButton;
+
+/**
+ * Hook for adding custom buttons below the login form
+ *
+ * Extend this class to display additional buttons on the Icinga Web login page,
+ * e.g. for alternative authentication flows such as SSO.
+ * To register your implementation, call `YourLoginButtons::register()` during module initialization.
+ */
+abstract class LoginButtonHook
+{
+    /**
+     * Get the buttons to display below the login form
+     *
+     * Implement this method to return any number of {@link LoginButton} instances.
+     * Each button is rendered below the standard login form in the order provided.
+     *
+     * @return LoginButton[]
+     */
+    abstract public function getButtons(): array;
+
+    /**
+     * Get all registered implementations
+     *
+     * @return static[]
+     */
+    public static function all(): array
+    {
+        return Hook::all('LoginButton');
+    }
+
+    /**
+     * Register the class as a LoginButton hook implementation
+     *
+     * Call this method on your implementation during module initialization to make Icinga Web aware of your hook.
+     */
+    public static function register(): void
+    {
+        Hook::register('LoginButton', static::class, static::class, true);
+    }
+}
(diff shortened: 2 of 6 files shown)


library/Icinga/Application/LBH.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        return [
+            new LoginButton(
+                function () {
+                    var_dump(42);
+                    die;
+                },
+                'LolCat',
+                'https://al2klimov.de/teapot.jpg'
+            ),
```
Al2Klimov: This is a minimal hook example I use for testing.

Al2Klimov (commented):
@lippserd Any feedback on the tech side before anyone applies makeup on the buttons?

application/controllers/AuthenticationController.php
```diff
@@ -93,7 +98,34 @@ public function loginAction()
             }
             $form->handleRequest();
         }
+
+        $loginButtons = [];
+        $request = ServerRequest::fromGlobals();
+
+        foreach (Hook::all('LoginButton') as $class => $hook) {
```
lippserd: I know that this is how we do it almost everywhere, but it is not helpful for finding hook usages. I would change the implementation so that we call `LoginButtonHook::all()`.

library/Icinga/Application/Hook/LoginButton/LoginButton.php
```diff
@@ -0,0 +1,72 @@
...
+    protected $onSuccess;
+
+    /**
+     * Create an additional button to be displayed below the login form
+     *
+     * @param callable $onSuccess What to do if the button is pressed
+     * @param string $label Text to show on the button
+     * @param ?string $logoUrl URL to an image to show on the button before the text
+     */
+    public function __construct(callable $onSuccess, string $label, ?string $logoUrl = null)
```
lippserd: Please make the class `readonly`  and use constructor property promotion.

lippserd: I would also use `logoUri` instead of `Url`.

lippserd: If I think about it a little longer, we could make this more flexible by providing `ValidHtml $content` and `?Attributes $attributes` instead of `label` and `logoUri`. Then it would be possible to provide either images or icons, as well as other important attributes such as `title`. What do you think?

application/forms/Authentication/LoginWithForm.php
```diff
@@ -0,0 +1,64 @@
...
+use ipl\Html\Form;
+use ipl\Html\FormElement\SubmitButtonElement;
+use ipl\Html\Html;
+use ipl\Web\Common\CsrfCounterMeasure;
+use ipl\Web\Common\FormUid;
+
+/**
+ * Form for user authentication via external identity providers
+ */
+class LoginWithForm extends Form
```
lippserd: I would name this form `ButtonForm` since it is not tied to login, and I would also mark it as `readonly` and use constructor property promotion.

lippserd (changes requested):

application/forms/Authentication/LoginWithForm.php
Al2Klimov: In short, Icinga Web 2.13 will require PHP 8.2?

lippserd: Sure, why not?

Al2Klimov: No why not. Just a clarifying question from my side. :-)

library/Icinga/Application/Hook/LoginButton/LoginButton.php
Al2Klimov: Title can be another property (if you want) and an image is already possible, though unnecessary. IW2 users don't login with Google/ Facebook/ Apple/ ... At best they ["Login with id.netways.de"](https://rt.icinga.com). In an actual big company, how many additional login buttons do you expect? Even when 2+, there's text on them saying what they're for.

At most I'd support arbitrary attributes. If we let users add whatever they want, they'll need to style it (CSP friendly) sooner or later, but you said _we_ are responsible for the style of those buttons. Though, users already can style global things using the following hack.

Icinga Web wraps module LESS like this:

```css
/* wrapper */ icinga-module-something {
/* module LESS */ button {
/* module LESS */ color: #234269;
/* module LESS */ }
/* wrapper */ }
```

Knowing this you can jailbreak:

```css
/* wrapper */ icinga-module-something {
/* module LESS */ }
/* module LESS */ button {
/* module LESS */ color: #234269;
/* module LESS */ }
/* module LESS */ html {
/* wrapper */ }
```

lippserd: I just think we should design the hook so that it works optimally for us and is flexible enough to do everything we want. Therefore, I suggest to use `content` and `attributes` and embed the buttons in a module CSS container so that module CSS can be applied to the buttons.

application/controllers/AuthenticationController.php
```diff
@@ -93,7 +99,33 @@ public function loginAction()
...
+        $loginButtons = [];
+        $request = ServerRequest::fromGlobals();
+
+        foreach (LoginButtonHook::all() as $class => $hook) {
+            try {
+                foreach ($hook->getButtons() as $index => $button) {
+                    assert($button instanceof LoginButton);
+
+                    $loginButtons[] = (new ButtonForm(
+                        "$class!$index",
```
lippserd: Since `$index` is provided by the hook implementation, it can be anything, but the `name` attribute is subject to certain rules. Have you checked what happens if these are violated? Also, `$class` contains backslashes, which is technically possible in `name` but discouraged. Anyway, can you come up with something else?

application/forms/Authentication/ButtonForm.php
```diff
@@ -0,0 +1,42 @@
...
+use ipl\Html\Form;
+use ipl\Html\FormElement\SubmitButtonElement;
+use ipl\Html\Html;
+use ipl\Web\Common\CsrfCounterMeasure;
+use ipl\Web\Common\FormUid;
+
+/**
+ * Form for user authentication via external identity providers
+ */
+class ButtonForm extends Form
```
lippserd: We have probably already discussed the naming, but please rename it to `LoginButtonForm` and move it to `Icinga\Authentication`. `ButtonForm` would be appropriate if we could use it with "Check now" for example, but this implementation is tightly coupled with `LoginButton`.

library/Icinga/Authentication/LoginButtonForm.php
```diff
@@ -0,0 +1,42 @@
...
+{
+    use FormUid;
+    use CsrfCounterMeasure;
+
+    public function __construct(
+        string $name,
+        protected readonly LoginButton $button,
+        protected readonly ?string $moduleName = null
+    ) {
+        $this->defaultAttributes['name'] = $name;
```
lippserd, line 26: To be honest, I had to check why this line is necessary, so maybe you could add a comment explaining that `name` is needed for the subsequent creation of the `uid` element.

library/Icinga/Application/Hook/LoginButton/LoginButton.php
```diff
@@ -0,0 +1,25 @@
+<?php
+/* Icinga Web 2 | (c) 2025 Icinga GmbH | GPLv2+ */
+
+namespace Icinga\Application\Hook\LoginButton;
```
lippserd: Please move it to `Icinga\Authentication`.

library/Icinga/Application/Hook/LoginButtonHook.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
lippserd, line 1: The class and its methods have no documentation. Please add appropriate documentation about the purpose of the hook, its methods, and its usage.

lippserd (changes requested):

library/Icinga/Application/Hook/LoginButtonHook.php
Al2Klimov: * Also, most of the methods will be superfluous once we #5474

lippserd (approved):

PR #5454, opened 2025-11-24
Support PHP 8.5

### LINKED ISSUE #5269: Support PHP 8.4
RC1 is already in Debian experimental.

Admittedly, I haven't tried it out, yet.

### LINKED ISSUE #5450: Support for PHP 8.5
PHP 8.5 is currently in RC4 and is due to be released in one week. Compatibility should be added in icingaweb.

BastianLedererIcinga:
### PHP 8.4

- Change implicit nullable type declaration to explicit
- Only set `session.sid_bits_per_character` for PHP < 8.4 

[Migration Docs](https://www.php.net/manual/en/migration84)

### PHP 8.5

- Remove no-op `ReflectionProperty::setAccessible()` calls
- Replace backtick operator with `shell_exec()`
- Avoid passing `null` as key in `array_key_exists()`
- Avoid passing `null` to non-nullable parameters in built-in functions
- Use `Pdo\Mysql` driver-specific constants   

[Migration Docs](https://www.php.net/manual/en/migration85)

### Other compatibility fixes

- Avoid passing `null` as `ipl\Html` form element `$name`
- Use latest PHPUnit version

resolves #5269
resolves #5450 
requires Icinga/ipl-sql#99


library/Icinga/Application/Hook/ApplicationStateHook.php
```diff
@@ -40,7 +40,7 @@ final public function getMessages()
     final public function addError($id, $timestamp, $message)
     {
         $id = trim($id);
-        $timestamp = (int) $timestamp;
+        $timestamp = (int) floor($timestamp);
```
nilmerg: Please make *absolutely* sure we're dealing with floats here. My feeling is that this change isn't strictly necessary since we're dealing with timestamps and those are *usually* a result of `time()` or `DateTime->getTimestamp()`, in which case it's always an int.

There are probably other cases as well, I won't comment on each one. Please review them again and ensure on your own that you don't floor unnecessarily. It's maybe also an option to prevent decimals in the first place, given the calls are internal only. Though, an exception is `Benchmark`, as it already floored its timestamp previously so there's no problem to do it still.

library/Icinga/Application/ClassLoader.php
```diff
@@ -258,7 +258,7 @@ protected function isApplicationPrefix($prefix)
      */
     protected function namespaceHasApplictionDirectory($namespace)
     {
-        return array_key_exists($namespace, $this->applicationDirectories);
+        return array_key_exists($namespace ?? '', $this->applicationDirectories);
```
nilmerg: Why exactly does this need to be guarded but `isApplicationPrefix` does not? If I follow the calls of `namespaceHasApplictionDirectory`, I don't seem to find a case when the argument is null.

library/Icinga/Application/Hook/AuditHook.php
```diff
@@ -96,7 +96,7 @@ public function formatMessage($message, array $messageData)
      */
     protected function extractMessageValue(array $path, array $messageData)
     {
-        $key = array_shift($path);
+        $key = array_shift($path) ?? '';
```
nilmerg: null won't be part of the array, that's for sure. The regex above must match before this is called and then the match is passed. The recursion on the other hand handles the case already that it stops once the arrays is empty.

library/Icinga/Authentication/AdmissionLoader.php
```diff
@@ -235,7 +235,7 @@ public static function migrateLegacyPermissions(array $permissions)
         $refusals = [];
 
         foreach ($permissions as $permission) {
-            if (array_key_exists($permission, self::LEGACY_PERMISSIONS)) {
+            if ($permission && array_key_exists($permission, self::LEGACY_PERMISSIONS)) {
```
nilmerg: I'm pretty sure, that if this would be null, remaining code parts will still fail and would have already anyway. So this can't be necessary. :thinking:

modules/monitoring/library/Monitoring/Backend/Ido/Query/EventgridQuery.php
```diff
@@ -42,7 +42,7 @@ protected function joinBaseTables()
      */
     public function order($columnOrAlias, $dir = null)
     {
-        if (array_key_exists($columnOrAlias, $this->additionalColumns)) {
+        if ($columnOrAlias && array_key_exists($columnOrAlias, $this->additionalColumns)) {
```
nilmerg: that's another case of, if you'd follow the alternative path, you'd see that null is not a probably candidate as nothing else works anyway if that's the case.

modules/monitoring/library/Monitoring/Backend/Ido/Query/IdoQuery.php
```diff
@@ -1283,7 +1283,7 @@ protected function customvarNameToTypeName($customvar)
 
     protected function hasJoinedVirtualTable($name)
     {
-        return array_key_exists($name, $this->joinedVirtualTables);
+        return array_key_exists($name ?? '', $this->joinedVirtualTables);
```
nilmerg: Hard to believe that this is required, as my usage search didn't reveal a single case of *unguessable* types, so to say.

nilmerg:
Also, the changes to the monitoring module seem implausible to be complete. You already acknowledged in person that you didn't test it yet since you don't have the IDO enabled/configured. Please talk to @lippserd whether that's necessary to begin with.

nilmerg:
The fact that 730af4c4a00bb1a18a8443bc95dd3be9d4adcba4 doesn't conditionally call `setAccessible` still, means that this only works on PHP >= 8.1, so this plays into the question what will happen with monitoring.

library/Icinga/Common/Database.php
```diff
@@ -35,9 +35,18 @@ protected function getDb(): Connection
             $config->charset = 'utf8mb4';
         }
 
+        // In PHP 8.5+, driver specific constants of the PDO class are deprecated,
+        // but the replacements are ony available since php 8.4
+        if (version_compare(PHP_VERSION, '8.4.0', '<')) {
+            $mysqlConstantPrefix = 'PDO::MYSQL_ATTR_';
```
sukhwinder33445: Please assign the constant directly.

library/Icinga/Application/Hook/ApplicationStateHook.php
sukhwinder33445: The change has been reversed.

modules/monitoring/library/Monitoring/Backend/Ido/Query/IdoQuery.php
sukhwinder33445: The `monitoring` module is no longer part of `Icingaweb2`.

modules/monitoring/library/Monitoring/Backend/Ido/Query/EventgridQuery.php
sukhwinder33445: The `monitoring` module is no longer part of `Icingaweb2`.

library/Icinga/Authentication/AdmissionLoader.php
sukhwinder33445: This change has been reversed.

library/Icinga/Application/Hook/AuditHook.php
sukhwinder33445: This change has been reversed.

library/Icinga/Application/ClassLoader.php
sukhwinder33445: This change has been reversed

library/Icinga/Common/Database.php
```diff
@@ -35,9 +35,18 @@ protected function getDb(): Connection
             $config->charset = 'utf8mb4';
         }
 
+        // In PHP 8.5+, driver specific constants of the PDO class are deprecated,
+        // but the replacements are ony available since php 8.4
```
sukhwinder33445: There is a typo: `ony`.

sukhwinder33445 (approved):
LGTM 👍🏽

library/Icinga/Application/Hook/AuditHook.php
```diff
@@ -16,13 +16,14 @@ abstract class AuditHook
...
-     * @param   string  $message    A detailed description possibly referencing parameters in $data
-     * @param   array   $data       Additional information (How this is stored or used is up to each implementation)
-     * @param   string  $identity   An arbitrary name identifying the responsible subject, defaults to the current user
-     * @param   int     $time       A timestamp defining when the activity occurred, defaults to now
+     * @param   string      $type       An arbitrary name identifying the type of activity
+     * @param   string      $message    A detailed description possibly referencing parameters in $data
+     * @param   array|null  $data       Additional information (How this is stored or used is up to each implementation)
+     * @param   string      $identity   An arbitrary name identifying the responsible subject,
+     *                                   defaults to the current user
+     * @param   int         $time       A timestamp defining when the activity occurred, defaults to now
```
lippserd:
```suggestion
     * @param string $type An arbitrary name identifying the type of activity
     * @param string $message A detailed description possibly referencing parameters in $data
     * @param ?array $data Additional information (How this is stored or used is up to each implementation)
     * @param string $identity An arbitrary name identifying the responsible subject, defaults to the current user
     * @param int $time A timestamp defining when the activity occurred, defaults to now
```

library/Icinga/Authentication/User/LdapUserBackend.php
```diff
@@ -356,15 +356,15 @@ public function requireTable($table, RepositoryQuery $query = null)
...
-     * @param   string              $table  The table where to look for the column or alias
-     * @param   string              $name   The name or alias of the column to validate
-     * @param   RepositoryQuery     $query  An optional query to pass as context
+     * @param   string                   $table  The table where to look for the column or alias
+     * @param   string                   $name   The name or alias of the column to validate
+     * @param   RepositoryQuery|null     $query  An optional query to pass as context
      *
      * @return  string                      The given column's name
      *
      * @throws  QueryException              In case the given column is not a valid query column
```
lippserd:
```suggestion
     * @param string $table The table where to look for the column or alias
     * @param string $name The name or alias of the column to validate
     * @param ?RepositoryQuery $query An optional query to pass as context
    *
    * @return string The given column's name
    *
    * @throws QueryException In case the given column is not a valid query column
```

library/Icinga/Web/Widget/SortBox.php
```diff
@@ -154,11 +154,11 @@ protected function getSortDefaults($column = null)
     /**
      * Apply the sort rules from the given or current request on the query
      *
-     * @param   Request     $request
+     * @param   Request|null     $request
      *
      * @return  $this
```
lippserd:
```suggestion
     * @param ?Request $request
    *
    * @return $this
```

library/Icinga/Web/Widget/FilterEditor.php
```diff
@@ -94,11 +94,11 @@ public function getFilter()
     /**
      * Set columns to search in
      *
-     * @param array $searchColumns
+     * @param array|null $searchColumns
```
lippserd:
```suggestion
     * @param ?array $searchColumns
```

library/Icinga/Authentication/AuthChain.php
```diff
@@ -85,9 +85,9 @@ class AuthChain implements Authenticatable, Iterator
     /**
      * Create a new authentication chain from config
      *
-     * @param Config $config User backends configuration
+     * @param Config|null $config User backends configuration
```
lippserd:
```suggestion
     * @param ?Config $config User backends configuration
```

lippserd:
I will push an upgrading entry for the IPL requirement after I have determined which version we will use for introducing backward-incompatible changes, and then merge the PR.

PR #5456, opened 2025-12-02
Only collapse the y-axis for collapsible elements

jrauh01:
Only collapse the y-axis if an collapsible item is collapsed. Additional styling in the horizontal level that overflows the x-axis  should still be possible.

### DIFF
--- a/public/css/icinga/main.less
+++ b/public/css/icinga/main.less
@@ -385,7 +385,7 @@ details.collapsible:not([open]) > .collapsible-control {
 // Collapsibles
 
 .collapsible.collapsed:not(details) {
-  overflow: hidden;
+  overflow-y: clip;
 }
 
 .collapsible.collapsed:not([data-toggle-element], details) {


public/css/icinga/main.less
```diff
@@ -404,7 +404,7 @@ details.collapsible:not([open]) > .collapsible-control {
 // Collapsibles
 
 .collapsible.collapsed:not(details) {
-  overflow: hidden;
+  overflow-y: clip;
```
lippserd, line 388: Is there a reason why you chose `clip` instead of `hidden`? After a quick search, `clip` seems to be better than `hidden` in terms of scrolling, which sounds good, but should be justified in the commit and PR description.

jrauh01: Good point, of course there should absolutely be a justification for that! Indeed there is a strong reason why I use `clip` instead of `hidden`. As soon as you use `hidden` on one axis, **the other axis can no longer remain visible**. This was the reason why the `clip` option was introduced in the first place. To clip overflow of one axis while leaving the other axis unaffected.

This is exactly what this PR should do:
1. Clip of the overflow on the y axis
2. Keep the overflow on the x axis visible

jrauh01:
**Note:** If this PR is merged the part marked for removal in https://github.com/Icinga/icinga-kubernetes-web/pull/167 can be undone.

nilmerg (approved):

PR #5466, opened 2026-01-13
Correct Icinga\Application\Web doc comment

Al2Klimov:
In contrast to EmbeddedWeb, Web is specifically for Icinga Web itself, no "other web projects".


lippserd:
I would prefer to replace the entire description, as both sentences really don't say anything. Since this is a very low priority, I will simply close the PR.

PR #5469, opened 2026-01-21
GitHub Actions: Add `monitoring` as dependency

sukhwinder33445:
Since the `monitoring` module has been separated from `Icingaweb2`, PHPStan fails to load files that refer to the monitoring module. The module must therefore be added as a dependency in `php.yml`.


sukhwinder33445:
No, instead, the PHPStan baseline must be cleaned up.

PR #5419, opened 2025-08-26
Implement password policy with hook

JolienTrog:
Add password policy support for database-backed user accounts. Administrators can select a policy under **Configuration > Application > General**. It is then enforced on every form that sets or resets a password. The account change-password form, the admin user create and edit form, and the setup wizard admin account page.

Two built-in policies ship with this change. **None** (the default) accepts any password without validation. **Common** enforces a baseline: minimum 12 characters, at least one digit, one special character, one lowercase letter, and one uppercase letter. The active policy's requirements are shown as an info callout below the password field so users know what to enter before submitting. A configuration error falls back to an error callout rather than breaking the form.

Third-party modules can provide additional policies by implementing `PasswordPolicyHook`. The new section in `doc/05-Authentication.md` covers the configuration UI and a worked example of a custom policy module.

resolve #4401

PR #5474, opened 2026-02-16
Introduce `HookEssentials`

Al2Klimov:
Introduce `Icinga\Application\Hook\HookEssentials` as the shared API for hook base classes to register and retrieve hook implementations through their own type.

This moves existing core hook classes to methods such as `register()`, `all()`, `first()`, and `isRegistered()`, while keeping the underlying hook registry behavior unchanged.

The new trait centralizes common hook behavior and lets hook base classes define their registry name once through `getHookName()`. It also adds `getModule()` for hook implementations that belong to a module namespace.

Existing direct uses of `Hook::has()`, `Hook::all()`, `Hook::first()`, and `Hook::register()` are replaced where a typed hook base class is available. The old low-level APIs and `Module::provideHook()` are marked as deprecated in favor of `Hook\HookEssentials`.

### DIFF
--- a/library/Icinga/Application/Hook/Essentials.php
+++ b/library/Icinga/Application/Hook/Essentials.php
@@ -1,0 +1,59 @@
+<?php
+
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+namespace Icinga\Application\Hook;
+
+use Icinga\Application\Hook;
+
+/**
+ * All hooks are provided and consumed - this trait provides the mechanism
+ */
+trait Essentials
+{
+    /**
+     * Get the name for {@link Hook::register} and {@link Hook::all}
+     */
+    abstract protected static function getHookName(): string;
+
+    /**
+     * Get whether the hook is registered
+     *
+     * @return bool
+     */
+    public static function isRegistered(): bool
+    {
+        return Hook::has(static::getHookName());
+    }
+
+    /**
+     * Get all instances of the hook
+     *
+     * @return static[]
+     */
+    public static function all(): array
+    {
+        return Hook::all(static::getHookName());
+    }
+
+    /**
+     * Get the first hook if any
+     *
+     * @return ?static
+     */
+    public static function first(): ?static
+    {
+        return Hook::first(static::getHookName());
+    }
+
+    /**
+     * Register a hook provider
+     *
+     * @param bool $alwaysRun Whether to always run the hook, without permission check
+     */
+    public static function register(bool $alwaysRun = false): void
+    {
+        Hook::register(static::getHookName(), static::class, static::class, $alwaysRun);
+    }
+}
--- a/library/Icinga/Application/Hook/HookEssentials.php
+++ b/library/Icinga/Application/Hook/HookEssentials.php
@@ -1,0 +1,61 @@
+<?php
+
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+namespace Icinga\Application\Hook;
+
+use Icinga\Application\Hook;
+
+/**
+ * Provides the mechanism for hook registration and retrieval
+ */
+trait HookEssentials
+{
+    /**
+     * Get the name that identifies this hook type
+     *
+     * @return string
+     */
+    abstract protected static function getHookName(): string;
+
+    /**
+     * Get whether the hook is registered
+     *
+     * @return bool
+     */
+    public static function isRegistered(): bool
+    {
+        return Hook::has(static::getHookName());
+    }
+
+    /**
+     * Get all instances of the hook
+     *
+     * @return static[]
+     */
+    public static function all(): array
+    {
+        return Hook::all(static::getHookName());
+    }
+
+    /**
+     * Get the first hook if any
+     *
+     * @return ?static
+     */
+    public static function first(): ?static
+    {
+        return Hook::first(static::getHookName());
+    }
+
+    /**
+     * Register the hook
+     *
+     * @param bool $alwaysRun Whether to always run the hook, without permission check
+     */
+    public static function register(bool $alwaysRun = false): void
+    {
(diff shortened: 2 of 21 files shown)


Al2Klimov (commented):
# Test

1. Install Fedora 42
2. Apply https://github.com/Al2Klimov/ansible-icinga-fedora
3. Downgrade to https://git.icinga.com/packages/icingaweb/-/jobs/817857
4. Ensure /icingaweb2/health shows Icinga DB and Redis
5. In `/usr/share/icingaweb2/modules/icingadb/run.php`, comment out `$this->provideHook('health', 'RedisHealth');`
6. Ensure /icingaweb2/health shows Icinga DB only
7. In `/usr/share/icingaweb2/modules/icingadb/run.php`, add `\Icinga\Module\Icingadb\ProvidedHook\RedisHealth::register();`
8. Ensure /icingaweb2/health shows Icinga DB and Redis again = it works

Al2Klimov:
@lippserd Please have a quick look here. This PR would allow to make

* ~#5442~
* #5419
* ~#5433~
* #4774
* #5477
* and #5337

gleich gescheit.

library/Icinga/Application/Hook/Essentials.php
```diff
@@ -0,0 +1,57 @@
+<?php
+/* Icinga Web 2 | (c) 2026 Icinga GmbH | GPLv2+ */
```
sukhwinder33445: Please replace:
```suggestion
// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
// SPDX-License-Identifier: GPL-3.0-or-later
```

library/Icinga/Application/Hook/Essentials.php
```diff
@@ -0,0 +1,57 @@
...
+     * Get the name for {@link Hook::register} and {@link Hook::all}
+     */
+    abstract protected static function getHookName(): string;
+
+    /**
+     * Whether someone registered itself for the hook
+     *
+     * @return bool
+     */
+    public static function registered(): bool
```
sukhwinder33445: I would prefer `isRegistred` as method name.

library/Icinga/Application/Hook/Essentials.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        return Hook::has(static::getHookName());
+    }
+
+    /**
+     * Return all instances of the hook
```
sukhwinder33445: Get all instances of the hook

library/Icinga/Application/Hook/Essentials.php
```diff
@@ -0,0 +1,57 @@
...
+ */
+trait Essentials
+{
+    /**
+     * Get the name for {@link Hook::register} and {@link Hook::all}
+     */
+    abstract protected static function getHookName(): string;
+
+    /**
+     * Whether someone registered itself for the hook
```
sukhwinder33445: someone? `Get whether the hook is registred` fits better.

sukhwinder33445 (changes requested):
Please rebase.

sukhwinder33445 (dismissed):

lippserd:
I made the following [changes](https://github.com/Icinga/icingaweb2/compare/f4df525b8e97abd173dda762b0bb59d0f0ae775c..eed491fee2415a4569f914e896578dc228f48486):

- Renamed `Essentials` to `HookEssentials` throughout
- Replaced the trait-alias pattern in `LoginButtonHook` and `RequestHook` (`use Essentials { register as protected parentRegister; }` + `register()` override) with an `isAlwaysRun(): bool` class method override; `register()` now accepts `?bool $alwaysRun = null` and falls back to `static::isAlwaysRun()`
- Added `getModule(): ?Module`, ported from Icinga DB Web's `HookUtils`, generalized to return `null` for hooks outside a module namespace
- Completed two missed migrations: `Hook::first('Pdfexport')` in `Pdf.php` and `Hook::first('ThemeLoader')` in `StyleSheet.php`

@Al2Klimov please have a look since I have amended some of the changes into your commits.
@TheSyscall please review.

library/Icinga/Application/Hook/HookEssentials.php
```diff
@@ -0,0 +1,98 @@
...
+    public function getModule(): ?Module
+    {
+        if (! ClassLoader::classBelongsToModule(static::class)) {
+            return null;
+        }
+
+        $moduleName = ClassLoader::extractModuleName(static::class);
+
+        return Icinga::app()->getModuleManager()->getModule($moduleName);
+    }
```
TheSyscall, line 71: A module developer could just return an arbitrary module here, so we are forced to check it with our own implementation anyway. Maybe it would be a good idea to document this fact. If not here then in some developer guideline.

TheSyscall (changes requested):
- I searched through the codebase and I can't find any hook that doesn't use the new HookEssentials methods. (At least in icingaweb itself)
- Tested the functionality behind these hooks, and they seem to work just fine.
- `getModule` should probably be final, but that is sadly not possible in PHP.

Finding(s):

`getHookName` inside the hook implementations should be made final in all instances, overriding it could lead to strange bugs. But that would be entirely self-inflicted.

library/Icinga/Application/Hook/HookEssentials.php
Al2Klimov: > final, but that is sadly not possible in PHP

Why not?

TheSyscall: That is a question for the PHP developers.

PHP 8.2 added the ability for traits to define constants, but methods in a class always override the one in the trait. Which makes overriding trait methods always possible. Methods in traits act more like a default implementation in that sense.

https://symfony-exam.com/blog/php-74-declare-a-final-method-in-a-trait-in-php

Al2Klimov: > A module developer could just return an arbitrary module here

An IW2 module is like an npm package or an AUR repo: if you install some crap you didn't check, you're screwed.

> Maybe it would be a good idea to document this fact

The ability to override methods isn't anything fundamentally new.

> getHookName inside the hook implementations should be made final in all instances

Done.

TheSyscall (dismissed):
LGTM 👍

lippserd (approved):

PR #5480, opened 2026-03-23
Add `ConfigForm` based on `CompatForm`

TheSyscall:
An implementation of an INI based configuration form for CompatForms.

It allows developers to create a configuration form for an INI file and have it automatically populate form elements and store the results back into the specified file or section.

If writing the configuration file fails, an error is displayed alongside the full contents of the file to copy and paste manually by the admin.

As suggested [here](https://github.com/Icinga/icingaweb2-module-pdfexport/pull/88/changes/BASE..e42ceeb7b1a17aef15af59b6adeb3efaded2c518#r2965155112)

resolves #5479

### DIFF
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -8,6 +8,7 @@ namespace Icinga\Web\Form;
 use Exception;
 use Icinga\Application\Config;
 use Icinga\Web\Widget\ShowConfiguration;
+use ipl\Stdlib\Str;
 use ipl\Web\Compat\CompatForm;
 
 /**
@@ -116,7 +117,7 @@ class ConfigForm extends CompatForm
 
             $value = $this->getPopulatedValue($element->getName());
             $configSection = $this->config->getSection($section);
-            if ((string) $value === '') {
+            if (Str::isEmpty($value)) {
                 unset($configSection[$key]);
             } else {
                 $configSection->$key = $value;
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -7,6 +7,7 @@ namespace Icinga\Web\Form;
 
 use Exception;
 use Icinga\Application\Config;
+use Icinga\Exception\ProgrammingError;
 use Icinga\Web\Widget\ShowConfiguration;
 use ipl\Web\Compat\CompatForm;
 
@@ -65,6 +66,10 @@ class ConfigForm extends CompatForm
      */
     protected function populateFromConfig(): void
     {
+        if (! $this->config) {
+            throw new ProgrammingError("A config object must be set before populating the form.");
+        }
+
         $populate = [];
         foreach ($this->getElements() as $element) {
             [$section, $key] = $this->getIniKeyFromName($element->getName());
@@ -108,15 +113,17 @@ class ConfigForm extends CompatForm
      */
     public function getConfigValue(string $name, mixed $default = null): mixed
     {
-        if (! $this->hasElement($name)) {
-            return $default;
-        }
-
-        if (($value = $this->getPopulatedValue($name)) !== null) {
-            return $value;
-        }
-
         [$section, $key] = $this->getIniKeyFromName($name);
+        if ($section !== null && $key !== null) {
+            if (! $this->hasElement($name)) {
+                return $default;
+            }
+
+            if (($value = $this->getPopulatedValue($name)) !== null) {
+                return $value;
+            }
+        }
+
         return $this->config->get($section, $key, $default);
     }
 
@@ -130,6 +137,10 @@ class ConfigForm extends CompatForm
      */
     protected function save(): void
     {
+        if (! $this->config) {
+            throw new ProgrammingError("A config object must be set before saving the configuration.");
+        }
+
         foreach ($this->getElements() as $element) {
             if (in_array($element->getName(), $this->ignoredElements)) {
                 continue;
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -38,6 +38,7 @@ class ConfigForm extends CompatForm
     {
         if (! $this->hasBeenAssembled) {
             parent::ensureAssembled();
+            $this->addRequiredElements();
             $this->populateFromConfig();
         }
 
@@ -151,12 +152,7 @@ class ConfigForm extends CompatForm
         }
     }
 
-    /**
-     * Add the store button to the form
-     *
-     * @return void
-     */
-    protected function addButtonElements(): void
+    protected function addRequiredElements(): void
     {
         $this->addElement('submit', static::SUBMIT_BUTTON_NAME, [
             'label' => $this->translate('Store'),
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -65,16 +65,19 @@ class ConfigForm extends CompatForm
      */
     protected function populateFromConfig(): void
     {
+        $populate = [];
         foreach ($this->getElements() as $element) {
             [$section, $key] = $this->getIniKeyFromName($element->getName());
             if ($section === null || $key === null) {
                 continue;
             }
             $value = $this->getPopulatedValue($element->getName()) ?? $this->config->get($section, $key);
-            $this->populate([
-                $element->getName() => $value,
-            ]);
+            if ($value === null) {
+                continue;
+            }
+            $populate[$element->getName()] = $value;
         }
+        $this->populate($populate);
     }
 
     /**
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -241,10 +241,6 @@ class ConfigForm extends CompatForm
         }
 
         [$section, $key] = $this->getIniKeyFromName($name);
-        if (! $this->config->hasSection($section)) {
-            return $default;
-        }
-
         return $this->config->get($section, $key, $default);
     }
 
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -107,21 +107,15 @@ class ConfigForm extends CompatForm
      * Get the value of a configuration key from an element name
      *
      * @param string $name The element name
-     * @param mixed $default The default value to return if the element does not exist or the value is empty
+     * @param mixed $default The default value to return if the config entry does not exist
      *
      * @return mixed The value of the configuration key or the default value
      */
     public function getConfigValue(string $name, mixed $default = null): mixed
     {
         [$section, $key] = $this->getIniKeyFromName($name);
-        if ($section !== null && $key !== null) {
-            if (! $this->hasElement($name)) {
-                return $default;
-            }
-
-            if (($value = $this->getPopulatedValue($name)) !== null) {
-                return $value;
-            }
+        if ($section === null || $key === null) {
+            return $default;
         }
 
         return $this->config->get($section, $key, $default);
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -18,12 +18,15 @@ class ConfigForm extends CompatForm
     /** @var string Name of the submit button element */
     protected const SUBMIT_BUTTON_NAME = 'store';
 
+    /** @var string Delimiter used to separate the section and key in the element name */
+    protected string $sectionKeyDelimiter = '__';
+
     /**
      * A list of elements that should not be saved to the configuration
      *
      * @var string[]
      */
-    protected array $ignoredElements = [self::SUBMIT_BUTTON_NAME];
+    protected array $ignoredElements = [];
 
     public function __construct(
         protected Config $config,
@@ -71,7 +74,7 @@ class ConfigForm extends CompatForm
      */
     protected function getIniKeyFromName(string $name): ?array
     {
-        $parts = explode('__', $name, 2);
+        $parts = explode($this->sectionKeyDelimiter, $name, 2);
 
         return count($parts) === 2 ? $parts : null;
     }
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -1,0 +1,195 @@
+<?php
+/* Icinga Web 2 | (c) 2026 Icinga GmbH | GPLv2+ */
+
+namespace Icinga\Web\Form;
+
+use Exception;
+use Icinga\Application\Config;
+use Icinga\Web\Widget\ShowConfiguration;
+use ipl\Web\Compat\CompatForm;
+
+/**
+ * Form base-class providing standard functionality for configuration forms
+ */
+class ConfigForm extends CompatForm
+{
+    /**
+     * A list of elements that should not be saved to the configuration
+     *
+     * @var string[]
+     */
+    protected array $ignoredElements = [];
+
+    /**
+     * The configuration to work with
+     *
+     * @var Config|null
+     */
+    protected ?Config $config = null;
+
+    /**
+     * The section to work with.
+     * If not set, the section is determined from the element name.
+     *
+     * @var string|null
+     */
+    protected ?string $section = null;
+
+    /**
+     * Set the configuration to use when populating and saving
+     *
+     * @param Config $config The configuration to use
+     *
+     * @return $this
+     */
+    public function setConfig(Config $config): static
+    {
+        $this->config = $config;
+
+        return $this;
+    }
+
+    /**
+     * Set the section to use when populating and saving
+     *
+     * @param string $section The section to use
+     *
+     * @return $this
+     */
+    public function setSection(string $section): static
+    {
+        $this->section = $section;
+
+        return $this;
+    }
+
+    public function ensureAssembled(): static
+    {
+        if (! $this->hasBeenAssembled) {
+            parent::ensureAssembled();
+            $this->populateFromConfig();
+        }
+
+        return $this;
+    }
+
+    /**
+     * Populate the form elements from the configuration
+     *
+     * @return void
+     */
+    protected function populateFromConfig(): void
+    {
+        foreach ($this->getElements() as $element) {
+            [$section, $key] = $this->getIniKeyFromName($element->getName());
+            if ($section === null || $key === null) {
+                continue;
+            }
+            $value = $this->getPopulatedValue($element->getName()) ?? $this->config->get($section, $key);
+            $this->populate([
+                $element->getName() => $value,
+            ]);
+        }
+    }
+
+    /**
+     * Get the section and key from the element name.
+     * If `$this->section` it is used as the section name, with the key being the element name.
+     * Otherwise, the section is determined from the element name.
+     *
+     * @param string $name The element name
+     *
+     * @return string[]|null
+     */
+    protected function getIniKeyFromName(string $name): ?array
+    {
+        if ($this->section !== null) {
+            return [$this->section, $name];
+        }
+
+        $parts = explode('__', $name, 2);
+        if (count($parts) !== 2) {
+            return null;
+        }
+
+        return $parts;
+    }
+
+    /**
+     * Get the value of a configuration key from an element name
+     *
+     * @param string $name The element name
+     * @param mixed $default The default value to return if the element does not exist or the value is empty
+     *
+     * @return mixed The value of the configuration key or the default value
+     */
+    public function getConfigValue(string $name, mixed $default = null): mixed
+    {
+        if (! $this->hasElement($name)) {
+            return $default;
+        }
+
+        if (($value = $this->getPopulatedValue($name)) !== null) {
+            return $value;
+        }
+
+        [$section, $key] = $this->getIniKeyFromName($name);
+        if (! $this->config->hasSection($section)) {
+            return $default;
+        }
+
+        return $this->config->get($section, $key, $default);
+    }
+
+    /**
+     * Persist the current configuration to disk
+     *
+     * If an error occurs, the form will be re-rendered with the error message and the raw INI configuration.
+     */
+    protected function save(): void
+    {
+        foreach ($this->getElements() as $element) {
+            if (in_array($element->getName(), $this->ignoredElements)) {
+                continue;
+            }
+            [$section, $key] = $this->getIniKeyFromName($element->getName());
+            if ($section === null || $key === null) {
+                continue;
+            }
+            $value = $this->getConfigValue($element->getName());
+
+            $configSection = $this->config->getSection($section);
+            if (empty($value)) {
+                unset($configSection[$key]);
+            } else {
+                $configSection->$key = $value;
+            }
+
+            if ($configSection->isEmpty()) {
+                $this->config->removeSection($section);
+            } else {
+                $this->config->setSection($section, $configSection);
+            }
+        }
+
+        $this->config->saveIni();
+    }
+
+    protected function onSuccess(): void
+    {
+        try {
+            $this->save();
+        } catch (Exception $e) {
+            $content = $this->getContent();
+            array_unshift(
+                $content,
+                new ShowConfiguration(
+                    $e,
+                    $this->config,
+                )
+            );
+            $this->setContent($content);
+            throw $e;
+        }
+    }
+}
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -22,13 +22,6 @@ class ConfigForm extends CompatForm
     /** @var string Delimiter used to separate the section and key in the element name */
     protected string $sectionKeyDelimiter = '__';
 
-    /**
-     * A list of elements that should not be saved to the configuration
-     *
-     * @var string[]
-     */
-    protected array $ignoredElements = [];
-
     public function __construct(
         protected Config $config,
     ) {
@@ -107,16 +100,12 @@ class ConfigForm extends CompatForm
      */
     protected function save(): void
     {
-        foreach ($this->getElements() as $element) {
-            if (in_array($element->getName(), $this->ignoredElements)) {
-                continue;
-            }
-            if (($parts = $this->getIniKeyFromName($element->getName())) === null) {
+        foreach ($this->getValues() as $element => $value) {
+            if (($parts = $this->getIniKeyFromName($element)) === null) {
                 continue;
             }
             [$section, $key] = $parts;
 
-            $value = $this->getPopulatedValue($element->getName());
             $configSection = $this->config->getSection($section);
             if (Str::isEmpty($value)) {
                 unset($configSection[$key]);
@@ -156,6 +145,7 @@ class ConfigForm extends CompatForm
     {
         $this->addElement('submit', static::SUBMIT_BUTTON_NAME, [
             'label' => $this->translate('Store'),
+            'ignore' => true,
         ]);
     }
 }
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -138,10 +138,10 @@ class ConfigForm extends CompatForm
             if ($section === null || $key === null) {
                 continue;
             }
-            $value = $this->getConfigValue($element->getName());
 
+            $value = $this->getPopulatedValue($element->getName());
             $configSection = $this->config->getSection($section);
-            if (empty($value)) {
+            if ((string) $value === '') {
                 unset($configSection[$key]);
             } else {
                 $configSection->$key = $value;
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -49,10 +49,10 @@ class ConfigForm extends CompatForm
     {
         $populate = [];
         foreach ($this->getElements() as $element) {
-            [$section, $key] = $this->getIniKeyFromName($element->getName());
-            if ($section === null || $key === null) {
+            if (($parts = $this->getIniKeyFromName($element->getName())) === null) {
                 continue;
             }
+            [$section, $key] = $parts;
             $value = $this->getPopulatedValue($element->getName()) ?? $this->config->get($section, $key);
             if ($value === null) {
                 continue;
@@ -67,17 +67,13 @@ class ConfigForm extends CompatForm
      *
      * @param string $name The element name
      *
-     * @return string[]|null
+     * @return array{string, string}|null
      */
     protected function getIniKeyFromName(string $name): ?array
     {
         $parts = explode('__', $name, 2);
 
-        if (count($parts) !== 2) {
-            return [null, null];
-        }
-
-        return $parts;
+        return count($parts) === 2 ? $parts : null;
     }
 
     /**
@@ -90,10 +86,10 @@ class ConfigForm extends CompatForm
      */
     public function getConfigValue(string $name, mixed $default = null): mixed
     {
-        [$section, $key] = $this->getIniKeyFromName($name);
-        if ($section === null || $key === null) {
+        if (($parts = $this->getIniKeyFromName($name)) === null) {
             return $default;
         }
+        [$section, $key] = $parts;
 
         return $this->config->get($section, $key, $default);
     }
@@ -110,10 +106,10 @@ class ConfigForm extends CompatForm
             if (in_array($element->getName(), $this->ignoredElements)) {
                 continue;
             }
-            [$section, $key] = $this->getIniKeyFromName($element->getName());
-            if ($section === null || $key === null) {
+            if (($parts = $this->getIniKeyFromName($element->getName())) === null) {
                 continue;
             }
+            [$section, $key] = $parts;
 
             $value = $this->getPopulatedValue($element->getName());
             $configSection = $this->config->getSection($section);
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -60,6 +60,8 @@ class ConfigForm extends CompatForm
      * Populate the form elements from the configuration
      *
      * @return void
+     *
+     * @throws ProgrammingError
      */
     protected function populateFromConfig(): void
     {
@@ -76,7 +78,7 @@ class ConfigForm extends CompatForm
     }
 
     /**
-     * Get the section and key from the element name.
+     * Get the section and key from the element name
      *
      * @param string $name The element name
      *
@@ -118,7 +120,10 @@ class ConfigForm extends CompatForm
     /**
      * Persist the current configuration to disk
      *
-     * If an error occurs, the form will be re-rendered with the error message and the raw INI configuration.
+     * If an error occurs, the form will be re-rendered with the error message
+     * and the raw INI configuration.
+     *
+     * @throws ProgrammingError
      */
     protected function save(): void
     {
@@ -168,7 +173,7 @@ class ConfigForm extends CompatForm
     }
 
     /**
-     * Add the store button to the form.
+     * Add the store button to the form
      *
      * @return void
      */
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -1,0 +1,217 @@
+<?php
+
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+namespace Icinga\Web\Form;
+
+use Exception;
+use Icinga\Application\Config;
+use ipl\Html\Contract\FormElement;
+use ipl\Html\Contract\ValueCandidates;
+use ipl\Html\HtmlElement;
+use ipl\Html\ValidHtml;
+use ipl\Stdlib\Str;
+use ipl\Web\Compat\CompatForm;
+use ipl\Web\Compat\DisplayFormElement;
+use ipl\Web\Widget\CopyToClipboard;
+use LogicException;
+
+/**
+ * Form base-class providing standard functionality for configuration forms
+ *
+ *  Element names follow a `section__key` convention: the part before the
+ *  {@see $sectionKeyDelimiter} (default: `__`) is the INI section name and the
+ *  part after is the configuration key within that section. Subclasses add
+ *  elements whose names match this pattern; `populateFromConfig` and `save`
+ *  use it to map form values to and from the INI file automatically.
+ */
+class ConfigForm extends CompatForm
+{
+    /** @var string Name of the submit button element */
+    protected const SUBMIT_BUTTON_NAME = 'store';
+
+    /**
+     * Delimiter used to separate the section and key in the element name
+     *
+     * This is used to determine whether an element is a configuration key or a
+     * section. The default delimiter is '__', this is chosen to allow for section
+     * names that contain underscores.
+     *
+     * @var string
+     */
+    protected string $sectionKeyDelimiter = '__';
+
+    /**
+     * Create a new configuration form
+     *
+     * @param Config $config The ini file configuration object to use for the form
+     */
+    public function __construct(
+        protected Config $config,
+    ) {
+        $this->on(static::ON_ELEMENT_REGISTERED, function (FormElement $element) {
+            [$section, $key] = Str::symmetricSplit($element->getName(), $this->sectionKeyDelimiter, 2);
+            if ($key === null || $element->hasValue()) {
+                return;
+            }
+
+            $configValue = $this->config->get($section, $key);
+            if ($configValue === null) {
+                return;
+            }
+
+            if (! ($element instanceof ValueCandidates)) {
+                $element->setValue($configValue);
+
+                return;
+            }
+
+            $candidates = $element->getValueCandidates();
+            if (empty($candidates)) {
+                // Initial render: no prior submission, set value and seed candidates
+                $element->setValue($configValue);
+                $element->setValueCandidates([$configValue]);
+            } else {
+                // POST: candidates were set by registerElement; append config value so
+                // PasswordElement's (count > 1) condition resolves DUMMYPASSWORD on re-render
+                $element->setValueCandidates(array_merge($candidates, [$configValue]));
+            }
+        });
+    }
+
+    /**
+     * Ensure that all required form elements are present
+     *
+     * @return $this
+     */
+    public function ensureAssembled(): static
+    {
+        if (! $this->hasBeenAssembled) {
+            parent::ensureAssembled();
+            $this->addRequiredElements();
+        }
+
+        return $this;
+    }
+
+    /**
+     * Create a hint that explains a configuration save failure and how to fix it manually
+     *
+     * @param Exception $exception The exception thrown while saving
+     * @param Config $config The configuration that failed to save
+     *
+     * @return ValidHtml
+     */
+    public static function createConfigurationErrorHint(Exception $exception, Config $config): ValidHtml
+    {
+        $code = HtmlElement::create('code', null, (string) $config);
+        CopyToClipboard::attachTo($code);
+
+        return HtmlElement::create('div', null, [
+            HtmlElement::create('h4', null, t('Saving Configuration Failed!')),
+            HtmlElement::create('p', null, [
+                sprintf(
+                    t("The file %s couldn't be stored. (Error: '%s')"),
+                    $config->getConfigFile(),
+                    $exception->getMessage(),
+                ),
+                HtmlElement::create('br'),
+                t('This could have one or more of the following reasons:'),
+            ]),
+            HtmlElement::create('ul', null, [
+                HtmlElement::create('li', null, t("You don't have file-system permissions to write to the file")),
+                HtmlElement::create('li', null, t('Something went wrong while writing the file')),
+                HtmlElement::create('li', null, t(
+                    "There's an application error preventing you from persisting the configuration",
+                )),
+            ]),
+            HtmlElement::create('p', null, [
+                t(
+                    'Details can be found in the application log. ' .
+                    "(If you don't have access to this log, call your administrator in this case)",
+                ),
+                HtmlElement::create('br'),
+                t('In case you can access the file by yourself, you can open it and insert the config manually:'),
+            ]),
+            HtmlElement::create('p', null, HtmlElement::create('pre', null, $code)),
+        ]);
+    }
+
+    /**
+     * Persist the current configuration to disk
+     *
+     * If an error occurs, the form will be re-rendered with the error message
+     * and the raw INI configuration.
+     *
+     * @return void
+     *
+     * @throws LogicException If an array value is encountered in the form
+     */
+    protected function save(): void
+    {
+        foreach ($this->getValues() as $element => $value) {
+            [$section, $key] = Str::symmetricSplit($element, $this->sectionKeyDelimiter, 2);
+            if ($key === null) {
+                continue;
+            }
+
+            if (is_array($value)) {
+                throw new LogicException(sprintf('Cannot save element "%s": array values are not supported', $element));
+            }
+
+            $configSection = $this->config->getSection($section);
+            if (Str::isEmpty($value)) {
+                unset($configSection[$key]);
+            } else {
+                $configSection[$key] = $value;
+            }
+
+            if ($configSection->isEmpty()) {
+                $this->config->removeSection($section);
+            } else {
+                $this->config->setSection($section, $configSection);
+            }
+        }
+
+        $this->config->saveIni();
+    }
+
+    /**
+     * Handle the form submission
+     *
+     * If the form submission is successful, the configuration is saved to disk.
+     * If an error occurs, the form is re-rendered with the error message and the
+     * raw INI configuration. The original exception is re-thrown.
+     *
+     * @return void
+     */
+    protected function onSuccess(): void
+    {
+        try {
+            $this->save();
+        } catch (Exception $e) {
+            $content = $this->getContent();
+            array_unshift($content, new DisplayFormElement(static::createConfigurationErrorHint($e, $this->config)));
+            $this->setContent($content);
+
+            throw $e;
+        }
+    }
+
+    /**
+     * Add the submit button to the form
+     *
+     * Called automatically during assembly. Subclasses may override this to add
+     * additional required elements but must call the parent implementation.
+     *
+     * @return void
+     */
+    protected function addRequiredElements(): void
+    {
+        $this->addElement('submit', static::SUBMIT_BUTTON_NAME, [
+            'label' => $this->translate('Store'),
+            'ignore' => true,
+        ]);
+    }
+}
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -1,5 +1,7 @@
 <?php
-/* Icinga Web 2 | (c) 2026 Icinga GmbH | GPLv2+ */
+
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
 
 namespace Icinga\Web\Form;
 
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -87,7 +87,7 @@ class ConfigForm extends CompatForm
         $parts = explode('__', $name, 2);
 
         if (count($parts) !== 2) {
-            return null;
+            return [null, null];
         }
 
         return $parts;
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -87,7 +87,7 @@ class ConfigForm extends CompatForm
      *
      * @return mixed The value of the configuration key or the default value
      */
-    public function getConfigValue(string $name, mixed $default = null): mixed
+    protected function getConfigValue(string $name, mixed $default = null): mixed
     {
         if (($parts = $this->getIniKeyFromName($name)) === null) {
             return $default;
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -7,10 +7,7 @@ namespace Icinga\Web\Form;
 
 use Exception;
 use Icinga\Application\Config;
-use Icinga\Exception\ProgrammingError;
 use Icinga\Web\Widget\ShowConfiguration;
-use ipl\Html\Contract\FormSubmitElement;
-use ipl\Validator\CallbackValidator;
 use ipl\Web\Compat\CompatForm;
 
 /**
@@ -18,24 +15,15 @@ use ipl\Web\Compat\CompatForm;
  */
 class ConfigForm extends CompatForm
 {
-    /** @var string Event emitted when the form has successfully deleted a configuration section */
-    public const ON_DELETE = 'delete';
-
     /** @var string Name of the submit button element */
     protected const SUBMIT_BUTTON_NAME = 'store';
-
-    /** @var string Name of the delete button element */
-    protected const DELETE_BUTTON_NAME = 'delete';
-
-    /** @var string Name of the element containing the section name */
-    protected const NAME_ELEMENT_NAME = 'name';
 
     /**
      * A list of elements that should not be saved to the configuration
      *
      * @var string[]
      */
-    protected array $ignoredElements = ['name', 'store', 'delete'];
+    protected array $ignoredElements = [self::SUBMIT_BUTTON_NAME];
 
     /**
      * The configuration to work with
@@ -43,49 +31,6 @@ class ConfigForm extends CompatForm
      * @var Config|null
      */
     protected ?Config $config = null;
-
-    /**
-     * The section to work with.
-     * If not set, the section is determined from the element name.
-     *
-     * @var string|null
-     */
-    protected ?string $section = null;
-
-    /**
-     * Whether the form is used for creating a new configuration section
-     *
-     * @var bool
-     */
-    protected bool $isCreateForm = false;
-
-    /**
-     * Whether the form allows deletion of the configuration section
-     *
-     * @var bool
-     */
-    protected bool $allowDeletion = true;
-
-    public function __construct()
-    {
-        $this->on(static::ON_SENT, function () {
-            if ($this->shouldDelete()) {
-                $this->handleDelete();
-                $this->emit(static::ON_DELETE, [$this]);
-            }
-        });
-    }
-
-    public function isValidEvent($event): bool
-    {
-        // Check for our new event and return true if it is valid
-        if ($event === static::ON_DELETE) {
-            return true;
-        }
-
-        // Call the parent function to still validate all previous added events
-        return parent::isValidEvent($event);
-    }
 
     /**
      * Set the configuration to use when populating and saving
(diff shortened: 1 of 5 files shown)


library/Icinga/Web/Widget/ShowConfiguration.php
```diff
@@ -0,0 +1,96 @@
+<?php
+/* Icinga Web 2 | (c) 2026 Icinga GmbH | GPLv2+ */
```
Al2Klimov: See #5486.

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,426 @@
+<?php
+/* Icinga Web 2 | (c) 2026 Icinga GmbH | GPLv2+ */
```
Al2Klimov: See #5486.

library/Icinga/Web/Widget/ShowConfiguration.php
```diff
@@ -0,0 +1,96 @@
...
+
+    protected function assemble(): void
+    {
+        $this->addHtml(HtmlElement::create(
+            'h4',
+            null,
+            t('Saving Configuration Failed!'),
+        ));
+
+        $this->addHtml(HtmlElement::create(
```
Al2Klimov: I wonder whether multiple individual HtmlElements are necessary...

library/Icinga/Web/Widget/ShowConfiguration.php
```diff
@@ -0,0 +1,96 @@
...
+            'p',
+            null,
+            [
+                sprintf(
+                    t("The file %s couldn't be stored. (Error: '%s')"),
+                    $this->config->getConfigFile(),
+                    $this->exception->getMessage(),
+                ),
+                HtmlString::create('<br>'),
```
Al2Klimov: I'd use multiple p instead of br.

@flourish86 What do you think?

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,426 @@
...
+    }
+
+    /**
+     * Set whether the form is used for creating a new configuration section, with a name that can be chosen by the user
+     *
+     * @param bool $create
+     *
+     * @return void
+     */
+    public function setIsCreateForm(bool $create = true): void
```
Al2Klimov: Don't we use fluent setters?

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,426 @@
...
+     * If `$this->section` it is used as the section name, with the key being the element name.
+     * Otherwise, the section is determined from the element name.
+     *
+     * @param string $name The element name
+     *
+     * @return string[]|null
+     */
+    protected function getIniKeyFromName(string $name): ?array
+    {
+        $parts = explode('__', $name, 2);
```
Al2Klimov: Strange, the OG ConfigForm uses _, you use __.

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,426 @@
...
+        if (! $this->hasElement($name)) {
+            return $default;
+        }
+
+        if (($value = $this->getPopulatedValue($name)) !== null) {
+            return $value;
+        }
+
+        [$section, $key] = $this->getIniKeyFromName($name);
+        if (! $this->config->hasSection($section)) {
```
Al2Klimov: I wonder if this check is necessary.

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,426 @@
...
+                continue;
+            }
+            [$section, $key] = $this->getIniKeyFromName($element->getName());
+            if ($section === null || $key === null) {
+                continue;
+            }
+            $value = $this->getConfigValue($element->getName());
+
+            $configSection = $this->config->getSection($section);
+            if (empty($value)) {
```
Al2Klimov: `php -r 'echo (int) empty("0");'` says "1", so `empty()` seems not the best idea, does it?

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,426 @@
...
+            throw $e;
+        }
+    }
+
+    protected function onSuccess(): void
+    {
+        if ($this->isCreateForm()) {
+            $this->section = $this->getValue(static::NAME_ELEMENT_NAME);
+
+            if (empty($this->section)) {
```
Al2Klimov: I have no idea about this section stuff, but it sounds interesting...

Al2Klimov (dismissed):
# Test

1. Install Fedora 42
2. Apply https://github.com/Al2Klimov/ansible-icinga-fedora
3. `yum remove icingaweb2-module-monitoring`
4. Downgrade to https://git.icinga.com/packages/icingaweb/-/jobs/856109
5. `tar -C /usr/share/icingaweb2/modules -vxzf ~fedora/test5480.tgz`
6. `icingacli mod en test5480`
7. Visit /icingaweb2/test5480
8. Input something
9. Press enter
10. `cat /etc/icingaweb2/modules/test5480/config.ini`
11. Ok
12. `chmod 0440 /etc/icingaweb2/modules/test5480/config.ini`
13. Repeat steps 7-9
14. Fails

Seems to work. So far – so good.

[test5480.tgz](https://github.com/user-attachments/files/26414008/test5480.tgz)

library/Icinga/Web/Widget/ShowConfiguration.php
TheSyscall: `ShowConfiguration` perfectly mimics the old `showConfiguration.phtml`. (Except adding the copy to clipboard button)

library/Icinga/Web/Widget/ShowConfiguration.php
TheSyscall: `ShowConfiguration` perfectly mimics the old `showConfiguration.phtml`. (Except adding the copy to clipboard button)

But yes, I'm totally open for a new design and/layout and an HTML rework.

library/Icinga/Web/Form/ConfigForm.php
TheSyscall: Yes that was mentioned in #5479 because the original ConfigForm does not support section names that contain `_` which is quite common. I think it's a small improvement, because you still can't have anything containing `__` but that should be quite rare.

library/Icinga/Web/Form/ConfigForm.php
TheSyscall: It was not. Changed it!

library/Icinga/Web/Form/ConfigForm.php
TheSyscall: I have no problems with storing "0" values in the configuration.
Since it could be any type of element here, I don't think there is a reliable way determine if the field should be included or not.

library/Icinga/Web/Form/ConfigForm.php
TheSyscall: It is pretty common for configuration files that manage multiple identical instances of something (resources, users, pdf print backends, etc.).

So this new ConfigForm supports being pointed to a specific section and handling just that. (updaing and deleting)
It also can be used to create a new section.

An alternative way to having all three possible use cases in one form would be to split it up into two. Something like `ConfigForm` for the whole INI-file and `ConfigSectionForm` That focuses just one section, possibly creating or deleting it.

library/Icinga/Web/Widget/ShowConfiguration.php
TheSyscall: Changed

library/Icinga/Web/Form/ConfigForm.php
TheSyscall: Changed

library/Icinga/Web/Form/ConfigForm.php
TheSyscall: Changed

library/Icinga/Web/Form/ConfigForm.php
Al2Klimov: Kind of a reference implementation would be helpful for testing.

TheSyscall: There already is an implementation this class in "section mode" in https://github.com/Icinga/icingaweb2-module-pdfexport/pull/88

TheSyscall: The longer I think about it, the more I like the idea of having two separate forms for the two purposes.

- `ConfigForm`: Represents the whole configuration file, like a modules `config.ini`
- `ConfigSectionForm`: Manages a configuration file section, where each section is an instance of something. Like `resources.ini`

This would reduce the confusing naming requirements and the need for dynamic checks if the form is in section mode.
The delete mode doesn't need to be in the `ConfigForm` at all since it can only affect a specific section.

library/Icinga/Web/Widget/ShowConfiguration.php
flourish86: `<p>` vs. `<br>`: It doesn't really matter in my opinion if it looks the same in the browser. Less HTML-Elements result in less overhead, so I personally would prefer the `br` version.

flourish86 (dismissed):

library/Icinga/Web/Form/ConfigForm.php
Al2Klimov: Currently, your empty() doesn't store "0" values. Is this intentional?

TheSyscall: What would be a better check here? `$value === null || $value === ''`?

Al2Klimov: I personally would cast to string and compare to ''.

```
php > echo (string) false;
php > echo (string) null;
php > echo (string) '';
php > echo (string) 0;
0
php >
```

TheSyscall: Seems to work great

Al2Klimov: There are still empty() calls here.

TheSyscall: Removed empty calls

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,164 @@
...
+
+use Exception;
+use Icinga\Application\Config;
+use Icinga\Web\Widget\ShowConfiguration;
+use ipl\Web\Compat\CompatForm;
+
+/**
+ * Form base-class providing standard functionality for configuration forms
+ */
+class ConfigForm extends CompatForm
```
Al2Klimov, line 29: # Test

1. Install Fedora 42
2. Apply https://github.com/Al2Klimov/ansible-icinga-fedora **using snapshot packages**
3. `yum remove icingaweb2-module-monitoring`
4. Upgrade to https://git.icinga.com/packages/icingaweb/-/jobs/905522
5. Create `/usr/share/icingaweb2/modules/test5480`
6. `icingacli mod en test5480`
7. Visit /icingaweb2/test5480
8. Input something
9. `Press enter`
10. `cat /etc/icingaweb2/modules/test5480/config.ini` (ok)
11. `chmod 0440 /etc/icingaweb2/modules/test5480/config.ini`
12. Repeat steps 7-9 (fails, prints helpful message)

So far so good...

## `/usr/share/icingaweb2/modules/test5480`

### `library/Test5480/Form.php`

```php
<?php

namespace Icinga\Module\Test5480;

use Icinga\Web\Form\ConfigForm;

class Form extends ConfigForm
{
    protected function assemble()
    {
        $this->addElement('text', 'foo__bar');
    }
}
```

### `application/controllers/IndexController.php`

```php
<?php

namespace Icinga\Module\Test5480\Controllers;

use GuzzleHttp\Psr7\ServerRequest;
use Icinga\Module\Test5480\Form;
use ipl\Web\Compat\CompatController;

class IndexController extends CompatController
{
    public function indexAction(): void
    {
        $form = new Form($this->Config());
        $form->handleRequest(ServerRequest::fromGlobals());
        $this->addContent($form);
    }
}
```

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,368 @@
...
+     * @var bool
+     */
+    protected bool $isCreateForm = false;
+
+    /**
+     * Whether the form allows deletion of the configuration section
+     *
+     * @var bool
+     */
+    protected bool $allowDeletion = true;
```
Al2Klimov: Seems to work.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,368 @@
...
+
+    /**
+     * Whether the form allows renaming of the configuration section
+     *
+     * @var bool
+     */
+    protected bool $allowRename = true;
```
Al2Klimov: Seems to work.

Al2Klimov (commented):
Tested with https://github.com/Al2Klimov/icingaweb2-module-test5480.

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,164 @@
...
+
+    /**
+     * Get the value of a configuration key from an element name
+     *
+     * @param string $name The element name
+     * @param mixed $default The default value to return if the config entry does not exist
+     *
+     * @return mixed The value of the configuration key or the default value
+     */
+    public function getConfigValue(string $name, mixed $default = null): mixed
```
Al2Klimov: This is public only for tests, right?

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,368 @@
...
+     * @var string[]
+     */
+    protected array $ignoredElements = [self::SUBMIT_BUTTON_NAME, self::DELETE_BUTTON_NAME, self::NAME_ELEMENT_NAME];
+
+    /**
+     * Whether the form is used for creating a new configuration section
+     *
+     * @var bool
+     */
+    protected bool $isCreateForm = false;
```
Al2Klimov: I wonder whether this property as such is necessary or its value could be inlined.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,368 @@
...
+
+    public function __construct(
+        Config $config,
+        protected ?string $section = null,
+    ) {
+        parent::__construct($config);
+
+        $this->isCreateForm = $section === null;
+
+        $this->on(static::ON_SENT, function () {
```
Al2Klimov: I personally like to outline functions, i.e to make a protected method and later reference it.\*

https://www.php.net/releases/8.1/en.php#first_class_callable_syntax

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,368 @@
...
+    /**
+     * Set whether the form allows deletion of the configuration section
+     *
+     * @param bool $allowDeletion
+     *
+     * @return static
+     */
+    public function setAllowDeletion(bool $allowDeletion = true): static
+    {
+        $this->allowDeletion = $allowDeletion;
```
Al2Klimov: Interesting that you allow this unconditionally, despite this may be a create-form. (See below.)

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,368 @@
...
+    /**
+     * Set the ability to rename the configuration section
+     *
+     * @param bool $allowRename Whether the form is allowed to rename the configuration section
+     *
+     * @return $this
+     */
+    public function setAllowRename(bool $allowRename = true): static
+    {
+        $this->allowRename = $allowRename;
```
Al2Klimov: Same here.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,368 @@
...
+                $validator->addMessage($this->translate('An entry with this name already exists.'));
+                return false;
+            }
+
+            return true;
+        });
+
+        if (! isset($params['validators'])) {
+            $params['validators'] = [];
+        }
```
Al2Klimov:
```suggestion
```

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,368 @@
...
+            return;
+        }
+
+        $params['required'] = true;
+
+        if (! isset($params['label'])) {
+            $params['label'] = $this->translate('Name');
+        }
+
+        $uniqueValidator = new CallbackValidator(function ($value, CallbackValidator $validator) {
```
Al2Klimov: Variable can be inlined.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,368 @@
...
+     * @throws LogicException
+     */
+    protected function handleRename(): void
+    {
+        if ($this->section === null) {
+            throw new LogicException('Section must be set before renaming a configuration section.');
+        }
+
+        $newName = $this->getPopulatedValue(static::NAME_ELEMENT_NAME);
+        $section = $this->config->getSection($this->section);
```
Al2Klimov: Variable $section can be inlined.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,368 @@
...
+            return null;
+        }
+
+        return [$this->section, $name];
+    }
+
+    protected function onSuccess(): void
+    {
+        if ($this->isCreateForm()) {
+            $this->section = $this->getValue(static::NAME_ELEMENT_NAME);
```
Al2Klimov: Sure you want to overwrite a class member in such a function?\*

Al2Klimov (commented):
Whereever I've written a `*` below, consult guidelines whether I'm right and fall back to other people, so that finally there are guidelines.

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: It’s easier to implement a system that returns the correct values as long as we always use the getters for these two attributes. Since `allowDeletion` and `allowRename` default to true, we can’t rely on developers remembering to explicitly disable them when necessary.

There is a check for `isCreateForm` in the `allowDeletion` and `allowRename` methods respectively.

Al2Klimov: What I wanted to express is what about throwing a logic exception?

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: This is done to allow calling the `ConfigForm::onSuccess` and `ConfigForm::save` methods, which in term use the `getIniKeyFromName` method to select the section to save values to.

`ConfigSectionForm::getIniKeyFromName` gets the section name from `ConfigSectionForm::$section`.

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: This is done to keep the value of `SectionConfigForm::isCreateForm()` consistent after form submission. The `ON_SUCCESS` event handler might want to check this field and it shouldn't change because of the internals of `ConfigSectionForm::onSuccess()`

See: https://github.com/Icinga/icingaweb2/pull/5480#discussion_r3309102092

library/Icinga/Web/Form/ConfigForm.php
TheSyscall: Yes. Made protected and adjusted tests

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Good suggestion.

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Added checks to no longer allow changing the value after the form has been assembled as this would break the form submit logic.

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Same as above.

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Removed

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Variable inlined

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Variable inlined

library/Icinga/Web/Form/ConfigSectionForm.php
Al2Klimov: So... no exception, but a silent refusal?

library/Icinga/Web/Form/ConfigSectionForm.php
Al2Klimov: Now you remove the section before getting it. 🙈

TheSyscall: Oops! Fixed

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Added LogicExceptions

Al2Klimov (approved):

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,381 @@
...
+
+    public function __construct(
+        Config $config,
+        protected ?string $section = null,
+    ) {
+        parent::__construct($config);
+
+        $this->isCreateForm = $section === null;
+
+        $this->on(static::ON_SENT, $this->onSent(...));
```
Al2Klimov: # CSRF – ConfigSectionForm section deletion bypasses all validation

## Summary

`ConfigSectionForm` (branch `origin/config-form-5479`) is vulnerable to
Cross-Site Request Forgery.  An attacker who can lure an authenticated Icinga
Web 2 user to a malicious page can delete any configuration section managed by
that form without knowledge of the victim's CSRF token — because the delete
path in `handleRequest()` never reaches `isValid()`.

## Affected code

| File | Method / location | Class |
|---|---|---|
| `library/Icinga/Web/Form/ConfigSectionForm.php` | constructor (ON_SENT registration), `onSent()`, `handleDelete()`, `shouldDelete()` | `ConfigSectionForm` |
| `library/Icinga/Web/Form/ConfigForm.php` | `addButtonElements()` | `ConfigForm` |

## Root-cause analysis

### 1 — How `handleRequest()` routes the request

`ConfigSectionForm` extends `ConfigForm` → `CompatForm` → `ipl\Html\Form`.
`CompatForm` adds neither `FormUid` nor `CsrfCounterMeasure`, so `hasBeenSent()`
reduces to a plain HTTP-method check:

```php
// ipl\Html\Form::hasBeenSent()
return $this->request->getMethod() === $this->getMethod(); // true for any POST
```

Inside `handleRequest()` the branching is:

```php
if ($this->hasBeenSubmitted()) {   // store button pressed?
    if ($this->isValid()) {        // CSRF would be checked here
        $this->emit(ON_SENT, …);
        $this->onSuccess();
    } else {
        $this->onError();
    }
} else {                           // ← taken when delete button is pressed
    $this->validatePartial();
    $this->emit(ON_SENT, …);       // fires WITHOUT isValid()
}
```

`hasBeenSubmitted()` calls `getSubmitButton()->hasBeenPressed()`.  The submit
button tracked by ipl-html is the **first** `addElement('submit', …)` call,
which is the `store` button added by `ConfigForm::addButtonElements()`.  The
delete button is registered via `registerElement()` (not `addElement()`), so
it is **never** the tracked submit button.

Pressing only the delete button means `store` was not pressed →
`hasBeenSubmitted()` = false → the `else` branch runs → `ON_SENT` fires.

### 2 — Delete handler registered on ON_SENT

```php
// ConfigSectionForm::__construct()
$this->on(static::ON_SENT, $this->onSent(...));

// ConfigSectionForm::onSent()
protected function onSent(): void
{
    if ($this->shouldDelete()) {
        $this->handleDelete();         // deletes the INI section
        $this->emit(static::ON_DELETE, [$this]);
    }
}
```

`shouldDelete()` only checks whether the delete button element (`delete`) has
been pressed:

```php
return $deleteButton->hasBeenPressed(); // getValue() === 'y'
```

`getValue()` reads directly from the populated request body.  If the POST
body contains `delete=y`, this returns `true` — regardless of whether a CSRF
token was presented.

### 3 — No CSRF anywhere on this form stack

Neither `CompatForm` nor `ConfigForm` adds a CSRF counter-measure element.
The `ipl\Web\Common\CsrfCounterMeasure` trait is absent from the entire
inheritance chain.  All state-mutating actions on `ConfigSectionForm` (create,
save, rename, **delete**) are therefore CSRF-vulnerable.  The delete path is
the most critical because it bypasses even field-level validation.

## Exploit

### Minimal PoC (attacker-controlled HTML page)

```html
<form id="f" method="POST"
      action="https://icinga.example.com/icingaweb2/<module>/edit?name=<section>">
  <input type="hidden" name="delete" value="y">
</form>
<script>document.getElementById('f').submit();</script>
```

The victim only needs to have an active Icinga Web 2 session.  No CSRF token,
no form UID, no knowledge of any secret is required.  The endpoint must use a
`ConfigSectionForm` subclass that has not called `setAllowDeletion(false)`.

### Step-by-step request trace

1. Victim's browser sends `POST /<module>/edit?name=<section>` with session
   cookie and body `delete=y`.
2. `hasBeenSent()` → true (POST).
3. `populate(['delete' => 'y'])`.
4. `ensureAssembled()` → form built, delete button element added with
   `submitValue = 'y'`.
5. `hasBeenSubmitted()` → false (`store` button absent from body).
6. `else` branch: `validatePartial()` (no elements with values to validate
   because no field names match), then `emit(ON_SENT)`.
7. `ON_SENT` handler (`onSent()`): `shouldDelete()` →
   `deleteButton->getValue() === 'y'` → true.
8. `handleDelete()`:
   - `$this->config->removeSection('<section>')`
   - `$this->config->saveIni()`
   - Section permanently deleted from disk.

## Impact

* Any authenticated Icinga Web 2 user can be CSRF'd into deleting arbitrary
  named sections from any INI config file managed by a `ConfigSectionForm`
  subclass that permits deletion (i.e. has not called `setAllowDeletion(false)`).
* Every module that adopts `ConfigSectionForm` from this branch inherits the
  vulnerability without any additional mistakes on the module author's part.

## Fix recommendation

Register the delete logic on `ON_SUBMIT` (fires only after `isValid()`) rather
than `ON_SENT`.  If skipping field validation for invalid-config deletion is
genuinely needed, validate *at least* the CSRF element explicitly before
calling `handleDelete()`.  Alternatively, add `CsrfCounterMeasure` to
`ConfigForm` so all subclasses inherit CSRF protection automatically.

Al2Klimov (changes requested):

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,156 @@
...
+    public function __construct(
+        protected Config $config,
+    ) {
+    }
+
+    public function ensureAssembled(): static
+    {
+        if (! $this->hasBeenAssembled) {
+            parent::ensureAssembled();
+            $this->addRequiredElements();
```
Al2Klimov, line 92: Nice that I get CSRF protection automatically, but https://github.com/Icinga/icingaweb2/pull/5480#discussion_r3289029138 seems not to work anymore:

```
No CSRF counter measure ID set

#0 /usr/share/php/Icinga/Web/Form/ConfigForm.php(149): Icinga\Web\Form\ConfigForm->addCsrfCounterMeasure()
#1 /usr/share/php/Icinga/Web/Form/ConfigForm.php(37): Icinga\Web\Form\ConfigForm->addRequiredElements()
#2 /usr/share/icinga-php/ipl/vendor/ipl/html/src/Form.php(235): Icinga\Web\Form\ConfigForm->ensureAssembled()
#3 /usr/share/icingaweb2/modules/test5480/application/controllers/IndexController.php(14): ipl\Html\Form->handleRequest()
#4 /usr/share/icinga-php/vendor/vendor/icinga/zf1/library/Zend/Controller/Action.php(528): Icinga\Module\Test5480\Controllers\IndexController->indexAction()
#5 /usr/share/php/Icinga/Web/Controller/Dispatcher.php(78): Zend_Controller_Action->dispatch()
#6 /usr/share/icinga-php/vendor/vendor/icinga/zf1/library/Zend/Controller/Front.php(954): Icinga\Web\Controller\Dispatcher->dispatch()
#7 /usr/share/php/Icinga/Application/Web.php(296): Zend_Controller_Front->dispatch()
#8 /usr/share/php/Icinga/Application/webrouter.php(107): Icinga\Application\Web->dispatch()
#9 /usr/share/icingaweb2/public/index.php(4): require_once(String)
#10 {main}
```

Al2Klimov (changes requested):

library/Icinga/Web/Form/ConfigForm.php
TheSyscall: That is correct. All ConfigForms require CsrfCounterMeasure. This is on request of @lippserd

Al2Klimov: Yeah, but shouldn't your ensureAssembled() make it working automagically in my mentioned test form?

TheSyscall: @lippserd would like to have the controller call it explicitly rather than burry the call site of `setCsrfCounterMeasureId` inside the form.

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,156 @@
...
+    protected function save(): void
+    {
+        foreach ($this->getValues() as $element => $value) {
+            if (($parts = $this->getIniKeyFromName($element)) === null) {
+                continue;
+            }
+            [$section, $key] = $parts;
+
+            $configSection = $this->config->getSection($section);
+            if (Str::isEmpty($value)) {
```
Al2Klimov, line 164: * We must get https://github.com/Icinga/ipl-stdlib/pull/77 done before merging this PR

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,402 @@
...
+ * Base class for configuration forms that manage a single INI section
+ *
+ * Extends {@see ConfigForm} with support for creating, renaming, and deleting
+ * named sections. Element names map directly to keys within the section rather
+ * than encoding the section via the {@see ConfigForm} double-underscore convention.
+ *
+ * Emits {@see self::ON_DELETE} after a section is deleted and {@see self::ON_RENAME}
+ * after a section is renamed.
+ */
+class ConfigSectionForm extends ConfigForm
```
Al2Klimov: Passwords vanish from the config if I edit something, but not the password.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,402 @@
...
+     * @var bool
+     */
+    protected bool $isCreateForm = false;
+
+    /**
+     * Whether the form allows deletion of the configuration section
+     *
+     * @var bool
+     */
+    protected bool $allowDeletion = true;
```
Al2Klimov: I can bypass the enforced delete button absence by emptying all fields and saving.

Al2Klimov (changes requested):

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Changed password elements to revert to their old value if submitted empty.

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Removed automatic removal of sections for `ConfigSectonForm`

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,176 @@
...
+     *
+     * @return bool
+     */
+    protected function shouldRevertIfEmpty(string $elementName): bool
+    {
+        $element = $this->getElement($elementName);
+        return match (true) {
+            $element instanceof PasswordElement => true,
+            default => false,
+        };
```
Al2Klimov:
```suggestion
        return $this->getElement($elementName) instanceof PasswordElement;
```

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,417 @@
...
+        return parent::isValidEvent($event);
+    }
+
+    public function isValid(): bool
+    {
+        if ($this->hasBeenSubmitted() && $this->shouldDelete()) {
+            if (! $this->hasElement('CSRFToken')) {
+                return true;
+            }
+            $csrf = $this->getElement('CSRFToken');
```
Al2Klimov: That variable can be inlined.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,417 @@
...
+
+            return $csrf->isValid();
```
Al2Klimov: While on it, you can make one return with an `||`.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,417 @@
...
+     *
+     * @return bool
+     */
+    public function allowDeletion(): bool
+    {
+        if ($this->isCreateForm()) {
+            return false;
+        }
+
+        return $this->allowDeletion;
```
Al2Klimov: You could collapse this whole method into one return via `&&`.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,417 @@
...
+     *
+     * @return bool
+     */
+    public function allowRename(): bool
+    {
+        if ($this->isCreateForm()) {
+            return false;
+        }
+
+        return $this->allowRename;
```
Al2Klimov: Same here.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,417 @@
...
+
+        $params['required'] = true;
+        $params['ignore'] = true;
+        $params['label'] ??= $this->translate('Name');
+        $params['validators'][] = new CallbackValidator(function ($value, CallbackValidator $validator) {
+            if (Str::isEmpty($value)) {
+                return true;
+            }
+
+            if ($value === $this->section) {
```
Al2Klimov: These two ifs are mergeable via `||`.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,417 @@
...
+        $this->addElement('text', static::NAME_ELEMENT_NAME, $params);
+    }
+
+    protected function getIniKeyFromName(string $name): ?array
+    {
+        if ($this->section === null) {
+            return null;
+        }
+
+        return [$this->section, $name];
```
Al2Klimov: Have I already told you I'm a huge fan of the ternary operator?

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,417 @@
...
+        return $this->section !== $this->getPopulatedValue(static::NAME_ELEMENT_NAME);
+    }
+
+    protected function save(): void
+    {
+        foreach ($this->getValues() as $element => $value) {
+            if (Str::isEmpty($value) && $this->shouldRevertIfEmpty($element)) {
+                $value = $this->config->get($this->section, $element);
+            }
+            $configSection = $this->config->getSection($this->section);
```
Al2Klimov: Do you have to run this line for every iteration?

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,417 @@
...
+            if (Str::isEmpty($value)) {
+                unset($configSection[$element]);
+            } else {
+                $configSection->$element = $value;
+            }
+            $this->config->setSection($this->section, $configSection);
```
Al2Klimov: Do you have to run this line for every iteration?

Al2Klimov (commented):
Several style stuff:

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: This whole method has been removed. We always require csrf protection, so no special handling is needed anymore.

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: This whole method has been removed. We always require csrf protection, so no special handling is needed anymore.

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Changed

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Changed

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Merged

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: I'm not but I see your point. Changed

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: I don't. Changed.

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Changed.

library/Icinga/Web/Form/ConfigForm.php
TheSyscall: Inlined

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,386 @@
...
+            $this->getElement(static::SUBMIT_BUTTON_NAME)
+                ->getWrapper()
+                ->prepend($deleteButton);
+        }
+
+        if (($this->isCreateForm() || $this->allowRename()) && ! $this->hasElement(static::NAME_ELEMENT_NAME)) {
+            $this->addSectionNameElement();
+
+            $content = $this->getContent();
+            $index = array_find_key($content, function ($element) {
```
Al2Klimov:
```
[root@9f4e22791a6f /]# php -a
Interactive shell

php > array_find_key(); // Available since PHP 8.4
PHP Warning:  Uncaught Error: Call to undefined function array_find_key() in php shell code:1
Stack trace:
#0 {main}
  thrown in php shell code on line 1
php > ^D
[root@9f4e22791a6f /]# php -v
PHP 8.3.27 (cli) (built: Oct 21 2025 14:53:41) (NTS gcc x86_64)
Copyright (c) The PHP Group
Zend Engine v4.3.27, Copyright (c) Zend Technologies
[root@9f4e22791a6f /]#
```

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,386 @@
...
+                return $element->getName() === static::NAME_ELEMENT_NAME;
+            });
+            if ($index === false) {
```
Al2Klimov: It's null, not false.

```
php > var_dump(array_find_key([], function(){}));
NULL
php >
```

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,386 @@
...
+use LogicException;
+
+/**
+ * Base class for configuration forms that manage a single INI section
+ *
+ * Extends {@see ConfigForm} with support for creating, renaming, and deleting
+ * named sections. Element names map directly to keys within the section rather
+ * than encoding the section via the {@see ConfigForm} double-underscore convention.
+ *
+ * Emits {@see self::ON_DELETE} after a section is deleted and {@see self::ON_RENAME}
```
Al2Klimov: It doesn't emit ON_DELETE actually, see `git grep -nFwe ON_DELETE`.

Al2Klimov (changes requested):
(FYI: If you

1. read this
2. are a colleague of mine
3. and you are still skeptical about AI

this review is the proof AI sees things humans don't see.)

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: How did that ever pass PHPCS/PHPlint?
Changed it.

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Changed

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Fixed it.

library/Icinga/Web/Form/ConfigSectionForm.php
Al2Klimov: > How did that ever pass PHPCS/PHPlint?

Your next task is to figure that out. :)

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,394 @@
...
+    public function allowRename(): bool
+    {
+        return ! $this->isCreateForm() && $this->allowRename;
+    }
+
+    /**
+     * Handle the deletion of the configuration section
+     *
+     * This method is called when the delete button is pressed.
+     * It deletes the underlying section regardless of whether form validation passed.
```
Al2Klimov:
```suggestion
```

Another round of throwing AI at your PR broke this promise:

https://github.com/user-attachments/assets/5451a334-9b5a-4d3a-9ca7-3fb1d92dde84

You can just remove it if it's not super critical.

Al2Klimov (dismissed):

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: Mistakes were made during refactor. Fixed it.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,407 @@
...
+
+    /**
+     * Handle the renaming of the configuration section
+     *
+     * This method is called when the rename button is pressed.
+     * It renames the underlying section and updates the section name in the form.
+     *
+     * @return void
+     *
+     * @throws LogicException
```
jrauh01: Remove `@throws` annotation, as this method does not directly throw.

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,172 @@
...
+    {
+        return $this->getElement($elementName) instanceof PasswordElement;
+    }
+
+    /**
+     * Persist the current configuration to disk
+     *
+     * If an error occurs, the form will be re-rendered with the error message
+     * and the raw INI configuration.
+     */
```
jrauh01, line 150: `@return` annotation is missing.

```suggestion
     * @return void
     */

File: library/Icinga/Web/Form/ConfigForm.php [hunk 42]
```diff
@@ -0,0 +1,172 @@
...
+                    $e,
+                    $this->config,
+                )
+            );
+            $this->setContent($content);
+            throw $e;
+        }
+    }
+
+    protected function addRequiredElements(): void
```
jrauh01, line 210: Doc block is missing completely.

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,172 @@
...
+    use CsrfCounterMeasure;
+
+    /** @var string Name of the submit button element */
+    protected const SUBMIT_BUTTON_NAME = 'store';
+
+    /** @var string Delimiter used to separate the section and key in the element name */
+    protected string $sectionKeyDelimiter = '__';
+
+    public function __construct(
+        protected Config $config,
```
jrauh01, line 51: Doc block to document the property is missing.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,407 @@
...
+    /**
+     * Whether the form allows renaming of the configuration section
+     *
+     * @var bool
+     */
+    protected bool $allowRename = true;
+
+    public function __construct(
+        Config $config,
+        protected ?string $section = null,
```
jrauh01: Doc block to document the properties is missing.

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,172 @@
...
+    ) {
+    }
+
+    public function ensureAssembled(): static
+    {
+        if (! $this->hasBeenAssembled) {
+            parent::ensureAssembled();
+            $this->addRequiredElements();
+            $this->populateFromConfig();
+        }
```
jrauh01, line 93: Maybe a comment explaining this would make sense.

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,172 @@
...
+            $content = $this->getContent();
+            array_unshift(
+                $content,
+                new ShowConfiguration(
+                    $e,
+                    $this->config,
+                )
+            );
+            $this->setContent($content);
+            throw $e;
```
jrauh01, line 198: Document this please and add empty line above.

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,172 @@
...
+     *
+     * @param string $name The element name
+     *
+     * @return ?array<string, string>
+     */
+    protected function getIniKeyFromName(string $name): ?array
+    {
+        $parts = explode($this->sectionKeyDelimiter, $name, 2);
+
+        return count($parts) === 2 ? $parts : null;
```
jrauh01: Is there a reason why you return `null` here instead of just returning the result of `explode` and check the count at the call side? At the call sides i would then do something like:
```php
$parts = $this->getIniKeyFromName($name);
if (count($parts) !== 2) {
```
instead of:
```php
if (($parts = $this->getIniKeyFromName($name)) === null) {
```

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,407 @@
...
+            $this->emit(static::ON_DELETE, [$this]);
+        } catch (Exception $e) {
+            $content = $this->getContent();
+            array_unshift(
+                $content,
+                new ShowConfiguration(
+                    $e,
+                    $this->config,
+                )
+            );
```
jrauh01: Same as in `ConfigForm`.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,407 @@
...
+            $this->setContent($content);
+            throw $e;
```
jrauh01: Same as in `ConfigForm`.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,407 @@
...
+        $newName = $this->getPopulatedValue(static::NAME_ELEMENT_NAME);
+        $this->config->setSection($newName, $this->config->getSection($oldName));
+        $this->config->removeSection($oldName);
+        $this->section = $newName;
+        parent::onSuccess();
+        $this->emit(static::ON_RENAME, [
+            $this,
+            $oldName,
+            $this->section,
+        ]);
```
jrauh01: Why breaking this over multiple lines? Would fit easily into one.

```suggestion
       $this->emit(static::ON_RENAME, [this, oldName, this->section]);

File: library/Icinga/Web/Form/ConfigSectionForm.php [hunk 51]
```diff
@@ -0,0 +1,407 @@
...
+        $params['ignore'] = true;
+        $params['label'] ??= $this->translate('Name');
+        $params['validators'][] = new CallbackValidator(function ($value, CallbackValidator $validator) {
+            if (Str::isEmpty($value) || $value === $this->section) {
+                return true;
+            }
+
+            if ($this->config->hasSection($value)) {
+                $validator->addMessage($this->translate('An entry with this name already exists.'));
+                return false;
```
jrauh01: Please add blank line above.

library/Icinga/Web/Form/ConfigSectionForm.php
```diff
@@ -0,0 +1,407 @@
...
+        if (($this->isCreateForm() || $this->allowRename()) && ! $this->hasElement(static::NAME_ELEMENT_NAME)) {
+            $this->addSectionNameElement();
+
+            $content = $this->getContent();
+            $index = null;
+            foreach ($content as $key => $element) {
+                /** @var FormElement $element */
+                if ($element->getName() === static::NAME_ELEMENT_NAME) {
+                    $index = $key;
+                    break;
```
jrauh01: Please add blank line above.

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,172 @@
...
+            $this->save();
+        } catch (Exception $e) {
+            $content = $this->getContent();
+            array_unshift(
+                $content,
+                new ShowConfiguration(
+                    $e,
+                    $this->config,
+                )
+            );
```
jrauh01: Why are you splitting this over multiple lines? Wouldn't this be fine?

```suggestion
            array_unshift($content, new ShowConfiguration($e, $this->config));

**Review by jrauh01 — CHANGES_REQUESTED** [review]
Sometimes you do one line doc blocks for properties and sometimes multiline, but i cannot tell which pattern you are following. I think it would be better to standardize this.

File: library/Icinga/Web/Form/ConfigSectionForm.php [hunk 54]
```diff
@@ -0,0 +1,407 @@
...
+    {
+        return $this->isCreateForm;
+    }
+
+    /**
+     * Set whether the form allows deletion of the configuration section
+     *
+     * @param bool $allowDeletion
+     *
+     * @return static
```
jrauh01:
```suggestion
     * @return $this

File: library/Icinga/Web/Form/ConfigForm.php [hunk 55]
```diff
@@ -0,0 +1,172 @@
...
+    {
+        foreach ($this->getValues() as $element => $value) {
+            if (($parts = $this->getIniKeyFromName($element)) === null) {
+                continue;
+            }
+            [$section, $key] = $parts;
+            if (Str::isEmpty($value) && $this->shouldRevertIfEmpty($element)) {
+                $value = $this->config->get($section, $key);
+            }
+            $configSection = $this->config->getSection($section);
```
jrauh01:
```suggestion
            $configSection = $this->config->getSection($section);
            if (Str::isEmpty($value) && $this->shouldRevertIfEmpty($element)) {
                $value = $configSection->get($key);
            }

File: library/Icinga/Web/Form/ConfigForm.php [hunk 56]
```diff
@@ -0,0 +1,172 @@
...
+            if (Str::isEmpty($value)) {
+                unset($configSection[$key]);
+            } else {
+                $configSection->$key = $value;
```
jrauh01: Maybe use array annotation as in the line two lines above?

```suggestion
                $configSection[$key] = $value;

File: library/Icinga/Web/Form/ConfigForm.php [hunk 57]
```diff
@@ -0,0 +1,216 @@
...
+     *
+     * @return void
+     */
+    protected function onSuccess(): void
+    {
+        try {
+            $this->save();
+        } catch (Exception $e) {
+            $content = $this->getContent();
+            array_unshift($content, new ShowConfiguration($e, $this->config));
```
jrauh01: You could use our new `DisplayFormElement` from IPL Web. **Don't forget to import** if applying the suggestion! Also apply this for the other places where `ShowConfiguration` is used.

```suggestion
            array_unshift($content, new DisplayFormElement(new ShowConfiguration($e, $this->config)));

File: library/Icinga/Web/Form/ConfigSectionForm.php [hunk 58]
```diff
@@ -0,0 +1,426 @@
...
+        $configSection = $this->config->getSection($this->section);
+        foreach ($this->getValues() as $element => $value) {
+            if (Str::isEmpty($value) && $this->shouldRevertIfEmpty($element)) {
+                $value = $this->config->get($this->section, $element);
+            }
+
+            if (Str::isEmpty($value)) {
+                unset($configSection[$element]);
+            } else {
+                $configSection->$element = $value;
```
jrauh01: Maybe use array annotation as in the line two lines above?

```suggestion
                $configSection[$element] = $value;

**Review by jrauh01 — CHANGES_REQUESTED** [review]
Please look over `ShowConfiguration` and inline the element creations where possible.

File: library/Icinga/Web/Form/ConfigForm.php (hunk 57 above)
Comment (TheSyscall) [reply library/Icinga/Web/Form/ConfigForm.php]: Changed

File: library/Icinga/Web/Form/ConfigForm.php [hunk 59]
```diff
@@ -0,0 +1,228 @@
...
+     * @param string $name The element name
+     *
+     * @return string[] An array containing the section and key, or an empty array
+     * if the name is not a valid configuration key
+     */
+    protected function getIniKeyFromName(string $name): array
+    {
+        $parts = explode($this->sectionKeyDelimiter, $name, 2);
+        return count($parts) === 2 ? $parts : [];
+    }
```
lippserd, line 81:
```suggestion
     * @return array{?string, ?string} TODO: Change docs.
     */
    protected function getIniKeyFromName(string $name): array
    {
        return Str::symmetricSplit($name, $this->sectionKeyDelimiter, 2);
    }
```

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,228 @@
...
+     * @return mixed The value of the configuration key or the default value
+     */
+    protected function getConfigValue(string $name, mixed $default = null): mixed
+    {
+        $parts = $this->getIniKeyFromName($name);
+        if (count($parts) !== 2) {
+            return $default;
+        }
+
+        return $this->config->get($parts[0], $parts[1], $default);
```
lippserd:
```suggestion
        [$section, $key] = $this->getIniKeyFromName($name)
        if ($key === null) {
            return $default;
        }

        return $this->config->get($section, $key, $default);
```

lippserd (changes requested):
Changed my mind, sorry. I think the suggestion is better than the current implementation. Make sure to adjust all call sites.

library/Icinga/Web/Form/ConfigForm.php
TheSyscall: Changed method signature

library/Icinga/Web/Form/ConfigForm.php
TheSyscall: Changed call sites

library/Icinga/Web/Widget/ShowConfiguration.php
```diff
@@ -0,0 +1,76 @@
...
+        $this->addHtml(HtmlElement::create('ul', null, [
+            HtmlElement::create('li', null, $this->translate(
+                "You don't have file-system permissions to write to the file"
+            )),
+            HtmlElement::create('li', null, $this->translate('Something went wrong while writing the file')),
+            HtmlElement::create(
+                'li',
+                null,
+                $this->translate("There's an application error preventing you from persisting the configuration"),
+            ),
```
jrauh01:
```suggestion
            HtmlElement::create('li', null, $this->translate(
                "There's an application error preventing you from persisting the configuration"
            )),

File: library/Icinga/Web/Widget/ShowConfiguration.php [hunk 62]
```diff
@@ -0,0 +1,76 @@
...
+ */
+class ShowConfiguration extends BaseHtmlElement
+{
+    use Translation;
+
+    protected $tag = 'div';
+
+    /**
+     * Create a new ShowConfiguration widget
+     * @param Exception $exception The exception that was thrown
```
jrauh01:
```suggestion
     * Create a new ShowConfiguration widget
     * 
     * @param Exception $exception The exception that was thrown

File: library/Icinga/Web/Form/ConfigSectionForm.php [hunk 63]
```diff
@@ -0,0 +1,445 @@
...
+use ipl\Validator\CallbackValidator;
+use ipl\Web\Compat\DisplayFormElement;
+use LogicException;
+
+/**
+ * Base class for configuration forms that manage a single INI section
+ *
+ * Extends {@see ConfigForm} with support for creating, renaming, and deleting
+ * named sections. Element names map directly to keys within the section rather
+ * than encoding the section via the {@see ConfigForm} double-underscore convention.
```
jrauh01: Consider describing this using `ConfigForm::$sectionKeyDelimiter` instead of implementation details.

jrauh01 (changes requested):

library/Icinga/Web/Widget/ShowConfiguration.php
TheSyscall: Changed

library/Icinga/Web/Widget/ShowConfiguration.php
TheSyscall: Changed

library/Icinga/Web/Form/ConfigSectionForm.php
TheSyscall: `ConfigSectionForm` has been removed from this PR

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -0,0 +1,211 @@
...
+                    $exception->getMessage(),
+                ),
+                HtmlElement::create('br'),
+                t('This could have one or more of the following reasons:'),
+            ]),
+            HtmlElement::create('ul', null, [
+                HtmlElement::create('li', null, t("You don't have file-system permissions to write to the file")),
+                HtmlElement::create('li', null, t('Something went wrong while writing the file')),
+                HtmlElement::create('li', null, t(
+                    "There's an application error preventing you from persisting the configuration"
```
jrauh01:
```suggestion
                    "There's an application error preventing you from persisting the configuration",

File: library/Icinga/Web/Form/ConfigForm.php [hunk 65]
```diff
@@ -0,0 +1,211 @@
...
+            ]),
+            HtmlElement::create('p', null, [
+                t(
+                    'Details can be found in the application log. ' .
+                    "(If you don't have access to this log, call your administrator in this case)",
+                ),
+                HtmlElement::create('br'),
+                t('In case you can access the file by yourself, you can open it and insert the config manually:'),
+            ]),
+            HtmlElement::create('p', null, HtmlElement::create('pre', null, $code))
```
jrauh01:
```suggestion
            HtmlElement::create('p', null, HtmlElement::create('pre', null, $code)),

jrauh01 (changes requested):
Overall this looks good, I would only adjust those two lines to stay consistent.

jrauh01 (approved):
LGTM

lippserd (approved):

PR #5487, opened 2026-03-27
Remove monitoring module

Al2Klimov:
* Backport of #5458

Eric decided that Icinga Web 2.12.7 will drop the monitoring module.

Technically a breaking change, but people may still clone https://github.com/Icinga/icingaweb2-module-monitoring, so it should be mostly fine.

No need for hurry, but keep this in mind while doing Icinga Web 2.12.7, ok?


sukhwinder33445 (approved):
Note: The documentation still includes references to the monitoring module, which should be removed once we have completely dropped support for the monitoring module.

lippserd (changes requested):
Please use `git cherry-pick -x`

PR #5491, opened 2026-04-08
Introduce missing `first` and `htmlToPdf` methods to `PdfexportHook`

TheSyscall:
This PR adds two methods to the pdfexport hook: `first` and `htmlToPdf`

Both functions were previously called without actually being part of the hook.
The only possible implementation of this hook comes from the pdfexport module which already defines the same two methods.

Intentionally leaving out strict types and `HookEssentials` to be backward compatible with the original hook implementation.

### DIFF
--- a/library/Icinga/Application/Hook/PdfexportHook.php
+++ b/library/Icinga/Application/Hook/PdfexportHook.php
@@ -7,6 +7,7 @@ namespace Icinga\Application\Hook;
 
 use Icinga\Application\Hook;
 use Icinga\Application\Logger;
+use ipl\Html\ValidHtml;
 use RuntimeException;
 use Throwable;
 
@@ -44,15 +45,17 @@ abstract class PdfexportHook
     /**
      * Get whether PDF export is supported
      *
-     * @return  bool
+     * @return bool
      */
     abstract public function isSupported();
 
     /**
      * Render the specified HTML to PDF and stream it to the client
      *
-     * @param   string  $html       The HTML to render to PDF
-     * @param   string  $filename   The filename for the generated PDF
+     * @param ValidHtml $html The HTML to render to PDF
+     * @param string $filename The filename for the generated PDF
+     *
+     * @return never
      */
     abstract public function streamPdfFromHtml($html, $filename);
 
--- a/library/Icinga/Application/Hook/PdfexportHook.php
+++ b/library/Icinga/Application/Hook/PdfexportHook.php
@@ -29,11 +29,9 @@ abstract class PdfexportHook
 
         foreach (Hook::all('Pdfexport') as $exporter) {
             try {
-                if (! $exporter->isSupported()) {
-                    continue;
+                if ($exporter->isSupported()) {
+                    return $exporter;
                 }
-
-                return $exporter;
             } catch (Throwable $e) {
                 Logger::error('PDF exporter reported an error during support check: %s', $e);
             }
--- a/library/Icinga/Application/Hook/PdfexportHook.php
+++ b/library/Icinga/Application/Hook/PdfexportHook.php
@@ -7,6 +7,7 @@ namespace Icinga\Application\Hook;
 
 use Icinga\Application\Hook;
 use Icinga\Application\Logger;
+use Icinga\Exception\IcingaException;
 use ipl\Html\ValidHtml;
 use RuntimeException;
 use Throwable;
@@ -23,21 +24,18 @@ abstract class PdfexportHook
      */
     public static function first()
     {
-        if (! Hook::has('Pdfexport')) {
-            throw new RuntimeException('No PDF exporter available');
-        }
-
         foreach (Hook::all('Pdfexport') as $exporter) {
             try {
                 if ($exporter->isSupported()) {
                     return $exporter;
                 }
             } catch (Throwable $e) {
-                Logger::error('PDF exporter reported an error during support check: %s', $e);
+                Logger::error('PDF exporter reported an error during support check: %s', $e->getMessage());
+                Logger::error("%s\n%s", $e, IcingaException::getConfidentialTraceAsString($e));
             }
         }
 
-        throw new RuntimeException('Not supported PDF exporter available');
+        throw new RuntimeException('No supported PDF exporter available');
     }
 
     /**
--- a/library/Icinga/Application/Hook/PdfexportHook.php
+++ b/library/Icinga/Application/Hook/PdfexportHook.php
@@ -54,4 +54,13 @@ abstract class PdfexportHook
      * @param   string  $filename   The filename for the generated PDF
      */
     abstract public function streamPdfFromHtml($html, $filename);
+
+    /**
+     * Render the specified HTML to PDF and return the PDF document as a string
+     *
+     * @param ValidHtml $html The HTML to render to PDF
+     *
+     * @return string
+     */
+    abstract public function htmlToPdf($html);
 }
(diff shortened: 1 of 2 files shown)


Al2Klimov (commented):
> This function was previously called without actually being part of the hook.

Then it was in fact already required, so adding it formally is not a breaking change.

The typing, in contrast, is. In fact, you already broke your own https://github.com/Icinga/icingaweb2-module-pdfexport/pull/88:

`Fatal error: Declaration of Icinga\Module\Pdfexport\ProvidedHook\Pdfexport::isSupported() must be compatible with Icinga\Application\Hook\PdfexportHook::isSupported(): bool in /usr/share/icingaweb2/modules/pdfexport/library/Pdfexport/ProvidedHook/Pdfexport.php on line 58`

Consider omitting the types.

library/Icinga/Application/Hook/PdfexportHook.php
```diff
@@ -5,11 +5,32 @@
...
+     * @return static
+     */
+    public static function first()
+    {
+        if (! Hook::has('Pdfexport')) {
+            throw new RuntimeException('No PDF exporter available');
+        }
+        $pdfexport = Hook::first('Pdfexport');
+        if (! $pdfexport->isSupported()) {
+            throw new RuntimeException('PDF exporter is not supported');
```
Al2Klimov: So the second hook won't save you, even if it supports export? :-)

TheSyscall: Make sense, changed

library/Icinga/Common/PdfExport.php
```diff
@@ -5,6 +5,8 @@
 
 namespace Icinga\Common;
 
+use Icinga\Application\Hook;
```
Al2Klimov:
```suggestion
```

library/Icinga/Application/Hook/PdfexportHook.php
```diff
@@ -5,23 +5,66 @@
...
  * Base class for the PDF Export Hook
  */
 abstract class PdfexportHook
 {
+    /**
+     * Get the first hook
+     *
+     * @return static
+     */
+    public static function first()
```
Al2Klimov, line 25: * https://github.com/Icinga/developer-guidelines-drafts/issues/83

library/Icinga/Application/Hook/PdfexportHook.php
```diff
@@ -5,23 +5,66 @@
...
+    {
+        if (! Hook::has('Pdfexport')) {
+            throw new RuntimeException('No PDF exporter available');
```
Al2Klimov: I personally translate messages which appear in the UI.

library/Icinga/Application/Hook/PdfexportHook.php
```diff
@@ -5,23 +5,66 @@
...
+                    continue;
+                }
+
+                return $exporter;
+            } catch (Throwable $e) {
+                Logger::error('PDF exporter reported an error during support check: %s', $e);
+            }
+        }
+
+        throw new RuntimeException('Not supported PDF exporter available');
```
Al2Klimov: I personally translate messages which appear in the UI.

library/Icinga/Application/Hook/PdfexportHook.php
```diff
@@ -5,23 +5,66 @@
...
+            throw new RuntimeException('No PDF exporter available');
+        }
+
+        foreach (Hook::all('Pdfexport') as $exporter) {
+            try {
+                if (! $exporter->isSupported()) {
+                    continue;
+                }
+
+                return $exporter;
```
Al2Klimov: While on it, you could put the return into the if and invert latter. You decide.

library/Icinga/Application/Hook/PdfexportHook.php
Al2Klimov: Ah, wait, this could break https://github.com/Icinga/icingaweb2-module-pdfexport! 🙈

library/Icinga/Application/Hook/PdfexportHook.php
TheSyscall: There are two reasons why this message would appear in the UI:

- We show the export to PDF button when then should not -> Then we need to remove the button.
  - #5510
  - Icinga/icingaweb2-module-reporting#276
- An admin removed all PDF export backends during the time a user was using icingaweb2 -> I don't think this edge case is worth the translation effort.

library/Icinga/Application/Hook/PdfexportHook.php
TheSyscall: I don't think it is worth expanding the Hook behavior to add a translation here. While I don't expect anyone to have a translation method in a hook implementation, I don't want to change the Hook if I can avoid it.

Al2Klimov: I didn't say you need to change the hook for this.

library/Icinga/Application/Hook/PdfexportHook.php
Al2Klimov: > Then we need to remove the button.

Go for it! 👍

library/Icinga/Application/Hook/PdfexportHook.php
TheSyscall: @lippserd said not to translate exceptions like this. (Forms is a different situation)

library/Icinga/Application/Hook/PdfexportHook.php
TheSyscall: Code calling a `Hook::first()` is expected to call `Hook::has()` first. So this guard will be removed completely.

library/Icinga/Application/Hook/PdfexportHook.php
TheSyscall: Changed

library/Icinga/Common/PdfExport.php
TheSyscall: Removed

library/Icinga/Application/Hook/PdfexportHook.php
Al2Klimov: > Code calling a `Hook::first()` is expected to call `Hook::has()` first.

Err.. no? `Hook::first()` may return null.

TheSyscall: Let me rephrase, everyone calling this hook is expected to call `Hook::has()` first. That was the case before and this PR doesn't change this.

I asked @lippserd before making the previous statement and I falsely assumed his statement about this function to be universal.

In any case, this guard will be removed, and the exception will be thrown after the for loop (which will have zero elements anyway)

library/Icinga/Application/Hook/PdfexportHook.php
```diff
@@ -5,23 +5,62 @@
...
+    public static function first()
+    {
+        foreach (Hook::all('Pdfexport') as $exporter) {
+            try {
+                if ($exporter->isSupported()) {
+                    return $exporter;
+                }
+            } catch (Throwable $e) {
+                Logger::error('PDF exporter reported an error during support check: %s', $e->getMessage());
+                Logger::error("%s\n%s", $e, IcingaException::getConfidentialTraceAsString($e));
```
Al2Klimov: One last stupid question from my side: why **two** log messages, not one, especially given they're on the same level?

TheSyscall: Miscommunication. Changed

Al2Klimov (dismissed):

lippserd (changes requested):
Please rebase if needed and fixup your commits. It makes sense to mention in the commit message(s) that strict types have been left out on purpose to be compatible with the already existing implementation.

TheSyscall:
Reverted implementation of `HookEssentials` as it breaks backward compatibility with the original implementation.

lippserd (approved):

PR #5492, opened 2026-04-09
Migrate Less compilation to `ipl\Web\Less`

lippserd:
Because the in-tree Less visitor requires `wikimedia/less.php`, that dependency has to live in `icinga-php-thirdparty`. `icinga-php-library` already carries Less-related code with the same dependency, so this PR consolidates everything into `icinga-php-library` and removes the duplicate in-tree implementation.

- Remove the in-tree `Icinga\Less\` visitor layer in favor of the shared `ipl\Web\Less` library, which maintains modernized equivalent CSS var replacement and light-mode functionality.

- Strip unused variable-export support from `LessCompiler`. The feature was never used by any our modules or community modules.

- Update `LessCompiler` to delegate to `ipl\Web\Less`, assembling sources via `@import (less)` directives; pass the `$minify` flag directly to `render()` instead of via a separate `compress()` call.

- Modernize `LessCompiler` with PHP type declarations, short array syntax, and tightened PHPDoc.

**For testing**, use 7da4917 as the baseline, which adapts light mode rule evaluation order to match the new visitor.

Here is the test summary:

The stylesheets are nearly identical. Most changes are normalization noise introduced by the new version of the wikimedia Less parser:

| Category                  | Example                                      |
|---------------------------|----------------------------------------------|
| Decimal canonicalization  | `.5` → `0.5`, `2.0em` → `2em`                |
| Hex color shortening      | `#dddddd` → `#ddd`                           |
| Hex color case kept as is | `#535353` → `#535353`, `#c4c4c4` → `#c4c4c4` |
| `calc()` math preserved   | `calc(14.28571429%)` → `calc(100% / 7)`      |

Starting with Less.php 5.0, arithmetic is no longer evaluated inside `calc()`. This required changing one occurrence in our Less files directly; no other changes were necessary, and no known publicly available Icinga Web module is affected by similar changes in this regard.

There is one functional difference: the light-mode background color gains an extra level of indirection via a new CSS var call (because it uses `@light-body-bg-color`):

```diff
-  --body-bg-color: #f5f9fa;
+  --body-bg-color: var(--light-body-bg-color, #f5f9fa);
```

The new CSS var replacement visitor from `ipl\Web` runs independently from the light mode call visitor, and therefore also adds var overrides to such detached rulesets. Everything else is normalization noise.

closes #5457
requires Icinga/icinga-php-thirdparty#115
requires Icinga/ipl-web#365


library/Icinga/Application/webrouter.php
```diff
@@ -53,6 +53,7 @@
 );
 
 if (in_array($path, $special)) {
+    putenv('ICINGAWEB_ENVIRONMENT=production');
```
lippserd: Please note that this must be removed before merging the PR.

sukhwinder33445:
I compared the minified CSS output between the `main` branch and the `less` branch (both icingaweb2 and ipl-web) and found the following issues:

1. **Decimal leading zero**: The change is applied in the wrong direction: `0.5em` is being converted to `.5em`, but it should be the other way around (`.5em → 0.5em`) as you mentioned above.
2. **Missing semicolon**: CSS properties before the closing brace (`}`) are missing their trailing semicolon (`;`).
3. **Empty CSS values now throw an error**: previously, empty string values were rendered without issue. For example, when trying to configure a new schedule in the `Notifications Web` module, the following error is thrown:
Screenshot 2026-04-14 at 16 08 38

[Here](https://github.com/Icinga/icinga-notifications-web/blob/81b52406d1905668e61d1060a3e935c288492966/library/Notifications/Widget/TimeGrid/BaseGrid.php#L496) the `getColor()` returns empty string in this case.

4. **LESS division not evaluated**: [A LESS expression](https://github.com/Icinga/icingaweb2/blob/1d6fa0aabd2e7d01726520168a09b063d8eeed75/public/css/icinga/configmenu.less#L270) that previously compiled correctly to `max-height: calc(100vh - 4.16666667em);` now outputs invalid CSS: `max-height: calc(100vh - 50/12em);`. [According to the LESS docs](https://lesscss.org/#operations-calc-exception), to evaluate the division, a LESS variable must be used.

lippserd:
> 1. **Decimal leading zero**: The change is applied in the wrong direction: `0.5em` is being converted to `.5em`, but it should be the other way around (`.5em → 0.5em`) as you mentioned above.

Are you sure about that?

> 2\. **Missing semicolon**: CSS properties before the closing brace (`}`) are missing their trailing semicolon (`;`).

Why do you think that's an issue? 

> 3\. **Empty CSS values now throw an error**: previously, empty string values were rendered without issue. For example, when trying to configure a new schedule in the `Notifications Web` module [...]

That's interesting (since this results in `property: ;` during rendering, which is obviously just invalid), and I wonder what happened before — were the rules simply ignored? Anyway, PHPDoc specifies that a string must be passed, and the notification module violates that. That needs to be fixed there. I'll consider adapting `Style` /`LessRuleset` accordingly so that its contract is actually enforced.

> 4\. **LESS division not evaluated**: [A LESS expression](https://github.com/Icinga/icingaweb2/blob/1d6fa0aabd2e7d01726520168a09b063d8eeed75/public/css/icinga/configmenu.less#L270) that previously compiled correctly to `max-height: calc(100vh - 4.16666667em);` now outputs invalid CSS: `max-height: calc(100vh - 50/12em);`. [According to the LESS docs](https://lesscss.org/#operations-calc-exception), to evaluate the division, a LESS variable must be used.

That's a good find, and it's been invalid CSS/Less since it was introduced. Now that you've brought it up, would you like to add a fix to this PR?

sukhwinder33445:
> Are you sure about that?

Yes, looked again at the before and after CSS minified files and found that in most cases the leading zero has been removed, but in some places the change is applied as you mentioned above, and in others it remains unchanged. So we have three different behaviors, I think there should be consistency here.
####  Removed leading zero:

```css
/*Before*/
.icinga-module.module-icingadb .item-layout.downtime .visual {
    ...
    padding: 0.5em 0.25em;
    /* The source file (`icingadb/downtime.less`) does not contain the leading zero, so I guess the LESS compiler was adding it before. */
}

/*After*/
.icinga-module.module-icingadb .item-layout.downtime .visual {
    ...
    padding: .5em .25em;
}
```
#### A zero is added:

```css
 /*Before*/
.icinga-module.module-icingadb .row-item .performance-data {
    ...
    margin-left: -.25em;
}

 /*After*/
.icinga-module.module-icingadb .row-item .performance-data {
    ...
    margin-left: -0.25em
}
```
#### Remains the same in before and after cases:
```css
.icinga-module.module-icingadb .row-item .performance-data .inline-pie {
   ...    
   margin-left: .25em
}
```

> Why do you think that's an issue?

Yes, it is not an issue, but I thought it could be an unwanted change since it was not mentioned in the changes table above. Or did you mention it as normalization noise?

> That's a good find, and it's been invalid CSS/Less since it was introduced. Now that you've brought it up, would you like to add a fix to this PR?

Okey, i will push a commit.

lippserd:
> > Are you sure about that?
> 
> Yes, looked again at the before and after CSS minified files [...]

We have no control over how the Wikimedia Less parser minifies the CSS. Therefore, it only makes sense to compare the unminified versions.

> > Why do you think that's an issue?
> 
> Yes, it is not an issue, but I thought it could be an unwanted change since it was not mentioned in the changes table above. Or did you mention it as normalization noise?

Same as above.

sukhwinder33445:
Yes, you are right. That's the minifier which removes the leading zero and the semicolon. The non-minified file contains only the changes mentioned above.

library/Icinga/Web/LessCompiler.php
```diff
@@ -121,137 +88,96 @@ public function getLessFiles()
...
+     * @param ?string $theme Path to the Less theme file, or null to unset
      *
-     * @return  $this
+     * @return $this
      */
-    public function setTheme($theme)
+    public function setTheme(?string $theme): static
     {
         if ($theme === null || (is_file($theme) && is_readable($theme))) {
             $this->theme = $theme;
```
nilmerg: I wonder why we didn't use `realpath` here as we use it for other less files already. Now that you use `@import`, I think it's safer to use it for theme files as well.

library/Icinga/Web/LessCompiler.php
```diff
@@ -121,137 +88,96 @@ public function getLessFiles()
...
+     * @param string $themeMode Path to the Less theme mode file
      *
-     * @return  $this
+     * @return $this
      */
-    public function setThemeMode($themeMode)
+    public function setThemeMode(string $themeMode): static
     {
         if (is_file($themeMode) && is_readable($themeMode)) {
             $this->themeMode = $themeMode;
```
nilmerg: Same here, `realpath` won't hurt.

library/Icinga/Web/LessCompiler.php
```diff
@@ -121,137 +88,96 @@ public function getLessFiles()
...
+
+        $lightModeTemplate = <<<'LESS'
+@media (min-height: @prefer-light-color-scheme), print,
+(prefers-color-scheme: light) and (min-height: @enable-color-preference) {
+    {ruleset}
+}
+LESS;
+        $compiler = new WikimediaLessCompiler([
+            // Disable URL adjustment - Less.php tries to make URLs relative to the file path when using @import.
+            'relativeUrls' => false,
```
nilmerg, line 189: What's the exact reason for this? I enabled it again but didn't spot an issue. And looking at wikimedia's implementation, I think it's safer to enable since [it escapes certain characters it seems](https://github.com/wikimedia/less.php/blob/70ad62a1a38b35733950ddb9a8cf1eaf27785e4e/lib/Less/Tree/Url.php#L46).

nilmerg (commented):
> |Category|Example|
> |-------|-------|
> | Font shorthand explicit line-height | 1em → 1em/1 | #

That's not caused by normalization but [by not evaluating it as division](https://github.com/Icinga/ipl-web/blob/d430b0d553240e2a2060c20517132a4344f6f838/asset/css/awesome-icons.less#L2). But these variables don't seem to be used anyway, so that's why it wasn't an issue previously and the change won't be noticed now.

library/Icinga/Web/LessCompiler.php
lippserd: You should see issues, e.g., fonts not loading, logo not loading.

lippserd:
> > Font shorthand explicit line-height
> > 1em → 1em/1
> 
> That's not caused by normalization but [by not evaluating it as division](https://github.com/Icinga/ipl-web/blob/d430b0d553240e2a2060c20517132a4344f6f838/asset/css/awesome-icons.less#L2). But these variables don't seem to be used anyway, so that's why it wasn't an issue previously and the change won't be noticed now.

The new version of the Wikimedia Less parser handles mathematical expressions differently, which is why we have to specify `'math' => 'always'`. Nevertheless, this particular expression is no longer evaluated.

library/Icinga/Web/LessCompiler.php
nilmerg: Right. Then the comment explaining it is misleading as it doesn't affect `@import` in any way. Even tried to import a realative path with this enabled and it worked fine.

nilmerg:
The upgrading docs need to mention the changes, especially those to mathematical expressions.

lippserd:
> The upgrading docs need to mention the changes, especially those to mathematical expressions.

What should actually be included in the upgrading docs?

Users don't need to take any action. The Less integration continues to work without requiring any changes on their side. We also don't currently document Less at all: there's neither general guidance nor any stated supported version.

The only issue we encountered was an incorrect math expression passed to `calc()`. However, none of the publicly available Icinga Web modules use similar patterns. I reviewed all of them, and their `calc()` usage is valid.

Regarding the explicit `line-height` in font declarations: this appears to be an inconsistency in Less itself (or specifically in `wikimedia/less.php`). For example, with:

```less
:font {
  --fa-font-regular: normal 400 1em/1 "Font Awesome 6 Free";
}

.font {
  font: normal 400 1em/1 "Font Awesome 6 Free";
}
```

Using `"wikimedia/less.php": "^3.2.1"` produces:

```css
:font {
  --fa-font-regular: normal 400 1em "Font Awesome 6 Free";
}
.font {
  font: normal 400 1em/1 "Font Awesome 6 Free";
}
```

Here, the explicit line-height is preserved only in the `.font` rule.

With `"wikimedia/less.php": "^5.5"`:

```css
:font {
  --fa-font-regular: normal 400 1em/1 "Font Awesome 6 Free";
}
.font {
  font: normal 400 1em/1 "Font Awesome 6 Free";
}
```

In this case, the explicit line-height is also preserved in the custom property under `:font`.

Finally, regarding `'math' => 'always'`: this reflects the old default behavior. The new default `parens-division` would not evaluate most of our expressions because they are not wrapped in parentheses.

library/Icinga/Web/LessCompiler.php
lippserd: Refined the comment.

nilmerg:
> What should actually be included in the upgrading docs? Users don't need to take any action.

Granted. But we should document that we use a modernized version of the less-php parser as that may hint about new features not available beforehand.

library/Icinga/Application/webrouter.php
lippserd: Removed.

library/Icinga/Web/LessCompiler.php
lippserd: Done.

library/Icinga/Web/LessCompiler.php
lippserd: Done.

nilmerg (approved):
Since the compiler now throws an exception in case a theme file is not found, I tried to enable a theme of a module, renamed the theme file and disabled the module. Both cases did not throw an error and only the first one was logged as previously.

Please update the OP as you meant it is outdated as of now.

lippserd:
> Please update the OP as you meant it is outdated as of now.

Done.

PR #5497, opened 2026-04-28
Simplify installation command for Ubuntu

Al2Klimov:
The additional package libapache2-mod-php is already a dependency, hence doesn't have to be specified explicitly.

# Test

✅ `apt-get install --no-install-{recommends,suggests} icingaweb2` will also install `libapache2-mod-php`


sukhwinder33445 (approved):

lippserd:
Superseded by #5519.

PR #5512, opened 2026-05-21
modal.js: Do not wobble if ESC has been handled otherwise

nilmerg:
To reproduce the prior behavior, open a search editor, request e.g. column suggestions and push ESC.

prior: The modal wobbles
now: Nothing happens

### DIFF
--- a/public/js/icinga/behavior/modal.js
+++ b/public/js/icinga/behavior/modal.js
@@ -208,7 +208,7 @@
      * @param event {KeyboardEvent} The `keydown` event triggered by pushing a key
      */
     Modal.prototype.onEscapeKey = function(event) {
-        if (event.key !== 'Escape') {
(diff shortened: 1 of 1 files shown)


sukhwinder33445 (commented):
This works fine.
I noticed when adding new fields to SearchEditor (without filling them in) and pressing Escape, the modal closes because no input change is detected.
Shouldn't new empty OR deleted fields also count as unsaved changes? 
On the other hand, the user may also intentionally want to exit the modal and discard their changes, so the current behavoir is fine.

nilmerg:
adding empty ones isn't a worthy change in my eyes, removing a condition on the other hand is so you have a point here. But this change is about something else ;)

sukhwinder33445 (approved):

PR #5514, opened 2026-05-26
Two-Factor Authentication Contract

jrauh01:
Introduce a `TwoFactor` interface and `TwoFactorHook` abstract base class that modules can extend to provide 2FA methods. The core handles the full flow around them:

- **Login**: after password auth, users enrolled in a registered method are redirected to a token challenge page before the session is established
- **Enrollment**: a new "Two-Factor Auth" tab in account settings lets users enroll in or unenroll from any registered method
- **API**: HTTP Basic auth requests are rejected with 403 when the user has a 2FA credential, since the token step cannot be completed over Basic auth for now
- **Remember me**: cookies are revoked on enrollment and unenrollment. During a pending challenge the AES key is kept in the database and only the ciphertext is stored in the session, so the two halves are never co-located.

To provide a 2FA method, a module extends `TwoFactorHook`, implements the `TwoFactor` interface, and calls `TwoFactorHook::register()` in the `run.php`.

require https://github.com/Icinga/ipl-web/pull/393

### DIFF
--- a/application/controllers/AuthenticationController.php
+++ b/application/controllers/AuthenticationController.php
@@ -187,7 +187,8 @@ class AuthenticationController extends CompatController
     public function twofactorAction(): void
     {
         $session = Session::getSession();
-        if (! (new TwoFactorState($session))->isChallenged()) {
+        $twoFactorState = new TwoFactorState($session);
+        if (! $twoFactorState->isChallenged()) {
             $this->redirectToLogin();
         }
 
@@ -199,7 +200,7 @@ class AuthenticationController extends CompatController
                     $this->redirectNow($redirectUrl);
                 }
             })
-            ->on(Form::ON_SENT, function (TwoFactorChallengeForm $form) use ($session) {
+            ->on(Form::ON_SENT, function (TwoFactorChallengeForm $form) use ($session, $twoFactorState) {
                 // ON_SENT because cancel is not the primary submit button and never
                 // triggers ON_SUBMIT. CSRF is checked manually. Without it a forged
                 // request could destroy the session and drop the 2FA challenge.
@@ -208,6 +209,15 @@ class AuthenticationController extends CompatController
                     $form->getPressedSubmitElement()?->getName() === TwoFactorChallengeForm::SUBMIT_CANCEL;
 
                 if ($csrfValid && $cancelPressed) {
+                    // The login flow may have persisted a remember-me record before
+                    // issuing the challenge. Remove it now since the challenge was
+                    // canceled and the cookie was never delivered to the browser.
+                    if ($cookieData = $twoFactorState->getRememberMeCookieData()) {
+                        $data = explode('|', $cookieData);
+                        $iv = base64_decode(array_pop($data));
+                        (new RememberMe())->remove(bin2hex($iv));
+                    }
+
                     $session->purge();
                     $this->redirectNow($this->withRedirect('authentication/login'));
                 }
--- a/application/controllers/AuthenticationController.php
+++ b/application/controllers/AuthenticationController.php
@@ -103,7 +103,7 @@ class AuthenticationController extends CompatController
                     $this->httpBadRequest('Redirect to an external host is not allowed');
                 }
             } else {
-                $redirectUrl = $form->createRedirectUrl();
+                $redirectUrl = Url::fromPath(LoginForm::REDIRECT_URL);
             }
 
             $this->redirectNow($redirectUrl);
--- a/application/controllers/AuthenticationController.php
+++ b/application/controllers/AuthenticationController.php
@@ -12,11 +12,14 @@ use Icinga\Application\Icinga;
 use Icinga\Application\Logger;
 use Icinga\Authentication\LoginButton;
 use Icinga\Authentication\LoginButtonForm;
+use Icinga\Authentication\TwoFactorState;
 use Icinga\Common\Database;
 use Icinga\Exception\AuthenticationException;
 use Icinga\Forms\Authentication\LoginForm;
+use Icinga\Forms\Authentication\TwoFactorChallengeForm;
 use Icinga\Web\Helper\CookieHelper;
 use Icinga\Web\RememberMe;
+use Icinga\Web\Session;
 use Icinga\Web\Url;
 use Icinga\Web\Widget\LoginPage;
 use ipl\Html\Contract\Form;
@@ -47,6 +50,10 @@ class AuthenticationController extends CompatController
      */
     public function loginAction()
     {
+        if ((new TwoFactorState(Session::getSession()))->isChallenged()) {
+            $this->redirectNow($this->withRedirect('authentication/twofactor'));
+        }
+
         $icinga = Icinga::app();
         if (($requiresSetup = $icinga->requiresSetup()) && $icinga->setupTokenExists()) {
             $this->redirectNow(Url::fromPath('setup'));
@@ -139,7 +146,7 @@ class AuthenticationController extends CompatController
             }
         }
 
-        // Suppress the rendering of controls bar
+        // Suppress the rendering of controls bar.
         $this->view->compact = true;
         $this->setTitle($this->translate('Icinga Web 2 Login'));
         $this->addContent(new LoginPage($form, $loginButtons, $requiresSetup));
@@ -170,5 +177,67 @@ class AuthenticationController extends CompatController
 
             $this->redirectToLogin();
         }
+    }
+
+    /**
+     * Render the two-factor authentication challenge page
+     *
+     * @return void
+     */
+    public function twofactorAction(): void
+    {
+        $session = Session::getSession();
+        if (! (new TwoFactorState($session))->isChallenged()) {
+            $this->redirectToLogin();
+        }
+
+        $form = (new TwoFactorChallengeForm())
+            ->setCsrfCounterMeasureId($session->getId())
+            ->setAction(Url::fromRequest()->getAbsoluteUrl())
+            ->on(Form::ON_SUBMIT, function (TwoFactorChallengeForm $form) {
+                if ($redirectUrl = $form->getRedirectUrl()) {
+                    $this->redirectNow($redirectUrl);
+                }
+            })
+            ->on(Form::ON_SENT, function (TwoFactorChallengeForm $form) use ($session) {
+                // ON_SENT because cancel is not the primary submit button and never
+                // triggers ON_SUBMIT. CSRF is checked manually. Without it a forged
+                // request could destroy the session and drop the 2FA challenge.
+                $csrfValid = $form->getElement('CSRFToken')->isValid();
+                $cancelPressed =
+                    $form->getPressedSubmitElement()?->getName() === TwoFactorChallengeForm::SUBMIT_CANCEL;
+
+                if ($csrfValid && $cancelPressed) {
+                    $session->purge();
+                    $this->redirectNow($this->withRedirect('authentication/login'));
+                }
+            })
+            ->handleRequest($this->getServerRequest());
+
+        $this->setTitle($this->translate('Icinga Web 2 Two-Factor Auth'));
+
+        // Suppress the rendering of controls bar.
+        $this->view->compact = true;
+        $this->addContent(new LoginPage($form));
+    }
+
+    /**
+     * Build a URL for the given path that carries forward the redirect destination
+     *
+     * Copies the `redirect` query parameter from the current request, if present, so the
+     * post-authentication destination is not lost during multi-step authentication transitions.
+     *
+     * @param string $path Path to build the URL from
+     *
+     * @return Url
+     */
+    protected function withRedirect(string $path): Url
+    {
+        $url = Url::fromPath($path);
+        if ($redirect = $this->params->get('redirect')) {
+            $url->setParam('redirect', $redirect);
+        }
+
+        return $url;
     }
 }
--- a/application/controllers/MyDevicesController.php
+++ b/application/controllers/MyDevicesController.php
@@ -23,30 +23,27 @@ class MyDevicesController extends CompatController
     public function init()
     {
         $this->getTabs()
-            ->add(
-                'account',
-                [
-                    'title' => $this->translate('Update your account'),
-                    'label' => $this->translate('My Account'),
-                    'url'   => 'account'
-                ]
-            )
-            ->add(
-                'navigation',
-                [
-                    'title'     => $this->translate('List and configure your own navigation items'),
-                    'label'     => $this->translate('Navigation'),
-                    'url'       => 'navigation'
-                ]
-            )
-            ->add(
-                'devices',
-                [
-                    'title' => $this->translate('List of devices you are logged in'),
-                    'label' => $this->translate('My Devices'),
-                    'url'   => 'my-devices'
-                ]
-            )->activate('devices');
+            ->add('account', [
+                'title' => $this->translate('Update your account'),
+                'label' => $this->translate('My Account'),
+                'url'   => 'account',
+            ])
+            ->add('navigation', [
+                'title' => $this->translate('List and configure your own navigation items'),
+                'label' => $this->translate('Navigation'),
+                'url'   => 'navigation',
+            ])
+            ->add('devices', [
+                'title' => $this->translate('List of devices you are logged in'),
+                'label' => $this->translate('My Devices'),
+                'url'   => 'my-devices',
+            ])
+            ->add('two-factor', [
+                'title' => $this->translate('Configure two-factor authentication'),
+                'label' => $this->translate('Two-Factor Auth'),
+                'url'   => 'two-factor/config',
+            ])
+            ->activate('devices');
     }
 
     public function indexAction()
--- a/application/controllers/TwoFactorController.php
+++ b/application/controllers/TwoFactorController.php
@@ -1,0 +1,100 @@
+<?php
+
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+namespace Icinga\Controllers;
+
+use Icinga\Application\Hook\TwoFactorHook;
+use Icinga\Application\Logger;
+use Icinga\Exception\IcingaException;
+use Icinga\Authentication\Auth;
+use Icinga\Forms\Account\TwoFactorEnrollmentForm;
+use Icinga\Web\Session;
+use ipl\Web\Common\CalloutType;
+use ipl\Web\Widget\Callout;
+use Throwable;
+use ipl\Html\Contract\Form;
+use ipl\Html\HtmlElement;
+use ipl\Html\Text;
+use ipl\Web\Compat\CompatController;
+
+/**
+ * Two-factor authentication configuration
+ */
+class TwoFactorController extends CompatController
+{
+    public function init(): void
+    {
+        $this->getTabs()
+            ->add('account', [
+                'title' => $this->translate('Update your account'),
+                'label' => $this->translate('My Account'),
+                'url'   => 'account',
+            ])
+            ->add('navigation', [
+                'title' => $this->translate('List and configure your own navigation items'),
+                'label' => $this->translate('Navigation'),
+                'url'   => 'navigation',
+            ])
+            ->add('devices', [
+                'title' => $this->translate('List of devices you are logged in'),
+                'label' => $this->translate('My Devices'),
+                'url'   => 'my-devices',
+            ])
+            ->add('two-factor', [
+                'title' => $this->translate('Configure two-factor authentication'),
+                'label' => $this->translate('Two-Factor Auth'),
+                'url'   => 'two-factor/config',
+            ]);
+    }
+
+    /**
+     * Render the two-factor authentication configuration page
+     *
+     * Shows an informational notice when no two-factor method is registered. Otherwise,
+     * renders {@see TwoFactorEnrollmentForm}, letting the user enroll in or unenroll from
+     * registered two-factor methods.
+     *
+     * @return void
+     */
+    public function configAction(): void
+    {
+        $this->getTabs()->activate('two-factor');
+        $this->addContent(HtmlElement::create('h1', null, Text::create($this->translate('Two-Factor Authentication'))));
+
+        if (TwoFactorHook::all() === []) {
+            $this->addContent(new Callout(CalloutType::Info, $this->translate(
+                'No two-factor authentication method is available. Enable a module that provides'
+                . ' an implementation to configure two-factor authentication for your account.',
+            )));
+
+            return;
+        }
+
+        $user = Auth::getInstance()->getUser();
+
+        try {
+            $enrolledMethodName = TwoFactorHook::loadEnrolled($user)?->getName();
+        } catch (Throwable $e) {
+            Logger::error("%s\n%s", $e->getMessage(), IcingaException::getConfidentialTraceAsString($e));
+            $this->addContent(new Callout(CalloutType::Error, sprintf(
+                $this->translate('Two-factor authentication is currently unavailable: %s. Contact your administrator.'),
+                $e->getMessage(),
+            )));
+
+            return;
+        }
+
+        $enrollmentForm = (new TwoFactorEnrollmentForm($user, $enrolledMethodName))
+            ->setCsrfCounterMeasureId(Session::getSession()->getId())
+            ->on(Form::ON_SUBMIT, function (TwoFactorEnrollmentForm $form) {
+                if ($redirectUrl = $form->getRedirectUrl()) {
+                    $this->redirectNow($redirectUrl);
+                }
+            })
+            ->handleRequest($this->getServerRequest());
+
+        $this->addContent($enrollmentForm);
+    }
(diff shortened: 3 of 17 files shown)


application/controllers/AuthenticationController.php
```diff
@@ -171,4 +179,53 @@ public function logoutAction()
...
+                // destroy the session and drop the 2FA challenge.
+                $csrfValid = $form->getElement('CSRFToken')->isValid();
+                $cancelPressed =
+                    $form->getPressedSubmitElement()?->getName() === TwoFactorChallengeForm::SUBMIT_CANCEL;
+
+                if ($csrfValid && $cancelPressed) {
+                    // The login flow may have persisted a remember-me record before issuing the
+                    // challenge. Remove it now since the challenge was canceled and the cookie
+                    // was never delivered to the browser.
+                    RememberMe::removeAllByUsername((new TwoFactorState())->getChallengedUser()->getUsername());
```
raviks789: Why not just use already defined `$twoFactorState`. Instead of a separate `new TwoFactorState()`.

```suggestion
            ->on(Form::ON_SENT, function (TwoFactorChallengeForm $form) use ($twoFactorState) {
                // ON_SENT because cancel is not the primary submit button and never triggers
                // ON_SUBMIT. CSRF is checked manually. Without it a forged request could
                // destroy the session and drop the 2FA challenge.
                $csrfValid = $form->getElement('CSRFToken')->isValid();
                $cancelPressed =
                    $form->getPressedSubmitElement()?->getName() === TwoFactorChallengeForm::SUBMIT_CANCEL;

                if ($csrfValid && $cancelPressed) {
                    // The login flow may have persisted a remember-me record before issuing the
                    // challenge. Remove it now since the challenge was canceled and the cookie
                    // was never delivered to the browser.
                    RememberMe::removeAllByUsername($twoFactorState->getChallengedUser()->getUsername());
```

application/forms/Account/TwoFactorEnrollmentForm.php
```diff
@@ -0,0 +1,196 @@
...
+        if ($configFieldset->getName() !== $twoFactor->getName()) {
+            throw new LogicException(sprintf(
+                '%s::assembleEnrollmentFormElements() must not rename the fieldset. The name "%s" is used'
+                . ' as the element key in onSuccess() to retrieve the fieldset, but it was changed to "%s".',
+                get_class($twoFactor),
+                $twoFactor->getName(),
+                $configFieldset->getName()
+            ));
+        }
+        $configFieldset->addAttributes(Attributes::create([
```
raviks789, line 117: Add a new line after if block

application/forms/Authentication/LoginForm.php
```diff
@@ -151,6 +130,30 @@ protected function onSuccess(): void
         $password = $this->getElement('password')->getValue();
         $authenticated = $authChain->authenticate($user, $password);
         if ($authenticated) {
+            if (null !== $twoFactorMethod = TwoFactorHook::loadEnrolled($user)) {
```
raviks789: For readability I would recommend:


```suggestion
            $twoFactorMethod = TwoFactorHook::loadEnrolled($user)
            if ($twoFactorMethod !== null) {
```

application/forms/Authentication/TwoFactorChallengeForm.php
```diff
@@ -0,0 +1,177 @@
...
+            $this->addMessage($this->translate(
+                'Your two-factor authentication method is no longer available.'
+                . ' If this is unexpected, contact your administrator.'
+                . ' Otherwise use \'Back to login\' — you will not be prompted for a second factor.'
+            ));
+            $this->onError();
+
+            return;
+        }
+        try {
```
raviks789, line 145: Add a newline after if block

application/forms/Authentication/LoginForm.php
```diff
@@ -151,6 +130,30 @@ protected function onSuccess(): void
...
+                    $rememberMe = RememberMe::fromCredentials($user->getUsername(), $password);
+                    $twoFactorState->setRememberMeCookieData($rememberMe->getCookie()->getValue());
+                    $rememberMe->persist();
+                }
+
+                $redirectUrl = Url::fromPath('authentication/twofactor');
+                if ($redirect = Url::fromRequest()->getParam('redirect')) {
+                    $redirectUrl->setParam('redirect', $redirect);
+                }
+                $this->setRedirectUrl($redirectUrl);
```
raviks789, line 174: Add a new line after if-block

library/Icinga/Application/Hook/TwoFactorHook.php
```diff
@@ -0,0 +1,149 @@
...
+     *
+     * @return list<static>
+     */
+    public static function all(): array
+    {
+        $twoFactorMethods = [];
+        foreach (Hook::all(static::NAME) as $method) {
+            $twoFactorMethods[$method->getName()] = $method;
+        }
+        ksort($twoFactorMethods, SORT_STRING | SORT_FLAG_CASE);
```
raviks789: Add a new line after foreach block.

library/Icinga/Authentication/Auth.php
```diff
@@ -293,6 +294,16 @@ public function authHttp()
...
+                Logger::warning(
+                    'API request rejected for user "%s": two-factor authentication cannot be completed via the API',
+                    $user->getUsername()
+                );
+                $this->getResponse()->json()
+                    ->setHttpResponseCode(403)
+                    ->setErrorMessage('Two-factor authentication is required and cannot be completed via the API')
+                    ->sendResponse();
+            }
             $this->setAuthenticated($user, false);
```
raviks789, line 394: Same here, add a new line

raviks789 (changes requested):

application/forms/Account/TwoFactorEnrollmentForm.php
jrauh01: Done

application/forms/Authentication/TwoFactorChallengeForm.php
jrauh01: Done

application/forms/Authentication/LoginForm.php
jrauh01: Done

library/Icinga/Application/Hook/TwoFactorHook.php
jrauh01: Done

library/Icinga/Authentication/Auth.php
jrauh01: Done

application/controllers/AuthenticationController.php
```diff
@@ -96,7 +104,7 @@ public function loginAction()
                     $this->httpBadRequest('Redirect to an external host is not allowed');
                 }
             } else {
-                $redirectUrl = $form->createRedirectUrl();
+                $redirectUrl = Url::fromPath(LoginForm::REDIRECT_URL);
             }
```
lippserd, line 107: Why is this code path not using the new LoginRedirect element?

application/controllers/AuthenticationController.php
```diff
@@ -171,4 +179,53 @@ public function logoutAction()
...
+                    RememberMe::removeAllByUsername($twoFactorState->getChallengedUser()->getUsername());
+                    Session::getSession()->purge();
+                    $this->redirectNow($this->withRedirect('authentication/login'));
+                }
+            })
+            ->handleRequest($this->getServerRequest());
+
+        $this->setTitle($this->translate('Icinga Web 2 Two-Factor Auth'));
+
+        // Suppress the rendering of controls bar
```
lippserd: Also add a dot.

application/controllers/AuthenticationController.php
```diff
@@ -171,4 +179,53 @@ public function logoutAction()
...
+        $this->view->compact = true;
+        $this->addContent(new LoginPage($form));
+    }
+
+    protected function withRedirect(string $path): Url
```
lippserd, line 244: Please document this method.

application/controllers/MyDevicesController.php
```diff
@@ -23,30 +23,27 @@ class MyDevicesController extends CompatController
     public function init()
     {
         $this->getTabs()
```
lippserd, line 25: Can the tabs be a trait or a static public property or function? If I am not mistaken we have multiple places which state the same tabs.

application/forms/Authentication/LoginForm.php
```diff
@@ -44,6 +46,9 @@ class LoginForm extends CompatForm
     public function __construct()
     {
         $this->setAttribute('name', 'form_login');
+        // Gives this form a unique CSS path so that loader.js does not restore focus to the submit
+        // button of a subsequently rendered form whose button would otherwise share the same path.
+        $this->setAttribute('id', 'login-form-focus-anchor');
```
lippserd: Why not just `login-form`?

application/forms/Authentication/TwoFactorChallengeForm.php
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * Create a new TwoFactorChallengeForm
+     */
+    public function __construct()
+    {
+        $this->setAttribute('name', 'form_twofactor_challenge');
+        // Gives this form a unique CSS path so that loader.js does not incorrectly restore focus to
+        // this form's submit button after the login form submits and transitions to this view.
+        $this->setAttribute('id', 'twofactor-challenge-form-focus-anchor');
```
lippserd: Why not just `2fa-challenge-form`?

application/forms/Authentication/TwoFactorChallengeForm.php
```diff
@@ -0,0 +1,178 @@
...
+                    $rememberMe = RememberMe::fromCookieData($cookieData);
+                    $response->setCookie($rememberMe->getCookie());
+                } catch (Exception $e) {
+                    Logger::error('Failed to let user "%s" stay logged in: %s', $user->getUsername(), $e);
+                }
+            }
+
+            $twoFactorState->completeChallenge();
+
+            // Call provided AuthenticationHook(s) after successful login
```
lippserd: Add a dot.

doc/05-Authentication.md
```diff
@@ -291,3 +291,33 @@ asks that backend to authenticate the user with the sAMAccountName "jdoe".
 When the user "jdoe@icinga.com" logs in, Icinga Web 2 walks through all configured authentication backends until it
 finds one which is responsible for that user -- e.g. a MariaDB or MySQL backend (SQL database backends aren't domain-aware). Then
 Icinga Web 2 asks that backend to authenticate the user with the username "jdoe@icinga.com".
+
+## Two-Factor Authentication <a id="two-factor-authentication"></a>
+
+Icinga Web 2 supports two-factor authentication (2FA) through an implementable hook. When a module that provides a
```
lippserd: In new prose/product naming please always omit the 2 from Icinga Web.

lippserd: And please replace the em-dashes throughout ;).

lippserd (changes requested):

application/forms/Authentication/LoginForm.php
```diff
@@ -151,6 +130,32 @@ protected function onSuccess(): void
...
+                    'User "%s" has been challenged for two-factor verification using method "%s"',
+                    $user->getUsername(),
+                    $twoFactorMethod->getName()
+                );
+
+                if ($this->getElement('rememberme')->isChecked()) {
+                    $rememberMe = RememberMe::fromCredentials($user->getUsername(), $password);
+                    $twoFactorState->setRememberMeCookieData($rememberMe->getCookie()->getValue());
+                    $rememberMe->persist();
+                }
```
lippserd, line 165: The non-2fa branch already logs a remember-me failure and continues authenticating the user. The same behavior should apply before the 2fa challenge: remember-me failure should only disable "stay logged in" for that attempt, not abort login.

I would persist before storing the cookie data in `TwoFactorState`. Otherwise a failed persist can leave challenge state containing cookie data that cannot be resolved later in `TwoFactorChallengeForm::onSuccess()`. If I'm not mistaken the non-2fa branch should be adjusted in this regard as well, i.e. persist before setting the response cookie.

```suggestion
                if ($this->getElement('rememberme')->isChecked()) {
                    $rememberMe = RememberMe::fromCredentials($user->getUsername(), $password);
                    $rememberMe->persist();
                    $twoFactorState->setRememberMeCookieData($rememberMe->getCookie()->getValue());
                } catch (Throwable $e) {
                    Logger::error('Failed to let user "%s" stay logged in: %s', $user->getUsername(), $e);
                }
```

lippserd (changes requested):

application/controllers/MyDevicesController.php
jrauh01: They are used in four controllers. Three use it in their init, but the fourth (`NavigationController`) can have different tabs and adds them in it's action. I don't think it's in the scope of this PR to refactor this, however we could think of doing this in another PR.

application/forms/Authentication/LoginForm.php
jrauh01: I talked about this with @nilmerg and he wanted a self-describing name that states the reason why the id is necessary at all, if I'm remember correctly.

application/forms/Authentication/TwoFactorChallengeForm.php
jrauh01: Same reason as above.

application/controllers/AuthenticationController.php
jrauh01: It's explained in the commit introducing `LoginRedirect`. Short version: because it would always resolve to `LoginForm::REDIRECT_URL`. So why put it behind an extra abstraction layer? This wasn't necessary even before this PR and could have always been like this.

application/controllers/AuthenticationController.php
jrauh01: Done

application/controllers/AuthenticationController.php
jrauh01: Done

application/forms/Authentication/TwoFactorChallengeForm.php
jrauh01: Done

application/forms/Authentication/LoginForm.php
jrauh01: Sounds good. Please don't wonder why I don't apply your suggestion, but I'm not familiar with an `if-catch` statement. 😆

jrauh01: Done

doc/05-Authentication.md
jrauh01: Done

application/controllers/AuthenticationController.php
lippserd: What about replacing lines 100-110 with:

```php
$loginRedirect = new LoginRedirect('redirect', ['value' => $this->params->get('redirect')]);
try {
    $redirectUrl = $loginRedirect->getUrl();
} catch (HttpBadRequestException $e) {
    $this->httpBadRequest($e->getMessage());
}
$this->redirectNow($redirectUrl);
```

This would also forbid authentication/logout which the current path lacks.

application/controllers/MyDevicesController.php
lippserd: Ok.

application/forms/Authentication/LoginForm.php
lippserd: > Sounds good. Please don't wonder why I don't apply your suggestion, but I'm not familiar with an `if-catch` statement. 😆

It's PHP 9. Just wait for it! 😆 Just kidding :).

application/forms/Authentication/LoginForm.php
lippserd: Don't know if the comment was there before or whether you just added it, but it already states the intent. Calling it a "CSS path" is a bit loose when the code is assigning a unique DOM id, so that phrasing mixed implementation details in a way that reads slightly contradictory. Both id additions could be committed together, so the overall intent would be recoverable from the commit history.

Given how specific the id looks, I expect someone will grep for it later when debugging or looking at the code. At least I did.

Here is my suggestion:

```php
// Use a unique id so loader.js doesn't restore focus to the submit button
// of a subsequently rendered form that would match the same selector.
$this->setAttribute('id', 'login-form');
```

application/controllers/AuthenticationController.php
jrauh01: I played with this a bit and I think it is safe. However, I think it's kind of awkward to use a form element for that in this context, which `LoginRedirect` is since it's extending `HiddenElement`. What are your thoughts about this?

library/Icinga/Authentication/TwoFactorState.php
```diff
@@ -0,0 +1,103 @@
...
+    /** @var string Session key storing the raw remember-me cookie */
+    protected const SESSION_REMEMBER_ME_COOKIE_DATA = 'remember_me_cookie_data';
+
+    /** @var SessionNamespace Session namespace scoping the challenge state */
+    protected SessionNamespace $session;
+
+    /**
+     * Create a new TwoFactorState
+     */
+    public function __construct()
```
TheSyscall: This class is just helper functions that interact with the current session. Consider taking in a session object in the ctor instead of accessing the singleton.

application/forms/Authentication/LoginForm.php
jrauh01: I will rephrase that! I'm not quite sure about your "commiting together" point, as the additions of the ids in both `LoginForm` and `TwoFactorChallengeForm` already are in the same commit. Or do you mean something else?

lippserd: I mean only the id additions in a single commit plus justification. I didn't check whether this is already the case.

jrauh01: Yes it's https://github.com/Icinga/icingaweb2/commit/2681dc293f65a6148e5447259107a78111ef6988.

jrauh01: Adjusted to your suggestion.

library/Icinga/Authentication/TwoFactor.php
```diff
@@ -0,0 +1,92 @@
...
+     * Check whether a user is enrolled in this 2FA method
+     *
+     * Implementations typically query the database for a stored credential (secret, key, ...).
+     *
+     * @param ?User $user The user to check for. If null, the currently
+     *   authenticated user is resolved via {@see Auth::getUser()}
+     *
+     * @return bool True if enrolled, false otherwise
+     */
+    public function isEnrolled(?User $user = null): bool;
```
TheSyscall: Consider always requiring a User. This would stop forcing implementations to store/access authentication state.

library/Icinga/Authentication/TwoFactor.php
```diff
@@ -0,0 +1,92 @@
...
+     * Verify a 2FA token provided by the user
+     *
+     * Called during login to verify the two-factor challenge. Implementations should
+     * call this from within {@see enroll()} to confirm the credential works.
+     *
+     * @param string $token The raw token string entered by the user, e.g. a 6-digit TOTP code
+     *
+     * @return bool True if the token is valid, false otherwise
+     */
+    public function verify(string $token): bool;
```
TheSyscall: Related to above, pass the User to act on as a parameter.

library/Icinga/Authentication/TwoFactor.php
```diff
@@ -0,0 +1,92 @@
...
+     * that the credential works, and store it on success.
+     *
+     * @param FieldsetElement $fieldset The method-specific fieldset containing the submitted
+     *   credential elements
+     *
+     * @return bool True if the credential was verified and stored, false if verification
+     *   failed. On failure, attach a user-visible error to the relevant element in
+     *   $fieldset via {@see FieldsetElement::addMessage()}.
+     */
+    public function enroll(FieldsetElement $fieldset): bool;
```
TheSyscall: Related to above, pass the User to act on as a parameter.

library/Icinga/Authentication/TwoFactor.php
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * Remove the stored credential for the currently authenticated user
+     *
+     * Called from {@see TwoFactorEnrollmentForm::onSuccess()} when the unenroll button is pressed.
+     *
+     * @return void
+     */
+    public function unenroll(): void;
```
TheSyscall: Related to above, pass the User to act on as a parameter.

This would allow a button in an admin interface to unenroll a user.

TheSyscall (changes requested):

library/Icinga/Authentication/TwoFactorState.php
jrauh01: Done

library/Icinga/Authentication/TwoFactor.php
jrauh01: Done

library/Icinga/Authentication/TwoFactor.php
jrauh01: Done

library/Icinga/Authentication/TwoFactor.php
jrauh01: Done

library/Icinga/Authentication/TwoFactor.php
jrauh01: Done

jrauh01:
Additionally to the contract adjustments I have adjusted the call sites and doc blocks.

application/controllers/TwoFactorController.php
```diff
@@ -65,7 +66,7 @@ public function configAction(): void
             return;
         }
 
-        $enrolledMethodName = TwoFactorHook::loadEnrolled()?->getName();
+        $enrolledMethodName = TwoFactorHook::loadEnrolled(Auth::getInstance()->getUser())?->getName();
 
         $chooseMethodForm = (new TwoFactorEnrollmentForm())
```
TheSyscall: This form does more than chose the method. The whole enrollment form is part of it. Consider renaming it to `enrolementForm`

application/forms/Account/TwoFactorEnrollmentForm.php
```diff
@@ -108,7 +108,7 @@ protected function assemble(): void
         $configFieldset = new FieldsetElement($twoFactor->getName());
         $this->addElement($configFieldset);
         try {
-            $twoFactor->assembleEnrollmentFormElements($configFieldset);
+            $twoFactor->assembleEnrollmentFormElements(Auth::getInstance()->getUser(), $configFieldset);
```
TheSyscall: It might make sense to have the User be passed in by the controller.

application/forms/Account/TwoFactorEnrollmentForm.php
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        $this->enrolled = $enrolled;
+
+        return $this;
+    }
+
+    protected function assemble(): void
+    {
+        $this->addCsrfCounterMeasure(Session::getSession()->getId());
```
TheSyscall: `addCsrfCounterMeasure` is typically called without a parameter and the id is set in the controller.

TheSyscall (changes requested):

application/controllers/TwoFactorController.php
jrauh01: Renamed to $enrollmentForm

application/forms/Account/TwoFactorEnrollmentForm.php
jrauh01: Done

application/forms/Account/TwoFactorEnrollmentForm.php
jrauh01: Done

application/forms/Authentication/TwoFactorChallengeForm.php
```diff
@@ -0,0 +1,178 @@
...
+    {
+        $this->setAttribute('name', 'form_twofactor_challenge');
+        // Use a unique id so loader.js doesn't restore focus to the submit button
+        // of a subsequently rendered form that would match the same selector.
+        $this->setAttribute('id', 'twofactor-challenge-form');
+    }
+
+    protected function assemble(): void
+    {
+        $this->addCsrfCounterMeasure(Session::getSession()->getId());
```
TheSyscall: Call without a parameter and set the id inside the controller.

Could have noted that before, my bad.

TheSyscall (changes requested):

application/forms/Authentication/TwoFactorChallengeForm.php
jrauh01: No problem! Done.

TheSyscall (dismissed):
LGTM 👍

application/controllers/AuthenticationController.php
lippserd: Let's keep it as is for now. When we update the partnerportal, we may need to adjust the `LoginRedirect` element, since the module uses a custom redirect URL. We could then move the redirect logic into a helper class or method, which could then be used here.

lippserd (approved):

PR #5515, opened 2026-05-27
css; fix the negative margin for .control-group

DsgnrFH:
Replaces `margin: 1em -1em 1em 0` with `margin: 1em 0em` for consistent spacing and no negative margin for both better practice and future features.

### DIFF
--- a/public/css/icinga/forms.less
+++ b/public/css/icinga/forms.less
@@ -21,9 +21,7 @@ form.icinga-form {
     flex-wrap: wrap;
     align-items: flex-start;
 
-    // Negative margin-right because every child gets 1em right but we can't exclude
-    // the last element as it's impossible to identify the last *visible* element
-    margin: 1em -1em 1em 0;
+    margin: 1em 0;
 
     > fieldset {
       > .control-group:first-of-type {


nilmerg:
> for both better practice and future features

Please elaborate. Right now this just introduces an offset of form elements to other containers for no reason.

| Before | After |
| -- | -- |
| image | image |

DsgnrFH:
The demotour module needs the form elements aligned with their container, and the negative margin makes the element stick out (in my case out of the modal). Beyond that, I'd generally avoid negative margins where we can. If changing the shared style is too broad, we can modify only the CSS at the demo.icinga.com server and not include this change globally

nilmerg:
> aligned with their container

That's what the negative margin is intended for, at least visually.

> the negative margin makes the element stick out (in my case out of the modal)

My examples were from a form inside a modal, so I don't think that's a general issue.

> I'd generally avoid negative margins where we can

That's fine, but here we cannot :D (without changing the markup, hence the comment you removed)

DsgnrFH:
Alright, fair enough

nilmerg:
Closing as I didn't hear from you anymore that my alternative proposal doesn't work.

DsgnrFH:
Yup, I have found a workaround that works for my purposes and doesn't affect Icingaweb2 globally

Issue #5521, opened 2026-06-14
Migration DB privilege check fails for database names containing _


leeclemens:
## Describe the bug
The migration framework reports that the configured database user is missing privileges - "The provided credentials cannot be used to execute the required SQL commands…"  even when the user has a valid database-level GRANT. It only happens when the database name contains an underscore (or possibly %). Databases without those characters (e.g. icingaweb, director) validate fine, which is the giveaway.

## To Reproduce
1. Create a database whose name contains an underscore, e.g. icinga_reporting.
2. Grant the web user database-level privileges on it: GRANT ... ON `icinga\_reporting`.* TO icingaweb'@'localhost';
3. Run the setup wizard's database validation, or apply a pending migration for that resource.
4. The privilege check fails even though the grant is correct.

## Expected behavior
The check passes, since the user genuinely holds the required privileges.

## Screenshots
N/A

## Your Environment
Icinga Web 2 version: 2.13.0

## Additional context
In modules/setup/library/Setup/Utils/DbTool.php, checkMysqlPrivileges() runs the database name through escapeTableWildcards() (_ → \_) before binding it into the information_schema.schema_privileges query, which compares with =:

```
AND table_schema = :dbname   -- :dbname bound as 'icinga\_reporting'
```

Escaping only has meaning for LIKE. With =, the escaped value icinga\_reporting never equals the real schema name icinga_reporting, so the database-level check returns 0. The table_privileges query just below already binds the unescaped name, so the two arms are inconsistent. Removing the escapeTableWildcards() call from the = comparisons fixes it; the GRANT … ON statement should keep it, since GRANT does treat _/% as wildcards.

nilmerg:
Please read https://github.com/Icinga/icingaweb2/issues/3476#issuecomment-397964811 and make sure the grant you issued manually did not result in an unescaped (or improper) database wildcard.

PR #5522, opened 2026-06-14
Fix DB privilege check for database names containing `_` or `%`

leeclemens:
Fixes the DB privilege check failing for database names with _ or %. The schema_privileges queries escaped the db name (_ → \_) before an = comparison — escaping only works with LIKE, so it
  never matched. Bind the raw name instead, consistent with the table_privileges query right below.

refs #5521


leeclemens:
Signed, not sure which option to click on the generic Contact Us page.

bobapple:
@cla-bot check

PR #5524, opened 2026-06-22
`menu.less`: Add style for init setup icon for notifications web

sukhwinder33445:
This PR adds style for the menu icon of the icinga-notifications-web module, which appears when the initial setup is required.

### DIFF
--- a/public/css/icinga/menu.less
+++ b/public/css/icinga/menu.less
@@ -611,3 +611,20 @@ html.no-js  #toggle-sidebar {
     }
   }
 }
+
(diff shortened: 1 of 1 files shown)


lippserd (commented):
Even if this is just a PR draft, please include a description. It might be a good idea to add `module-*` classes to the navigation items provided by modules so that they can apply custom CSS instead of relying on changes in Icinga Web itself.

sukhwinder33445:
The notifications module must define the style.

PR #5533, opened 2026-07-06
Fix config form value & Simplify CspConfigForm

TheSyscall:
`ConfigForm`'s `ON_ELEMENT_REGISTERED` handler skipped seeding an element with its config value whenever the element already had a value set (`$element->hasValue()`), even on initial render. When it did seed the value, it called `$element->setValue()` directly instead of `$this->populate()`, so the value was visible on the element but not through `getPopulatedValue()`.

This broke forms that read defaults via `getPopulatedValue()` during rendering, most notably `CspConfigForm`, where checkboxes ended up rendering with the wrong checked state.

## Changes

- `ConfigForm`: replace the `ON_ELEMENT_REGISTERED` handler with an override of `registerElement()`. The config value is populated as the element's base value, so `getPopulatedValue()` reflects it, while any value populated earlier (e.g. a submitted one) still takes precedence.
- `ConfigForm`: remember each element's original (default) value. On save, a key whose submitted value equals its default is unset instead of written, and an element with a non-empty default that is submitted empty is stored as an empty string.
- `CspConfigForm`: move the CSP auto-generation default (off for installations that already had CSP enabled, on for new ones) out of `ConfigController` and into `addDirectiveCheckboxElement()`, which now takes the config key and derives the field name itself.
- `ConfigController`: drop the manual default computation and populate call, and construct the form as a fluent chain.
- Tests for the new population chain and for `CspConfigForm`'s defaults.

require Icinga/ipl-html#205

### DIFF
--- a/application/forms/Config/Security/CspConfigForm.php
+++ b/application/forms/Config/Security/CspConfigForm.php
@@ -163,186 +163,188 @@ class CspConfigForm extends ConfigForm
             $this->addElement('hidden', 'security__csp_enable_modules');
             $this->addElement('hidden', 'security__csp_enable_dashboards');
             $this->addElement('hidden', 'security__csp_enable_navigation');
-        } else {
-            $this->addHtml((new DisplayFormElement(
-                HtmlElement::create('div', null, [
-                    HtmlElement::create('h3', ['class' => $formHintClassList], $this->translate('Allowed Sources')),
-                    HtmlElement::create('p', ['class' => $formHintClassList], $this->translate(
-                        'Sources that are used in the generation of the CSP header.',
-                    )),
-                    HtmlElement::create('h4', ['class' => $formHintClassList], $this->translate('System')),
-                ]),
-            ))->addAttributes(Attributes::create(['class' => 'csp-control-group'])));
 
-            $this->addDirectiveContentElement(
-                [Csp::getSystemCsp()],
-                [$this->translate('Directive'), $this->translate('Value')],
-                function (StaticCspReason $reason, string $directive, string $expression) {
-                    return Table::tr([Table::td($directive), $this->buildExpression($directive, $expression)]);
-                },
-                ! $useCustomCsp,
-                $this->translate('No system policies defined.'),
-            );
-
-            $this->addDirectiveCheckboxElement(
-                $this->translate('Enable Modules'),
-                $this->translate(
-                    'Should module defined CSP directives be enabled?'
-                    . ' Modules can define or change csp directives at any point.',
-                ),
-                'csp_enable_modules',
-                ! $useCustomCsp,
-            );
-
-            $this->addDirectiveContentElement(
-                (new ModuleCspLoader())->loadForAllUsers(),
-                [$this->translate('Module'), $this->translate('Directive'), $this->translate('Value')],
-                function (ModuleCspReason $reason, string $directive, string $expression) {
-                    return Table::tr([
-                        Table::td($reason->module),
-                        Table::td($directive),
-                        $this->buildExpression($directive, $expression),
-                    ]);
-                },
-                ! $useCustomCsp && $this->getValue('security__csp_enable_modules') === '1',
-                $this->translate('No module policies defined.'),
-            );
-
-            $this->addDirectiveCheckboxElement(
-                $this->translate('Enable Dashboards'),
-                $this->translate(
-                    'Enable user defined dashboards. This table contains all dashboards for all users. The actual'
-                    . ' header that is sent to the user will only contain the subset of directives that actually'
-                    . ' matters to them.',
-                ),
-                'csp_enable_dashboards',
-                ! $useCustomCsp,
-            );
-
-            $this->addDirectiveContentElement(
-                (new DashboardCspLoader())->loadForAllUsers(),
-                [
-                    $this->translate('Dashboard'),
-                    $this->translate('Dashlet'),
-                    $this->translate('User'),
-                    $this->translate('Directive'),
-                    $this->translate('Value'),
-                ],
-                function (DashboardCspReason $reason, string $directive, string $expression) {
-                    return Table::tr([
-                        Table::td($reason->pane->getName()),
-                        Table::td($reason->dashlet->getName()),
-                        Table::td($reason->dashboard->getUser()->getUsername()),
-                        Table::td($directive),
-                        $this->buildExpression($directive, $expression),
-                    ]);
-                },
-                ! $useCustomCsp && $this->getValue('security__csp_enable_dashboards') === '1',
-                $this->translate('No dashboard policies found.'),
-            );
-
-            $this->addDirectiveCheckboxElement(
-                $this->translate('Enable Navigation Items'),
-                $this->translate(
-                    'Enable user defined navigation items. This table contains all navigation items for'
-                    . ' all users. The actual header that is sent to the user will only contain the subset of'
-                    . ' directives that actually matters to them.',
-                ),
-                'csp_enable_navigation',
-                ! $useCustomCsp,
-            );
-
-            $this->addDirectiveContentElement(
-                (new NavigationCspLoader())->loadForAllUsers(),
-                [
-                    $this->translate('Navigation'),
-                    $this->translate('Parent'),
-                    $this->translate('Name'),
-                    $this->translate('User'),
-                    $this->translate('Directive'),
-                    $this->translate('Value'),
-                ],
-                function (NavigationCspReason $reason, string $directive, string $expression) {
-                    $parentCell = $reason->parent === null
-                        ? Table::td($this->translate('None'))->setAttribute('class', 'empty-state')
-                        : Table::td($reason->parent);
-
-                    $sharedIcon = match ($reason->isShared) {
-                        true  => new Icon('share', [
-                            'class' => 'shared-item',
-                            'title' => $this->translate('Shared item. Displayed user is owner.'),
-                        ]),
-                        false => null,
-                    };
-
-                    $userCell = $reason->username === null
-                        ? Table::td([$sharedIcon, $this->translate('Unknown')])->setAttribute('class', 'empty-state')
-                        : Table::td([$sharedIcon, $reason->username]);
-
-                    return Table::tr([
-                        Table::td($reason->typeConfiguration['label'] ?? $reason->type),
-                        $parentCell,
-                        Table::td($reason->name),
-                        $userCell,
-                        Table::td($directive),
-                        $this->buildExpression($directive, $expression),
-                    ]);
-                },
-                ! $useCustomCsp && $this->getValue('security__csp_enable_navigation') === '1',
-                $this->translate('No navigation policies found.'),
-            );
-
-            $this->addElement('checkbox', 'security__use_custom_csp', [
-                'label'          => $this->translate('Enable Custom CSP'),
-                'description'    => $this->translate(
-                    'Specify whether to use a custom, user-provided string as the CSP header.',
-                ),
-                'class'          => 'autosubmit csp-form-content-aligned csp-label-header-h3 csp-form-header',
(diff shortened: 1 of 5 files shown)


library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -50,35 +49,29 @@ class ConfigForm extends CompatForm
...
+     * @return $this
+     */
+    public function registerElement(FormElement $element)
+    {
+        $name = $element->getName();
+        [$section, $key] = Str::symmetricSplit($name, $this->sectionKeyDelimiter, 2);
+        if ($key !== null) {
+            $submitted = $this->getPopulatedValue($name);
+            $this->clearPopulatedValue($name);
+            $this->populate([$name => $this->config->get($section, $key)]);
```
jrauh01: If the config key does not exist yet, the populated values would currently retrieve `null`. In `parent::registerElement()` there is a check on `array_key_exists()` which would return `true` even if the value is null. This would lead to default values being overridden with `null` which would render an empty element instead of the default value.

```suggestion
            $configValue = $this->config->get($section, $key);
            if ($configValue !== null) {
                $this->populate([$name => $configValue]);
            }

File: library/Icinga/Web/Form/ConfigForm.php [hunk 2]
```diff
@@ -50,34 +49,29 @@ class ConfigForm extends CompatForm
...
 
-                return;
+    /**
+     * Register the given element, seeding it with its configuration value first
+     *
+     * @param FormElement $element
+     *
+     * @return $this
+     */
+    public function registerElement(FormElement $element)
```
jrauh01: Consider adding `static` as return type, even if the parent has no declared return type.

application/forms/Config/Security/CspConfigForm.php
```diff
@@ -427,7 +427,7 @@ protected function addDirectiveCheckboxElement(
             'checkedValue'   => '1',
             'uncheckedValue' => '0',
             'disabled'       => ! $enabled,
-            'value'          => $this->getPopulatedValue($field),
+            'value'          => $this->isCspEnabled() ? '0' : '1',
```
jrauh01: You had a comment for this in `ConfigController` which you removed now. Wouldn't it make sense to add it here to keep the logic explained?

jrauh01 (changes requested):

library/Icinga/Web/Form/ConfigForm.php
TheSyscall: Changed

library/Icinga/Web/Form/ConfigForm.php
TheSyscall: Changed

application/forms/Config/Security/CspConfigForm.php
TheSyscall: Added comment

jrauh01 (changes requested):
I think it would make sense to add a test case to avoid the "default override" error in future changes.

test/php/library/Icinga/Web/Form/ConfigFormTest.php
```diff
@@ -82,6 +82,19 @@ public function exposeSave(): void
...
+    {
+        $form = new class(Config::fromArray([])) extends ConfigForm {
+            protected function assemble(): void
+            {
+                $this->addElement('text', 'mysection__key', ['value' => 'defaultvalue']);
+            }
+        };
+        $form->ensureAssembled();
+
+        $this->assertSame('defaultvalue', $form->getElement('mysection__key')->getValue());
```
jrauh01: Why not use this?

```suggestion
        $this->assertSame('defaultvalue', $form->getValue('mysection__key'));

**Review by jrauh01 — CHANGES_REQUESTED** [review]

**Review by jrauh01 — DISMISSED** [review]
LGTM

**Comment by jrauh01** [comment]
The unit tests will fail until icinga/ipl-html#205 is merged, due to the missing `FormElements::getPopulatedValues` method.

**Comment by jrauh01** [comment]
I removed the CSRF fix as it will be fixed with #5552.

File: library/Icinga/Web/Form/ConfigForm.php [hunk 5]
```diff
@@ -8,14 +8,16 @@
...
-use ipl\Html\Contract\ValueCandidates;
+use ipl\Html\Form;
 use ipl\Html\HtmlElement;
 use ipl\Html\ValidHtml;
 use ipl\Stdlib\Str;
 use ipl\Web\Compat\CompatForm;
 use ipl\Web\Compat\DisplayFormElement;
 use ipl\Web\Widget\CopyToClipboard;
 use LogicException;
+use ReflectionProperty;
```
lippserd: `ipl\Html\Form` and `ReflectionProperty` are remnants of the superseded
reflection implementation and are unused now. Remove them.

```suggestion
use ipl\Html\HtmlElement;
use ipl\Html\ValidHtml;
use ipl\Stdlib\Str;
use ipl\Web\Compat\CompatForm;
use ipl\Web\Compat\DisplayFormElement;
use ipl\Web\Widget\CopyToClipboard;
use LogicException;
```

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -50,34 +62,35 @@ class ConfigForm extends CompatForm
...
+    {
+        $name = $element->getName();
+        [$section, $key] = Str::symmetricSplit($name, $this->sectionKeyDelimiter, 2);
+        if ($key !== null) {
+            $this->originalValues[$name] = $element->getValue();
+
+            $sentinel = new stdClass();
+            $configValue = $this->config->get($section, $key, $this->originalValues[$name] ?? $sentinel);
+            if ($configValue !== $sentinel) {
+                $populatedValues = $this->getPopulatedValues($name);
```
lippserd, line 92: `getPopulatedValues()` is unavailable in Icinga PHP Library 1.0.0, while
the source installation requirements and setup wizard still accept that
version. Require `icinga-php-library >= 1.1.0` everywhere it is declared,
even though 1.1.0 has not been released yet.

test/php/library/Icinga/Web/Form/ConfigFormTest.php
```diff
@@ -75,17 +81,117 @@ public function exposeSave(): void
...
+            protected function assemble(): void
+            {
+                $this->addElement('text', 'mysection__key', ['value' => 'defaultvalue']);
+            }
+        };
+        $form->ensureAssembled();
+
+        $this->assertSame('defaultvalue', $form->getValue('mysection__key'));
+        $this->assertSame('defaultvalue', $form->getPopulatedValue('mysection__key'));
+    }
```
lippserd, line 167: This verifies the fallback when configuration is absent, but not the
regression that motivated the change: a nonempty element default must not
mask a different configured value. Add that case.

```suggestion
    }

    public function testConfiguredValueOverridesElementDefault(): void
    {
        $config = Config::fromArray(['mysection' => ['key' => 'configured']]);
        $form = new class ($config) extends ConfigForm {
            protected function assemble(): void
            {
                $this->addElement('text', 'mysection__key', ['value' => 'defaultvalue']);
            }
        };
        $form->ensureAssembled();

        $this->assertSame('configured', $form->getValue('mysection__key'));
        $this->assertSame('configured', $form->getPopulatedValue('mysection__key'));
    }
```

test/php/library/Icinga/Web/Form/ConfigFormTest.php
```diff
@@ -99,10 +205,108 @@ public function exposeSave(): void
...
+            }
+        };
+        $this->populateAroundAssembly(
+            $form,
+            [['mysection__key' => ''], ['mysection__key' => '0']],
+            $populateBeforeAssembly,
+        );
+
+        $this->assertSame('0', $form->getValue('mysection__key'));
+        $this->assertSame('0', $form->getPopulatedValue('mysection__key'));
```
lippserd, line 354: The test currently observes only the final value, so it would still pass if
the configured baseline or intermediate empty value were lost. Assert the
complete chain for both population timings.

```suggestion
        $this->assertSame('0', $form->getPopulatedValue('mysection__key'));
        $this->assertSame(
            ['configured', '', '0'],
            $form->getPopulatedValues('mysection__key'),
        );
```

lippserd (changes requested):
The form-value handling is coherent, particularly the preserved population order and expanded save coverage. I left four inline requests to make the dependency contract explicit, remove obsolete imports, and cover the central regressions directly.

test/php/library/Icinga/Web/Form/ConfigFormTest.php
jrauh01: Applied by myself, but inlined `$config` as in the other tests.

library/Icinga/Web/Form/ConfigForm.php
jrauh01: Done

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -57,40 +67,40 @@ class ConfigForm extends CompatForm
...
+     */
+    public function registerElement(FormElement $element): static
+    {
+        $name = $element->getName();
+        [$section, $key] = Str::symmetricSplit($name, $this->sectionKeyDelimiter, 2);
+        if ($key !== null) {
+            $this->originalValues[$name] = $element->getValue();
+
+            $sentinel = new stdClass();
+            $configValue = $this->config->get($section, $key, $this->originalValues[$name] ?? $sentinel);
```
lippserd: `Config::get()` also substitutes its non-null fallback for an existing key
whose value is `null`, so this sentinel still conflates a configured `null`
with a missing key and contradicts the commit's stated behavior. Read through
`getSection($section)->get()` instead, whose `array_key_exists()` check
preserves `null`, and add a regression test with a non-empty element default
plus a configured `null`.

```suggestion
$configValue = $this->config
    ->getSection($section)
    ->get($key, $this->originalValues[$name] ?? $sentinel);
```

test/php/application/forms/Config/Security/CspConfigFormTest.php
```diff
@@ -0,0 +1,74 @@
...
+                'csp_enable_navigation' => '1',
+            ],
+        ]);
+
+        $this->assertSame('1', $form->getValue('security__csp_enable_modules'));
+        $this->assertSame('1', $form->getValue('security__csp_enable_dashboards'));
+        $this->assertSame('1', $form->getValue('security__csp_enable_navigation'));
+    }
+
+    private function assertDirectiveDefaults(array $configData, string $expected): void
```
lippserd: Please declare the nested fixture shape on this helper so PHPStan and the
PHPDoc check can validate its input.

```suggestion
/**
 * @param array{security?: array<string, string>} $configData
 */
private function assertDirectiveDefaults(array $configData, string $expected): void
```

test/php/application/forms/Config/Security/CspConfigFormTest.php
```diff
@@ -0,0 +1,74 @@
...
+    {
+        $form = $this->createDirectiveForm($configData);
+
+        foreach (['csp_enable_modules', 'csp_enable_dashboards', 'csp_enable_navigation'] as $key) {
+            $this->assertSame($expected, $form->getValue('security__' . $key), $key);
+        }
+    }
+
+    private function createDirectiveForm(array $configData): CspConfigForm
```
lippserd: Please declare the same fixture shape on the factory helper.

```suggestion
/**
 * @param array{security?: array<string, string>} $configData
 */
private function createDirectiveForm(array $configData): CspConfigForm
```

lippserd (changes requested):

library/Icinga/Web/Form/ConfigForm.php
jrauh01: Added besides the test.

lippserd (approved):

PR #5536, opened 2026-07-07
Modernize repository form

jrauh01:
`\Icinga\Forms\RepositoryForm` has been the base class for repository-backed forms. This PR introduces a `CompatForm`-based replacement.

The new `\Icinga\Web\Form\RepositoryForm` takes a `RepositoryMode` enum at construction time, rather than having callers set the mode via `add()`, `edit()`, or `remove()`. The mode controls which elements are assembled, how the form is pre-populated, and which repository operation fires on success. Subclasses implement `assembleInsertElements()`, `assembleDeleteElements()`, and `createFilter()`. `assembleUpdateElements()` falls back to `assembleInsertElements()` by default but can be overridden when the update layout differs. The notification handling is now in the response of the controller.

### DIFF
--- a/library/Icinga/Web/Form/RepositoryForm.php
+++ b/library/Icinga/Web/Form/RepositoryForm.php
@@ -1,0 +1,337 @@
+<?php
+
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+namespace Icinga\Web\Form;
+
+use Icinga\Data\Extensible;
+use Icinga\Data\Filter\Filter;
+use Icinga\Data\Reducible;
+use Icinga\Data\Updatable;
+use Icinga\Exception\NotFoundError;
+use Icinga\Repository\Repository;
+use Icinga\Repository\RepositoryMode;
+use InvalidArgumentException;
+use ipl\Html\Contract\Form;
+use ipl\Web\Common\CsrfCounterMeasure;
+use ipl\Web\Common\FormUid;
+use ipl\Web\Compat\CompatForm;
+use Psr\Http\Message\ServerRequestInterface;
+use Stringable;
+
+/**
+ * Form base-class providing standard functionality for extensible, updatable and
+ * reducible repositories
+ *
+ * CSRF protection is enabled by default. Call {@see setCsrfCounterMeasureId()}
+ * to set the CSRF form element id or {@see disableCsrfCounterMeasure()}
+ * to opt out. The form UID element requires a form name. Set one via the
+ * `name` attribute before assembling.
+ */
+abstract class RepositoryForm extends CompatForm
+{
+    use CsrfCounterMeasure;
+    use FormUid;
+
+    /** @var ?array<string, mixed> The data of the entry to pre-populate the form with */
+    protected ?array $data = null;
+
+    /**
+     * Create a new repository form
+     *
+     * Applies the default element decorators and calls the mode-appropriate
+     * {@see onInsertRequest()}, {@see onUpdateRequest()}, or {@see onDeleteRequest()}
+     * when the form receives a request. $repository must implement the correct
+     * interface for the respective $mode:
+     *
+     * - {@see Extensible} for {@see RepositoryMode::Insert}
+     * - {@see Updatable} for {@see RepositoryMode::Update}
+     * - {@see Reducible} for {@see RepositoryMode::Delete}
+     *
+     * @param Repository $repository The repository to work with
+     * @param RepositoryMode $mode How to interact with the repository
+     * @param string|Stringable|int|null $identifier The identifier of the entry to handle.
+     *   Required for {@see RepositoryMode::Update} and {@see RepositoryMode::Delete}.
+     *
+     * @throws InvalidArgumentException If $repository does not implement the interface
+     *   required by $mode, or if $identifier does not match $mode's requirements
+     */
+    public function __construct(
+        protected Repository $repository,
+        protected RepositoryMode $mode,
+        protected string|Stringable|int|null $identifier = null,
+    ) {
+        $requiredInterface = $this->mode->getRequiredInterface();
+
+        if (! $this->repository instanceof $requiredInterface) {
+            throw new InvalidArgumentException(sprintf(
+                'Repository "%s" does not implement %s, which is required for %s mode',
+                $this->repository::class,
+                $requiredInterface,
+                $this->mode->name
+            ));
+        }
+
+        if ($this->mode->requiresIdentifier() && $this->identifier === null) {
+            throw new InvalidArgumentException(sprintf('An identifier is required for %s mode', $this->mode->name));
+        }
+
+        $this->applyDefaultElementDecorators();
+
+        $this->on(Form::ON_REQUEST, function (ServerRequestInterface $_, RepositoryForm $form): void {
+            match ($this->mode) {
+                RepositoryMode::Insert => $form->onInsertRequest(),
+                RepositoryMode::Update => $form->onUpdateRequest(),
+                RepositoryMode::Delete => $form->onDeleteRequest(),
+            };
+        });
+    }
+
+    /**
+     * Create and add elements to this form
+     *
+     * @return void
+     */
+    protected function assemble(): void
+    {
+        $this->addCsrfCounterMeasure();
+        $this->addElement($this->createUidElement());
+
+        match ($this->mode) {
+            RepositoryMode::Insert => $this->assembleInsertElements(),
+            RepositoryMode::Update => $this->assembleUpdateElements(),
+            RepositoryMode::Delete => $this->assembleDeleteElements(),
+        };
+    }
+
+    /**
+     * Apply the requested mode on the repository
+     *
+     * @return void
+     */
+    protected function onSuccess(): void
+    {
+        match ($this->mode) {
+            RepositoryMode::Insert => $this->insertEntry(),
+            RepositoryMode::Update => $this->updateEntry(),
+            RepositoryMode::Delete => $this->deleteEntry(),
+        };
+    }
+
+    /**
+     * Get the name of the entry to handle
+     *
+     * @return string|Stringable|int|null
+     */
+    public function getIdentifier(): string|Stringable|int|null
+    {
+        return $this->identifier;
+    }
+
+    /**
+     * Set the current data of the entry being handled
+     *
+     * In case of {@see RepositoryMode::Insert} the data will be used as default values.
+     * In case of {@see RepositoryMode::Update} the data is the current entry's values.
+     *
+     * @param array<string, mixed> $data Entry data or default values
+     *
+     * @return $this
+     */
+    public function setData(array $data): static
+    {
+        $this->data = $data;
+
+        return $this;
+    }
+
+    /**
+     * Get the extensible repository
+     *
+     * @return Repository&Extensible
+     */
+    protected function getExtensibleRepository(): Repository&Extensible
+    {
+        /** @var Repository&Extensible $repo */
+        $repo = $this->repository;
+
+        return $repo;
+    }
+
+    /**
+     * Get the updatable repository
+     *
+     * @return Repository&Updatable
+     */
+    protected function getUpdatableRepository(): Repository&Updatable
+    {
+        /** @var Repository&Updatable $repo */
+        $repo = $this->repository;
+
+        return $repo;
+    }
+
+    /**
+     * Get the reducible repository
+     *
+     * @return Repository&Reducible
+     */
+    protected function getReducibleRepository(): Repository&Reducible
+    {
+        /** @var Repository&Reducible $repo */
+        $repo = $this->repository;
+
+        return $repo;
+    }
+
+    /**
+     * Fetch and return the entry to pre-populate the form with when in mode update
+     *
+     * @return object|false False in case of no result
+     */
+    protected function fetchEntry(): object|false
+    {
+        return $this->repository
+            ->select()
+            ->addFilter($this->createFilter())
+            ->fetchRow();
+    }
+
+    /**
+     * Whether the entry supposed to be removed exists
+     *
+     * @return bool
+     */
+    protected function entryExists(): bool
+    {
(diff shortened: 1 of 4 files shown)


application/forms/RepositoryForm.php
```diff
@@ -14,6 +14,9 @@
 
 /**
  * Form base-class providing standard functionality for extensible, updatable and reducible repositories
+ *
+ * @deprecated Use {@see \Icinga\Web\Form\RepositoryForm} instead for
+ * {@see \ipl\Web\Compat\CompatForm} based implementation.
```
TheSyscall:
```suggestion
 * @deprecated Use {@see \Icinga\Web\Form\RepositoryForm} instead for a
 * {@see \ipl\Web\Compat\CompatForm} based implementation.
```

doc/60-Hooks.md
```diff
@@ -1,53 +1,5 @@
 # Hooks
 
-## ConfigFormEventsHook
-
```
TheSyscall, line 4: The hook isn't gone. Would it make sense to add a deprecation notice to the documentation as well?

library/Icinga/Web/Form/RepositoryForm.php
```diff
@@ -0,0 +1,372 @@
...
+     */
+    protected function getData(): ?array
+    {
+        return $this->data;
+    }
+
+    /**
+     * Fetch and return the entry to pre-populate the form with when in mode update
+     *
+     * @return object|false
```
TheSyscall: Describe the meaning of the return values when using union types

library/Icinga/Web/Form/RepositoryForm.php
```diff
@@ -0,0 +1,372 @@
...
+    {
+        return $this->repository
+            ->select()
+            ->from($this->getBaseTable())
+            ->applyFilter($this->createFilter())
+            ->fetchRow();
+    }
+
+    /**
+     * Return whether the entry supposed to be removed exists
```
TheSyscall:
```suggestion
     * Whether the entry supposed to be removed exists
```

library/Icinga/Web/Form/RepositoryForm.php
```diff
@@ -0,0 +1,372 @@
...
+     * @return void
+     *
+     * @throws Exception If inserting the entry fails
+     */
+    protected function onInsertSuccess(): void
+    {
+        try {
+            $this->insertEntry();
+        } catch (Exception $e) {
+            Notification::error($this->getInsertMessage(false));
```
TheSyscall: We should rethink if we need notifications for form success and error events.
I would expect these messages to be displayed in the controller or at least made conditional.

Alternatively add more methods that can be overwritten to hook into the form lifecycle. Something like `onInsertComplete(bool $success): void` so you have more control over what is happening inside this method. Maybe even display a `Notification`

library/Icinga/Web/Form/RepositoryForm.php
```diff
@@ -0,0 +1,372 @@
...
+
+        Notification::success($this->getDeleteMessage(true));
+    }
+
+    /**
+     * Create and add elements to this form to insert an entry
+     *
+     * @return void
+     */
+    abstract protected function createInsertElements(): void;
```
TheSyscall:
```suggestion
    abstract protected function assembleInsertElements(): void;
```

library/Icinga/Web/Form/RepositoryForm.php
```diff
@@ -0,0 +1,372 @@
...
+
+    /**
+     * Create and add elements to this form to update an entry
+     *
+     * Calls createInsertElements() by default. Overwrite this to add different elements when in mode update.
```
TheSyscall:
```suggestion
     * Calls {@see createInsertElements()} by default. Overwrite this to add different elements when in mode update.
```

TheSyscall (changes requested):

doc/60-Hooks.md
jrauh01: I talked about this with @lippserd. As we don't want users to implement it anymore, there is no reason to keep documentation about it.

library/Icinga/Web/Form/RepositoryForm.php
jrauh01: I kept the behavior of the old repository form in the first place, but I'm not opposed to rethink this as well.

library/Icinga/Web/Form/RepositoryForm.php
jrauh01: Adjusted this and every other occurrence of `create*Elements()` methods.

library/Icinga/Web/Form/RepositoryForm.php
jrauh01: Done

library/Icinga/Web/Form/RepositoryForm.php
jrauh01: I refactored the notifications handling. The controller is now responsible for adding notifications via the `ON_SUBMIT` or `ON_ERROR` events as `\ipl\Html\Form::handleRequest()` already catches exceptions thrown in `onSuccess()`. The repository form itself does now not send notifications per default anymore and calls `insertEntry()` / `updateEntry()` / or `deleteEntry()` directly in `onSuccess()`. The `get*Message()` and `on*Success()` methods are removed as they are now unused.

jrauh01:
The following changes have been made additionally to the last review:
- The `on*Request()` methods are now registered in the constructor
- `assemble()` and `onSuccess()` have been moved up
- Notifications are now in response of the controller

TheSyscall (dismissed):
Consider renaming `assemble*Elements` to just `assemble*`.

But otherwise it looks good to me.

jrauh01:
Personally I would keep the names as-is.

library/Icinga/Web/Form/RepositoryForm.php
```diff
@@ -0,0 +1,292 @@
...
+    /** @var ?array<string, mixed> The data of the entry to pre-populate the form with */
+    protected ?array $data = null;
+
+    /**
+     * Create a new repository form
+     *
+     * Applies the default element decorators and registers a {@see Form::ON_REQUEST} handler
+     * that calls the mode-appropriate {@see onInsertRequest()}, {@see onUpdateRequest()},
+     * or {@see onDeleteRequest()} when the form receives a request.
+     *
```
TheSyscall:
```suggestion
     * Applies the default element decorators and calls the mode-appropriate
     * {@see onInsertRequest()}, {@see onUpdateRequest()}, or {@see onDeleteRequest()}
     * when the form receives a request.
```

How the methods are called could be considered an implementation detail.

library/Icinga/Web/Form/RepositoryForm.php
```diff
@@ -0,0 +1,292 @@
...
+
+    /**
+     * Get the target being worked with
+     *
+     * @return string
+     */
+    public function getBaseTable(): string
+    {
+        return $this->repository->getBaseTable();
+    }
```
TheSyscall, line 130:
```suggestion
```

Why does this method exist? It always returns the same result as `$this->getRepository()->getBaseTable()`.

`Icinga\Forms\RepositoryForm` allowed the form to override the base table so it made sense there.

library/Icinga/Web/Form/RepositoryForm.php
```diff
@@ -0,0 +1,292 @@
...
+    {
+        $data = $this->data;
+        if ($data === null) {
+            $row = $this->fetchEntry();
+            if ($row === false) {
+                throw new NotFoundError('Entry "%s" not found', $this->getIdentifier());
+            }
+
+            $data = get_object_vars($row);
+        }
```
TheSyscall, line 283: Consider moving this logic into a `getData` method

TheSyscall (changes requested):

library/Icinga/Web/Form/RepositoryForm.php
jrauh01: `Icinga\Forms\RepositoryForm` had no setter, so you couldn't override the base table without extending the class due to the protected visibility. With `Icinga\Web\Form\RepositoryForm` you are still able to add a `$baseTable` property and override the getter if you want to. Additionally, it creates a direct API so you don't have to know that the base table comes from the repository. I temporarily had a use case for it in the modernizing of the `RoleForm`. But since it's gone, I wouldn't mind removing it.

jrauh01: But we would loose the extension point, as `getBaseTable()` is used inside `Icinga\Web\Form\RepositoryForm` for all repository regarding actions.

library/Icinga/Web/Form/RepositoryForm.php
jrauh01: I'm not quite sure about this, as this logic should only be used in the update mode. In insert mode I don't want an entry to be fetched if no default data is passed. But why should I move update specific logic to an general `getData()` method?

library/Icinga/Web/Form/RepositoryForm.php
TheSyscall: There is no actual usage of this in the original `RepositoryForm`. And dropping it also simplifies the queries, since a repository already contains the table.

I see no reason to keep it. If someone requires this behavior we can think about a way to bring it back.

library/Icinga/Web/Form/RepositoryForm.php
TheSyscall: I was thinking about a convenient way to access the entity data inside assemble to allow for dynamic forms, but that would only apply to exactly this case anyway.

library/Icinga/Web/Form/RepositoryForm.php
jrauh01: Removed with all callers

TheSyscall (dismissed):

library/Icinga/Web/Form/RepositoryForm.php
```diff
@@ -0,0 +1,269 @@
...
+    /** @var ?array<string, mixed> The data of the entry to pre-populate the form with */
+    protected ?array $data = null;
+
+    /**
+     * Create a new repository form
+     *
+     * Applies the default element decorators and calls the mode-appropriate
+     * {@see onInsertRequest()}, {@see onUpdateRequest()}, or {@see onDeleteRequest()}
+     * when the form receives a request.
+     * @param Repository $repository The repository to work with
```
lippserd, line 52: Please separate the constructor description from its tags.

```suggestion
     *
     * @param Repository $repository The repository to work with
```

library/Icinga/Web/Form/RepositoryForm.php
```diff
@@ -0,0 +1,269 @@
...
+     * @param RepositoryMode $mode How to interact with the repository
+     */
+    public function __construct(
+        protected Repository $repository,
```
lippserd, line 61: `Repository` only guarantees `select()`, but this class calls `insert()`, `update()`, and `delete()`. Please require `Repository&Extensible&Updatable&Reducible` here, with the corresponding imports, or validate the selected mode’s capability during construction. The current type accepts `LdapRepository`, and PHPStan reports all three mutation calls as undefined.

library/Icinga/Web/Form/RepositoryForm.php
```diff
@@ -0,0 +1,269 @@
...
+    /**
+     * Set the name of the entry to handle
+     *
+     * Required when in mode {@see RepositoryMode::Update} or {@see RepositoryMode::Delete}.
+     *
+     * @param string $name Entry name
+     *
+     * @return $this
+     */
+    public function setIdentifier(string $name): static
```
lippserd: Please establish the identifier together with the mode instead of leaving update and delete forms partially initialized. Make it a constructor argument, require an explicit `null` for insert, reject `null` for update and delete immediately, and remove this setter. Otherwise a POST skips `ON_REQUEST` and can reach `createFilter()` without an identifier.

lippserd (changes requested):
I’m requesting changes because the new persistence base drops the legacy form’s automatic CSRF and form-identity protection. Please make the base use and assemble `CsrfCounterMeasure` and `FormUid` before mode-specific elements, with stable identifiers that subclasses cannot omit.

Please add focused tests for all modes, security and form identity, identifier and repository capability validation, population and fetching, missing entries, filters, and `ON_ERROR`/`ON_SUBMIT`. Document the non-drop-in migration, including construction and request changes, notification ownership, and security and submit controls.

Please move the deprecation of `Icinga\Forms\RepositoryForm` and `ConfigFormEventsHook`, together with the removal of the hook documentation, into a separate pull request targeted at version 2.15 instead of the upcoming 2.14.1 release. Announce the new `Icinga\Web\Form\RepositoryForm` in `CHANGELOG.md` and note there that the old repository form will be deprecated starting with version 2.15.

library/Icinga/Web/Form/RepositoryForm.php
jrauh01: I've added constructor validation so the repository at least has to implement the interface required by the respective mode. The mode provides the required interface via a getter on the `RepositoryMode` enum. Additionally, I've added three getter functions returning the repository typed as the required interface so the call sites can rely on the interface implementation and PHPStan passes even at level 10.

library/Icinga/Web/Form/RepositoryForm.php
jrauh01: I've added constructor validator here as well. The `RepositoryMode` enum provides whether a mode requires an identifier or not. For null identifier where required an exception is thrown. I did not add the requirement of explicit `null` for insert mode, so implementors are free to use it.

jrauh01:
I moved the deprecations of `Icinga\Forms\RepositoryForm` and `Icinga\Application\Hook\ConfigFormEventsHook` to #5553 planned for version 2.15.

Additional changes:
- addition of CSRF and form UID
- unit tests for `RepositoryForm`
- changelog entry stating this is not a drop-in replacement
- the $identifier property now also accepts `Stringable`

lippserd (dismissed):

lippserd:
@jrauh01 Please squash.

jrauh01:
The commits are now squashed.

lippserd (approved):

PR #5537, opened 2026-07-08
Modernize password policy

jrauh01:
Moves the password policy off the Zend form stack. `PasswordPolicyHelper` was typed against `Icinga\Web\Form`, rendered its callouts as pre-rendered `note` elements and attached a `PasswordPolicyValidator`. The policy selection was a subform of `GeneralConfigForm`.

## Changes

- `PasswordPolicyHelper` is typed against `CompatForm`, uses an inline `CallbackValidator`, and emits its callouts via `addHtml()`.
- `PasswordPolicyValidator` is removed.
- `PasswordPolicyConfigForm` becomes a standalone `ConfigForm` on the security page above the CSP section.
- `ChangePasswordForm` extends `CompatForm`. The backend and user are constructor parameters, the confirmation uses `hash_equals`, and CSRF token and form UID are added explicitly.
- `UserForm` and `UserController` follow the `RepositoryForm` API from #5536. `getIdentifier()` is narrowed to `?string`.
- Password arguments are marked `#[SensitiveParameter]`.
- The setup wizard no longer enforces the policy, as it offers no UI to select beforehand.


application/forms/Account/ChangePasswordForm.php
```diff
@@ -8,118 +8,98 @@
...
+            static::NEW_PASSWORD_ELEMENT_NAME,
+            static::OLD_PASSWORD_ELEMENT_NAME,
+        );
+
+        $this->addElement('password', static::NEW_PASSWORD_ELEMENT_NAME . '_confirmation', [
+            'label'      => $this->translate('Confirm New Password'),
+            'required'   => true,
+            'validators' => [new CallbackValidator(
+                function (#[SensitiveParameter] mixed $value, CallbackValidator $validator): bool {
+                    if (! hash_equals($this->getValue(static::NEW_PASSWORD_ELEMENT_NAME), $value)) {
```
lippserd: A user can trigger an uncaught `TypeError` through the normal account form: enter the valid old password, leave **New Password** empty, enter any non-empty value under **Confirm New Password**, and submit. The empty new-password field is normalized to `null`, while the non-empty confirmation still runs this validator, resulting in `hash_equals(null, $value)`.

Please read the new value first and reject non-string operands before performing the comparison:

```php
$newPassword = $this->getValue(static::NEW_PASSWORD_ELEMENT_NAME);
if (
    ! is_string($newPassword)
    || ! is_string($value)
    || ! hash_equals($newPassword, $value)
) {
```

application/forms/Config/User/UserForm.php
```diff
@@ -5,228 +5,136 @@
...
-        }
+        $this->assembleCommonElements();
 
-        return $values;
+        $this->addElement('password', 'password', [
+            'description' => $this->translate('Leave empty for not updating the user\'s password'),
+            'label'       => $this->translate('Password'),
+        ]);
+
+        $user = new User($this->getIdentifier());
```
lippserd: When a username and password are changed in the same submission, `getIdentifier()` is still the original username. The policy therefore validates the password against the old user identity even though the repository subsequently stores it for the renamed user.

Do we actually want to require a `User` argument in `PasswordPolicy::validate()` now? This PR has no concrete policy that uses the user. Checking whether a password resembles the username is only a possible future use case. The rename behavior already exposes ambiguity about which identity the contract should provide. Is this breaking API change worth introducing before a real consumer defines the required semantics?

application/forms/Config/General/PasswordPolicyConfigForm.php
```diff
@@ -5,54 +5,60 @@
...
         $defaultPolicy = PasswordPolicyHook::DEFAULT_PASSWORD_POLICY;
-        $elementName = sprintf('%s_%s', PasswordPolicyHook::CONFIG_SECTION, PasswordPolicyHook::CONFIG_KEY);
+        $elementName = sprintf('%s__%s', PasswordPolicyHook::CONFIG_SECTION, PasswordPolicyHook::CONFIG_KEY);
+
+        try {
+            $policies = iterator_to_array(PasswordPolicyHook::yieldPolicies());
+        } catch (Throwable $e) {
+            $this->logAndShowError($e, $this->translate('Could not load password policies: {error}'));
+
+            return;
```
lippserd, line 48: Returning from `assemble()` does not make this error state non-submittable. `ConfigForm` still adds the Store and CSRF controls after assembly. Because no configuration-backed policy element exists, `save()` leaves the current policy unchanged, but the form can still emit its submit event, show a success notification, and redirect as though something had been stored.

There are several sound ways to avoid that false success: retain an error-state property that both suppresses the Store control and forces validation to fail; add a form-level validation failure whenever policy discovery failed; or let the exception propagate so the controller or global error handling aborts the request instead of rendering a submittable form.

application/forms/Config/General/PasswordPolicyConfigForm.php
```diff
@@ -5,54 +5,60 @@
...
-use Exception;
 use Icinga\Application\Config;
 use Icinga\Application\Hook\PasswordPolicyHook;
 use Icinga\Application\Logger;
 use Icinga\Authentication\PasswordPolicyHelper;
 use Icinga\Exception\IcingaException;
-use Icinga\Web\Form;
+use Icinga\Web\Form\ConfigForm;
+use ipl\Web\Common\CsrfCounterMeasure;
+use ipl\Web\Common\FormUid;
```
lippserd: `CsrfCounterMeasure` is unused here because `ConfigForm` already owns that trait.

```suggestion
use ipl\Web\Common\FormUid;
```

lippserd (changes requested):
In addition to the requested changes, I want to raise a few other concerns.

Why isn't `ConfigFormEventsHook` deprecated as part of this change?

Please keep the [`PasswordPolicyHelper` conversion commit](https://github.com/Icinga/icingaweb2/commit/524a5da21ae921377b55654a44c932c3a0785d84) and its callers buildable together. That commit changes `PasswordPolicyHelper::apply()` to require `CompatForm`, while `ChangePasswordForm` and `UserForm` still extend the legacy form classes until later commits. Please reorder or squash the caller conversions so each commit can be checked out and tested independently.

The [shipped custom-policy example](https://github.com/Icinga/icingaweb2/blob/0757352d09c22f0524392dcd016171736ac1a79c/doc/05-Authentication.md#L233) still implements the old `validate(string, ?string)` signature. If the required `User` argument remains part of this change, the example needs to be updated as well.

Moving password-policy configuration from General to Security changes its delegated administration permission from `config/general` to `config/security`. Please document this permission move in the 2.14.1 changelog so operators know that affected roles must be updated.

Please tighten these three commit messages before merge:

- [`9201558`](https://github.com/Icinga/icingaweb2/commit/9201558f21c4263ad56fd06c05d5438ef204a492): “Use `#[SensitiveParameter]` in CallbackValidator” suggests that the validator class changed. “Mark password-policy callback values as sensitive” describes the change more accurately.
- [`81debb2`](https://github.com/Icinga/icingaweb2/commit/81debb28536402230402b11b40b65f7cac33e290): replace the em-dash aside in the body with normal ASCII punctuation.
- [`0757352`](https://github.com/Icinga/icingaweb2/commit/0757352d09c22f0524392dcd016171736ac1a79c): replace the vague “Add changelog entry” subject with “Document user-aware password policies”; split the unrelated #5552 and #5536 link additions or broaden the subject if they remain.

application/forms/Account/ChangePasswordForm.php
jrauh01: What about splitting into two checks to provide differing error messages?

```php
$newPassword = $this->getValue(static::NEW_PASSWORD_ELEMENT_NAME);
if (! is_string($newPassword) || ! is_string($value)) {
    $validator->addMessage($this->translate('Invalid password provided'));

    return false;
}

if (! hash_equals($newPassword, $value)) {
    $validator->addMessage($this->translate('The passwords do not match'));

    return false;
}
```

application/forms/Config/User/UserForm.php
jrauh01: I would change the `getIdentifier()` to `getValue()`, which also provides a changed user name.

I would definitely require the username and to be honest I would extend our `CommonPasswordPolicy` by checking that the username must not be contained in the new password. At least the password and username must not match exactly.

lippserd: Ok.

application/forms/Account/ChangePasswordForm.php
lippserd: Sounds good. But the first error message should rather be something like "Password must be string".

application/forms/Config/General/PasswordPolicyConfigForm.php
```diff
@@ -5,54 +5,64 @@
...
         try {
-            $policy = PasswordPolicyHook::fromCanonicalName($formData[$elementName] ?? $defaultPolicy);
+            $policy = PasswordPolicyHook::fromCanonicalName($this->getPopulatedValue($elementName, $defaultPolicy));
         } catch (Throwable $e) {
             Logger::error("%s\n%s", $e, IcingaException::getConfidentialTraceAsString($e));
             PasswordPolicyHelper::addError($this, true);
 
-            return $this;
+            return;
         }
```
lippserd, line 84: Returning here leaves `policiesLoadable` true, so the form keeps its Store control and the `<select>` renders with nothing `selected`. The browser then shows the first option as current, and storing silently replaces the configured policy. The path above deliberately makes the form non-submittable. This one does not.

Render the stored value and disable it. `SelectElement`'s `DeferredInArrayValidator` skips disabled options, so it cannot be re-saved while another policy still can:

```php
$stored = $this->getPopulatedValue($elementName, $defaultPolicy);
/** @var SelectElement $element */
$element = $this->getElement($elementName);
$element
    ->setOptions([$stored => sprintf($this->translate('%s (unknown)'), $stored)] + $policies)
    ->setDisabledOptions([$stored]);
```

Needs a `SelectElement` import, hence no suggestion.

Unrelated, but adjacent: this path shows a generic message while the enumeration failure above surfaces the raw exception through `logAndShowError()`. Worth picking one.

doc/80-Upgrading.md
```diff
@@ -5,6 +5,13 @@ v2.6 to v2.8 requires to follow the instructions for v2.7 too.
 
 ## Upgrading to Icinga Web 2.14.1
 
+**Deprecations**
+
+* `Icinga\Application\Hook\ConfigFormEventsHook` is deprecated now. There will
+  be no replacement, as the hook never evolved into a proper implementation.
+  * Existing hook implementations continue to work until v2.15, which removes
+    the deprecated hook.
```
lippserd: `UserForm` stops invoking the hook in this release, so this does not hold. And since the hook is typed against Zend's form implementation, every form migrated to `CompatForm` drops it.

```suggestion
* `Icinga\Application\Hook\ConfigFormEventsHook` is now deprecated. There will
  be no replacement, as the hook never evolved into a proper implementation.
  * The hook is only invoked by config forms that still build on
    `Icinga\Web\Form`. Forms migrated to `CompatForm` no longer invoke it,
    because the hook's contract is typed against Zend's form implementation.
    In 2.14.1 this affects `Icinga\Forms\Config\User\UserForm`.
  * All remaining invocations are removed in v2.15.
```

library/Icinga/Authentication/PasswordPolicyHelper.php
```diff
@@ -58,26 +72,88 @@ public static function apply(
...
+                    $oldPasswordValue = $form->getValue($oldPasswordElementName);
+                    // A crafted request can submit an array, which Str::isEmpty() rejects with a
+                    // TypeError. Such a value is left to the old-password element's own validator.
+                    if (is_string($oldPasswordValue) && ! Str::isEmpty($oldPasswordValue)) {
+                        $oldPassword = $oldPasswordValue;
+                    }
+                }
+
+                try {
+                    $messages = $passwordPolicy->validate($value, $oldPassword);
```
lippserd, line 147: `$value` is `mixed` but `validate()` is typed `string`, while the old password above is guarded. `password[]=x` yields "Password validation failed" instead of the "Password must be a string" used elsewhere, and third-party policies get the `TypeError`.

```suggestion
                if (! is_string($value)) {
                    $validator->addMessage(t('Password must be a string'));

                    return false;
                }

                try {
                    $messages = $passwordPolicy->validate($value, $oldPassword);
```

The two `...IsRejectedWhenPolicyValidationThrows` tests pin the old message.

library/Icinga/Authentication/PasswordPolicyHelper.php
```diff
@@ -33,21 +37,31 @@ class PasswordPolicyHelper
...
     public static function apply(
-        Form $form,
+        CompatForm $form,
         string $newPasswordElementName,
-        ?string $oldPasswordElementName = null
+        ?string $oldPasswordElementName = null,
+        bool $adminFacing = false,
     ): void {
-        if ($oldPasswordElementName !== null && $form->getElement($oldPasswordElementName) === null) {
+        if ($oldPasswordElementName !== null && ! $form->hasElement($oldPasswordElementName)) {
```
lippserd, line 71: `$newPasswordElementName` is required but unchecked, then dereferenced at both `getElement()` calls below.

```suggestion
        if (! $form->hasElement($newPasswordElementName)) {
            throw new LogicException(sprintf(
                t('Form element "%s" was specified but does not exist in the form'),
                $newPasswordElementName
            ));
        }

        if ($oldPasswordElementName !== null && ! $form->hasElement($oldPasswordElementName)) {
```

library/Icinga/Authentication/PasswordPolicyHelper.php
```diff
@@ -33,21 +37,31 @@ class PasswordPolicyHelper
...
      *
      * @throws LogicException If the old password element is specified but does not exist in the form
      */
     public static function apply(
-        Form $form,
+        CompatForm $form,
         string $newPasswordElementName,
-        ?string $oldPasswordElementName = null
+        ?string $oldPasswordElementName = null,
+        bool $adminFacing = false,
```
lippserd, line 62: `$adminFacing` here, `$forAdmin` on `addError()`.

test/php/library/Icinga/Application/CommonPasswordPolicyTest.php
```diff
@@ -70,6 +70,7 @@ public function testValidatePasswordToShortAndOnlyUpperCaseLetters(): void
             'Password must contain at least one special character',
             'Password must contain at least one lowercase letter'
         ];
-        $this->assertSame($expected, (new CommonPasswordPolicy())->validate('ICINGAADMIN'));
+        $this->assertSame($expected, (new CommonPasswordPolicy())->validate('TEST'));
     }
+
 }
```
lippserd: Leftover blank line. phpcs: the closing brace must go on the next line.

```suggestion
}
```

application/forms/Config/User/UserForm.php
```diff
@@ -5,228 +5,131 @@
...
     }
 
     /**
-     * Create and add elements to this form to update a user
-     *
-     * @param array $formData The data sent by the user
+     * Create and add elements to this form to insert or update a user
      *
      * @return void
      */
```
lippserd, line 54: `assembleUpdateElements()` does not call this.

```suggestion
    /**
     * Create and add elements to this form to insert a user
     *
     * @return void
     */
```

lippserd (changes requested):
The PR description still advertises `PasswordPolicy::validate()` taking a required `User` first parameter. Please drop that bullet.

`PasswordPolicyConfigForm` still lives in `Icinga\Forms\Config\General` but renders only on the Security page. Please move it next to `CspConfigForm` under `Icinga\Forms\Config\Security`.

`doc/80-Upgrading.md`'s 2.14.1 breaking changes omit that managing the password policy now requires `config/security` instead of `config/general`. `CHANGELOG.md` has it with remediation. Please mirror it there.

Test gaps:

* The length threshold is never isolated. The only length case bundles three other violations.
* Several `UserFormTest` write-path tests call `exposeOnSuccess()` without asserting `isValid()` first.
* `testUpdateModeKeepsPasswordWhenLeftEmpty` runs only under `any`, not under a restrictive policy.
* Nothing covers `PasswordPolicyConfigForm` or `securityAction()`'s two-forms-one-request wiring, although `CspConfigFormTest` exists for the sibling.

library/Icinga/Authentication/PasswordPolicyHelper.php
jrauh01: I just removed the two tests, as the validation isn't throwing now anymore.

library/Icinga/Authentication/PasswordPolicyHelper.php
jrauh01: Changed to `$adminFacing` for `addError()` as well.

application/forms/Config/User/UserForm.php
jrauh01: Applied

application/forms/Config/General/PasswordPolicyConfigForm.php
jrauh01: Done both, the disabled option and the `logAndShowError()` unification.

PR #5539, opened 2026-07-09
Documentation updates

TheSyscall:
- Add a general introduction to the hooks chapter explaining how to implement and register a hook, and how to define a new hook type using `HookEssentials`
- Add missing `config/security` permission to the permissions table in the security chapter

### DIFF
--- a/doc/03-Configuration.md
+++ b/doc/03-Configuration.md
@@ -66,7 +66,7 @@ file = "/usr/share/icingaweb2/log/icingaweb2.log"
 
 Option                   | Description
 -------------------------|-----------------------------------------------
-default                  | **Optional.** Choose the default theme. Can be set to `Icinga`, `high-contrast`, `Winter`, 'colorblind' or your own installed theme. Defaults to `Icinga`. Note that this setting is case-sensitive because it refers to the filename of the theme.
+default                  | **Optional.** Choose the default theme. Can be set to `Icinga`, `high-contrast`, `Winter`, `colorblind` or your own installed theme. Defaults to `Icinga`. Note that this setting is case-sensitive because it refers to the filename of the theme.
 disabled                 | **Optional.** Set this to `1` if users should not be allowed to change their theme. Defaults to `0`.
 default_mode             | **Optional.** Choose the default theme mode. Can be set to `system`, `light` or `none` for the dark theme mode. Defaults to `none`.
 
--- a/doc/03-Configuration.md
+++ b/doc/03-Configuration.md
@@ -25,11 +25,12 @@ This configuration is stored in the `config.ini` file in `/etc/icingaweb2`.
 ### Global Configuration <a id="configuration-general-global"></a>
 
 
-Option                   | Description
--------------------------|-----------------------------------------------
-show\_stacktraces        | **Optional.** Set to `1` to show debug stacktraces. Defaults to `1`.
-module\_path             | **Optional.** Specifies the directories where modules can be installed. Multiple directories must be separated with colons.
-config\_resource         | **Required.** Specify a defined [resource](04-Resources.md#resources-configuration-database) name.
+Option                             | Description
+-----------------------------------|-----------------------------------------------
+show\_stacktraces                  | **Optional.** Set to `1` to show debug stacktraces. Defaults to `1`.
+show\_application\_state\_messages | **Optional.** Set to `1` to show application state messages. Defaults to `1`.
+module\_path                       | **Optional.** Specifies the directories where modules can be installed. Multiple directories must be separated with colons.
+config\_resource                   | **Required.** Specify a defined [resource](04-Resources.md#resources-configuration-database) name.
 
 
 Example for storing the user preferences in the database resource `icingaweb_db`:
@@ -37,6 +38,7 @@ Example for storing the user preferences in the database resource `icingaweb_db`
 ```
 [global]
 show_stacktraces = "1"
+show_application_state_messages = "1"
 config_resource = "icingaweb_db"
 module_path = "/usr/share/icingaweb2/modules"
 ```
--- a/doc/03-Configuration.md
+++ b/doc/03-Configuration.md
@@ -79,6 +79,19 @@ default = "Icinga"
 default_mode = "light"
 ```
 
+### Authentication Configuration <a id="configuration-general-authentication"></a>
+
+Option                   | Description
+-------------------------|-----------------------------------------------
+default\_domain          | **Optional.** Domain assumed when users log in without specifying a domain. It must match an LDAP backend's configured domain or the domain suffix stored in another backend's usernames, e.g. `icinga.com`.
+
+Example:
+
+```
+[authentication]
+default_domain = "icinga.com"
+```
+
 ## Security Configuration <a id="configuration-security"></a>
 
 Navigate into **Configuration > Application > Security**.
--- a/doc/03-Configuration.md
+++ b/doc/03-Configuration.md
@@ -115,19 +115,6 @@ Navigate into **Configuration > Application > Security**.
 
 This configuration is stored in the `config.ini` file in `/etc/icingaweb2`.
 
-### Password Policy Configuration <a id="configuration-security-password-policy"></a>
-
-Option            | Description
-------------------|-----------------------------------
-password\_policy  | **Optional.** Canonical name of the [password policy](05-Authentication.md#authentication-password-policy) to enforce. Defaults to `any` (None).
-
-Example:
-
-```
-[security]
-password_policy = "common"
-```
-
 ### Content Security Policy Configuration <a id="configuration-security-csp"></a>
 
 Option                  | Description
--- a/doc/03-Configuration.md
+++ b/doc/03-Configuration.md
@@ -48,9 +48,9 @@ module_path = "/usr/share/icingaweb2/modules"
 Option                   | Description
 -------------------------|-----------------------------------------------
 log                      | **Optional.** Specifies the logging type. Can be set to `syslog`, `file`, `php` (web server's error log) or `none`.
-level                    | **Optional.** Specifies the logging level. Can be set to `ERROR`, `WARNING`, `INFORMATION` or `DEBUG`.
+level                    | **Optional.** Specifies the logging level. Can be set to `ERROR`, `WARNING`, `INFO` or `DEBUG`.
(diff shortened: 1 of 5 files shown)


jrauh01:
Converted this to draft until #5537 is ready.

doc/03-Configuration.md
```diff
@@ -75,22 +77,64 @@ disabled = "1"
...
-| use\_custom\_csp        | **Optional.** Set this to `1` to enable the use of the user defined Content Security Policy. Defaults to `0`.                                             |
-| custom\_csp             | **Optional.** Specifies the user defined Content Security Policy. Overrides the automatically generated one. Only used if `use_custom_csp` is set to `1`. |
-| csp\_enable\_modules    | **Optional.** Specifies if modules should be included in the generated Content Security Policy. Defaults to `1`.                                          |
-| csp\_enable\_dashboards | **Optional.** Specifies if dashboards should be included in the generated Content Security Policy. Defaults to `1`.                                       |
-| csp\_enable\_navigation | **Optional.** Specifies if navigation menu items should be included in the generated Content Security Policy. Defaults to `1`.                            |
+Option                  | Description
+------------------------|------------------------------
+use\_strict\_csp        | **Optional.** Set this to `1` to enable strict [Content Security Policy](
```
lippserd: This link is truncated, and the conversion also dropped the `use_custom_csp` option even though it is used by the runtime and the example below. Please restore both rows.

```suggestion
use\_strict\_csp        | **Optional.** Set this to `1` to enable strict [Content Security Policy](20-Advanced-Topics.md#advanced-topics-csp). Defaults to `0`.
use\_custom\_csp        | **Optional.** Set this to `1` to use the user-defined Content Security Policy. Defaults to `0`.
```

doc/03-Configuration.md
```diff
@@ -75,22 +77,64 @@ disabled = "1"
 default = "high-contrast"
 ```
 
+### Authentication Configuration <a id="configuration-general-authentication"></a>
+
+Option                   | Description
+-------------------------|-----------------------------------------------
+default\_domain          | **Optional.** Domain assumed when users log in without specifying a domain. The value must match the domain of a configured authentication backend, e.g. `icinga.com`.
```
lippserd: This does not have to match a backend's configured `domain`. Database and other backends can instead store usernames containing the suffix.

```suggestion
default\_domain          | **Optional.** Domain assumed when users log in without specifying a domain. It must match an LDAP backend's configured domain or the domain suffix stored in another backend's usernames, e.g. `icinga.com`.
```

doc/60-Hooks.md
```diff
@@ -1,5 +1,74 @@
...
+{
+    return true;
+}
+```
+
+### Implement the new Hook <a id="hooks-implement-new-hook"></a>
+
+In other modules you can now provide an implementation for your own hook. It works
+basically the same like described earlier, but the class goes in a sub namespace
+of the module's `ProvidedHook` namespace this time:
```
lippserd: “The same like described earlier” is ungrammatical, and “sub namespace” is not the usual term.

```suggestion
Other modules can now provide implementations of the hook. Registration works
as described earlier, but the class goes in a subnamespace of the implementing
module's `ProvidedHook` namespace:
```

doc/60-Hooks.md
```diff
@@ -89,3 +160,174 @@ class Csp extends CspHook
...
+
+Method                                   | Description
+-----------------------------------------|-------------------------------------------------------------------
+`getName(): string`                      | Machine-readable identifier, e.g. `totp`. Must be unique within the providing module and a valid HTML `name` attribute value.
+`getDisplayName(): string`               | Human-readable name shown in the two-factor config UI, e.g. `TOTP`.
+`isEnrolled(): bool`                     | Whether the user has an active enrollment in this method. Implementations typically query the database for a stored credential.
+`verify(): bool`                         | Verify the token the user entered during login against the stored credential.
+`enroll(): bool`                         | Read the submitted credential from the fieldset, verify it works, and store it. Return `false` and attach an error to the relevant element via `addMessage()` on failure.
+`unenroll(): void`                       | Remove the stored credential for the user.
+`assembleEnrollmentFormElements(): void` | Add the method-specific form elements to the enrollment fieldset. Don't rename the fieldset.
```
lippserd: Honestly, I would remove this table. The interface is the authoritative contract, and implementers will work against that code rather than a duplicated documentation table that can become stale.

If the table stays, it must include the required arguments:

```suggestion
`isEnrolled(User $user): bool`                                      | Whether the user has an active enrollment in this method. Implementations typically query the database for a stored credential.
`verify(User $user, string $token): bool`                           | Verify the token the user entered during login against the stored credential.
`enroll(User $user, FieldsetElement $fieldset): bool`               | Read the submitted credential from the fieldset, verify it works, and store it. Return `false` and attach an error to the relevant element via `addMessage()` on failure.
`unenroll(User $user): void`                                        | Remove the stored credential for the user.
`assembleEnrollmentFormElements(User $user, FieldsetElement $fieldset): void` | Add the method-specific form elements to the enrollment fieldset. Don't rename the fieldset.
```

doc/60-Hooks.md
```diff
@@ -89,3 +160,174 @@ class Csp extends CspHook
...
+    {
+        return $this->translate('TOTP');
+    }
+
+    public function isEnrolled(User $user): bool
+    {
+        return $this->loadSecret($user) !== null;
+    }
+
+    public function verify(User $user, string $token): bool
```
lippserd: Please mark the live authentication token with `#[SensitiveParameter]` and add the corresponding import.

```suggestion
    public function verify(User $user, #[SensitiveParameter] string $token): bool
```

This should also have separate follow-up changes in the actual contracts and implementations. Parameter attributes are not inherited, so annotating only `Icinga\Authentication\TwoFactor` is insufficient. Icinga TOTP Web should annotate `TwoFactorTotp::verify()`, `TotpInterface::verify()`, and `Totp::verify()`. Complete redaction through the third-party call also requires an upstream OTPHP change for `TOTP::verify()` and the token operands passed to `compareOTP()`. Those follow-ups do not need to block this documentation fix.

doc/60-Hooks.md
```diff
@@ -89,3 +160,174 @@ class Csp extends CspHook
...
+    }
+
+    public function getDisplayName(): string
+    {
+        return $this->translate('TOTP');
+    }
+
+    public function isEnrolled(User $user): bool
+    {
+        return $this->loadSecret($user) !== null;
```
lippserd, line 212: This is the first of several helper methods that neither `TwoFactorHook` nor this example defines: `loadSecret()`, `checkToken()`, `generateSecret()`, `storeSecret()`, and `deleteSecret()`. Please state before the example that these are placeholders which a real implementation must provide, including secure secret generation, token verification, and protected persistent storage.

lippserd (changes requested):
In addition to the changes requested, please address the following before this can be merged:

The password-policy documentation changes belong in #5537 instead of being made compatible with the current `main` branch here. They already document #5537's `User` argument and Security-page placement, and keeping them in this PR would require both PRs to remain independently correct while that contract is still under discussion. Please move these changes to #5537.

While moving them, please also use `mb_strlen()` so that the advertised eight-character minimum counts characters rather than bytes, and explicitly state that `PasswordPolicyHook` always runs regardless of the providing module's `module/<module-name>` permission.

doc/60-Hooks.md
jrauh01: Done

doc/60-Hooks.md
jrauh01: Removed the table completely.

doc/60-Hooks.md
jrauh01: Added and I will provide a PR for the other changes.

doc/03-Configuration.md
```diff
@@ -77,22 +79,65 @@ default = "Icinga"
...
+secure = "1"
+```
+
 ## Security Configuration <a id="configuration-security"></a>
 
 Navigate into **Configuration > Application > Security**.
 
 This configuration is stored in the `config.ini` file in `/etc/icingaweb2`.
 
+### Password Policy Configuration <a id="configuration-security-password-policy"></a>
```
lippserd: Can't we move this to the modernize password policy PR as well?

doc/03-Configuration.md
```diff
@@ -77,22 +79,65 @@ default = "Icinga"
...
-| csp\_enable\_navigation | **Optional.** Specifies if navigation menu items should be included in the generated Content Security Policy. Defaults to `1`.                            |
+Option                  | Description
+------------------------|------------------------------
+use\_strict\_csp        | **Optional.** Set this to `1` to enable strict [Content Security Policy](20-Advanced-Topics.md#advanced-topics-csp). Defaults to `0`.
+use\_custom\_csp        | **Optional.** Set this to `1` to use the user-defined Content Security Policy. Defaults to `0`.
+custom\_csp             | **Optional.** Specifies the user defined Content Security Policy. Overrides the automatically generated one. Only used if `use_custom_csp` is set to `1`.
+csp\_enable\_modules    | **Optional.** Specifies if modules should be included in the generated Content Security Policy. Defaults to `1`.
+csp\_enable\_dashboards | **Optional.** Specifies if dashboards should be included in the generated Content Security Policy. Defaults to `1`.
+csp\_enable\_navigation | **Optional.** Specifies if navigation menu items should be included in the generated Content Security Policy. Defaults to `1`.
```
lippserd: The runtime fallback for `csp_enable_*` is `0`, not `1`. The UI preselects `1` only when strict CSP is enabled for the first time. An administrator enabling CSP directly in `config.ini` could omit these settings based on the documentation. What about something like the following:

```suggestion
csp\_enable\_modules    | **Optional.** Specifies if modules should be included in the generated Content Security Policy. Defaults to `0`.
csp\_enable\_dashboards | **Optional.** Specifies if dashboards should be included in the generated Content Security Policy. Defaults to `0`.
csp\_enable\_navigation | **Optional.** Specifies if navigation menu items should be included in the generated Content Security Policy. Defaults to `0`.

When strict CSP is enabled through the Security page for the first time, the `csp_enable_*` options are enabled by default.
```

lippserd (changes requested):

doc/03-Configuration.md
jrauh01: Removed. Will be added in password policy PR.

PR #5541, opened 2026-07-13
Add configurable default theme mode to theming configuration

### LINKED ISSUE #5257: option to set light-mode in confi.ini
the default is set to dark-mode and cannot be changed in the config.ini settings, only the themes are changeable. The possibility, for example, is to set the light-mode, bright mode or system by default in the config.ini.

TheSyscall:
Allow administrators to configure the default theme mode (system, dark, or light) via the theming configuration form. Previously, users without a personal theme mode preference always fell back to the hard-coded DEFAULT_MODE constant. With this change, the configured default is used as the fallback instead, giving administrators control over the out-of-the-box appearance for their users.

resolves #5257

### DIFF
--- a/application/forms/Config/General/ThemingConfigForm.php
+++ b/application/forms/Config/General/ThemingConfigForm.php
@@ -60,6 +60,21 @@ class ThemingConfigForm extends Form
             ]
         );
 
+        $this->addElement(
+            'select',
+            'themes_default_mode',
+            [
+                'description'  => $this->translate('The default theme mode', 'Form element description'),
+                'label'        => $this->translate('Default Theme Mode', 'Form element label'),
+                'value'        => StyleSheet::DEFAULT_MODE,
+                'multiOptions' => [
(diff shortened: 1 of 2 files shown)


jrauh01 (approved):

application/forms/Config/General/ThemingConfigForm.php
```diff
@@ -60,6 +60,21 @@ public function createElements(array $formData)
             ]
         );
 
+        $this->addElement(
+            'select',
+            'themes_default_mode',
+            [
+                'description'  => $this->translate('The default theme mode', 'Form element description'),
+                'label'        => $this->translate('Default Theme Mode', 'Form element label'),
```
nilmerg, line 68: If I may interrupt as I just seen this: Translation contexts are *usually* used to disambiguate similar messages, not to describe the usage and even if so, not in such a general manner.

Better would be to use source code comments, any comment above a translatable message will be visible to translators, besides the link to the source of course.

PR #5542, opened 2026-07-14
Check for a preference store before accepting the primary submit button

### LINKED ISSUE #5424: User Preferences: Saving for current Session results in error message
## Describe the bug
I get an error when saving the user preferences for the current session, when there is no configuration database resource defined.

## To Reproduce

1. Have no *Configuration Database* set in `/config/general`.
2. Navigate to the user account Preferences. `/account`
3. Press *Save for the current Session*.

## Expected behavior
There should be no need to configure a configuration database to store session settings.

## Your Environment

* Icinga Web 2 version and modules (System - About):  	2.12.5
* Web browser used: firefox
* Icinga 2 version used (`icinga2 --version`): r2.15.0-1
* PHP version used (`php --version`): 8.3.6
* Server operating system and version: Ubuntu 24.04.2 LTS

## Additional context
My initial attempt to debug this issue led me to the following observations.

1. If there is no *Configuration Database* set, there is no *Save to the Preferences* button
2. If that button is missing, the submit label in `icingaweb/application/forms/PreferenceForm.php` is still set to *Save to the Preferences*
3. When calling `isSubmitted` in `icingaweb2/library/Icinga/Web/Form.php`, the method assumes that the button with the submit label is called *btn_submit*. Which is not true because of observation 1.

TheSyscall:
Previously, `isSubmitted()` would treat the primary submit button `btn_submit` as a valid submission even when no preference store was configured. Without a store, saving preferences is not possible.

The fix gates the primary submit path on `$this->store !== null`, so only the session-only button `btn_submit_session` remains available when there is no store.

fixes #5424

### DIFF
--- a/application/forms/PreferenceForm.php
+++ b/application/forms/PreferenceForm.php
@@ -441,7 +441,7 @@ class PreferenceForm extends Form
 
     public function isSubmitted()
     {
-        if (parent::isSubmitted()) {
+        if ($this->store !== null && parent::isSubmitted()) {
             return true;
         }
 


jrauh01 (approved):
Works as intended!

PR #5551, opened 2026-07-20
Add documentation about the default theme mode

jrauh01:
Add documentation about the new default theme mode configuration, introduced in #5541. Additionally changed theme example from to `Icinga` because `high-contrast` has no `light` theme mode meaning the example config would not be valid.

### DIFF
--- a/doc/03-Configuration.md
+++ b/doc/03-Configuration.md
@@ -73,7 +73,7 @@ Example:
 ```
 [themes]
 disabled = "1"
-default = "high-contrast"
+default = "Icinga"
 default_mode = "light"
 ```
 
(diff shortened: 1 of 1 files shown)


doc/03-Configuration.md
```diff
@@ -66,13 +66,15 @@ Option                   | Description
...
 disabled                 | **Optional.** Set this to `1` if users should not be allowed to change their theme. Defaults to `0`.
+default_mode             | **Optional.** Choose the default theme mode. Can be set to `system`, `light` or `none` for the dark theme mode. Defaults to `none`.
 
 Example:
 
 ```
 [themes]
 disabled = "1"
 default = "high-contrast"
+default_mode = "light"
```
lippserd: Do not demonstrate `light` mode with `high-contrast`. That theme has no `@light-mode:` marker, and the preference form explicitly disables light and system modes for such themes. Use the default `Icinga` theme in the example:

```suggestion
default = "Icinga"
default_mode = "light"
```

lippserd (changes requested):

lippserd (approved):

PR #5552, opened 2026-07-21
Add `CsrfCounterMeasure` to `ConfigForm`

jrauh01:
Move CSRF protection from `CspConfigForm` into the base class so all `ConfigForm` subclasses inherit it. Replace the `ensureAssembled()` override with an `ON_ASSEMBLED` event listener, which is the idiomatic IPL Web lifecycle hook for post-assembly work.

require Icinga/ipl-html#207

### DIFF
--- a/library/Icinga/Web/Form/ConfigForm.php
+++ b/library/Icinga/Web/Form/ConfigForm.php
@@ -12,6 +12,7 @@ use ipl\Html\Contract\ValueCandidates;
 use ipl\Html\HtmlElement;
 use ipl\Html\ValidHtml;
 use ipl\Stdlib\Str;
+use ipl\Web\Common\CsrfCounterMeasure;
 use ipl\Web\Compat\CompatForm;
 use ipl\Web\Compat\DisplayFormElement;
 use ipl\Web\Widget\CopyToClipboard;
@@ -20,14 +21,20 @@ use LogicException;
 /**
  * Form base-class providing standard functionality for configuration forms
  *
- *  Element names follow a `section__key` convention: the part before the
- *  {@see $sectionKeyDelimiter} (default: `__`) is the INI section name and the
- *  part after is the configuration key within that section. Subclasses add
- *  elements whose names match this pattern; `populateFromConfig` and `save`
- *  use it to map form values to and from the INI file automatically.
+ * Element names follow a `section__key` convention: the part before the
+ * {@see $sectionKeyDelimiter} (default: `__`) is the INI section name and the
+ * part after is the configuration key within that section. Subclasses add
+ * elements whose names match this pattern; `populateFromConfig` and `save`
+ * use it to map form values to and from the INI file automatically.
+ *
+ * Before assembly, callers must either set a session-stable ID with
+ * {@see setCsrfCounterMeasureId()} or explicitly disable the countermeasure
+ * with {@see disableCsrfCounterMeasure()}.
  */
 class ConfigForm extends CompatForm
 {
+    use CsrfCounterMeasure;
+
     /** @var string Name of the submit button element */
     protected const SUBMIT_BUTTON_NAME = 'store';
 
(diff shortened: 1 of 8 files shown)


library/Icinga/Web/Form/ConfigForm.php
lippserd: Each line has two spaces instead of one after the `*`. And please add something like the following at the end:

```
*
* Controllers must call {@see setCsrfCounterMeasureId()} with a session-stable
* ID before the form is assembled.
*/
```

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -9,9 +9,11 @@
 use Icinga\Application\Config;
 use ipl\Html\Contract\FormElement;
 use ipl\Html\Contract\ValueCandidates;
+use ipl\Html\HtmlDocument;
```
lippserd: I think you can remove this import when you use `static::ON_ASSEMBLED` instead of `HtmlDocument::ON_ASSEMBLED` throughout.

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -78,21 +82,15 @@ public function __construct(
...
+        });
     }
 
-    /**
-     * Ensure that all required form elements are present
-     *
-     * @return $this
-     */
-    public function ensureAssembled(): static
+    public function isValidEvent($event)
```
lippserd: Since we require an ipl release anyway, `isValidEvent()` should be properly implemented in classes using the `FormElements` trait.

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -213,5 +211,7 @@ protected function addRequiredElements(): void
             'label' => $this->translate('Store'),
             'ignore' => true,
         ]);
+
+        $this->addCsrfCounterMeasure();
```
lippserd, line 213: PHPDoc is missing this addition. I suggest the following edit:

`Add the submit button and CSRF counter-measure element to the form`

test/php/library/Icinga/Web/Form/ConfigFormTest.php
lippserd, line 16: The PR's headline behavior has zero unit test coverage as all test cases call `disableCsrfCounterMeasure()`. Add a test that calls `setCsrfCounterMeasureId()` instead of `disableCsrfCounterMeasure()` and asserts `hasElement('CSRFToken')` after assembly:

```php
public function testCsrfElementIsAddedAfterAssembly(): void
{
    $form = new class(Config::fromArray([])) extends ConfigForm {};
    $form->setCsrfCounterMeasureId('bogus');
    $form->ensureAssembled();
    $this->assertTrue($form->hasElement('CSRFToken'));
}
```

lippserd (changes requested):

test/php/library/Icinga/Web/Form/ConfigFormTest.php
jrauh01: Added

library/Icinga/Web/Form/ConfigForm.php
jrauh01: Added

library/Icinga/Web/Form/ConfigForm.php
jrauh01: Replaced

library/Icinga/Web/Form/ConfigForm.php
jrauh01: Done

library/Icinga/Web/Form/ConfigForm.php
jrauh01: Done in Icinga/ipl-html#207.

jrauh01:
The unit tests fail due to the dependency on Icinga/ipl-html#207.

library/Icinga/Web/Form/ConfigForm.php
```diff
@@ -20,14 +21,18 @@
...
- *  part after is the configuration key within that section. Subclasses add
- *  elements whose names match this pattern; `populateFromConfig` and `save`
- *  use it to map form values to and from the INI file automatically.
+ * Element names follow a `section__key` convention: the part before the
+ * {@see $sectionKeyDelimiter} (default: `__`) is the INI section name and the
+ * part after is the configuration key within that section. Subclasses add
+ * elements whose names match this pattern; `populateFromConfig` and `save`
+ * use it to map form values to and from the INI file automatically.
+ * Controllers must call {@see setCsrfCounterMeasureId()} with a session-stable
+ * ID before the form is assembled.
```
lippserd: Please document both supported sides of the pre-assembly contract. Protection is enabled by default, but callers may explicitly opt out:

```suggestion
 * Before assembly, callers must either set a session-stable ID with
 * {@see setCsrfCounterMeasureId()} or explicitly disable the countermeasure
 * with {@see disableCsrfCounterMeasure()}.
```

lippserd (changes requested):

lippserd (approved):

PR #5566, opened 2026-08-19
Fix logging level description in configuration

chrnie:
Updated logging level option description to use 'INFO' instead of 'INFORMATION'. 

https://github.com/Icinga/icingaweb2/blob/622c30e7b7bb104afed560675fb9141111ca555d/library/Icinga/Application/Logger.php#L46-L51

If you use the UI to set the log level, it works correctly and the level is set to ‘INFO’.


nilmerg (approved):

nilmerg:
Thanks :)

